PR #21, opened 2018-09-13
add LICENSE-{MIT, APACHE}

### LINKED ISSUE #20: Needs license
This repo has no license. That's bad! The later you are thinking about this, the harder it is to introduce licenses. So better think about it now.

nikomatsakis:
Following the precedent of the [Rust reference], [Rust book], and
[RFC 2044], we adopt the LICENSE-{MIT,APACHE} dual licensing
scheme for the unsafe code guidelines as well.

Fixes #20

[Rust reference]: https://github.com/rust-lang-nursery/reference
[Rust book]: https://github.com/rust-lang/book
[RFC 2044]: http://rust-lang.github.io/rfcs/2044-license-rfcs.html

### DIFF
--- a/LICENSE-APACHE
+++ b/LICENSE-APACHE
@@ -1,0 +1,201 @@
+                              Apache License
+                        Version 2.0, January 2004
+                     http://www.apache.org/licenses/
+
+TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+1. Definitions.
+
+   "License" shall mean the terms and conditions for use, reproduction,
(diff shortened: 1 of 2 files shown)


asajeffrey:
This can be merged once @avadacatavra, @matklad and @RalfJung agree to the relicensing over in issue #20.

ubsan:
Please don't use the MIT and Apache licenses for documentation

asajeffrey:
@usban there's not much we can do about this, we'd like to at least keep open the option to merge with the language spec, which is MIT/Apache.

nikomatsakis:
OK, going to merge this.

PR #23, opened 2018-09-13
Add a skeleton Data Representation chapter

### LINKED ISSUE #22: Add a skeleton "Data Representation" chapter to the reference?
The other data representation issues can't be resolved until there's a document for them to edit.

asajeffrey:
Fixes #22


asajeffrey:
Some comments...
* File an issue for an introduction?
* Should structs and tuples be merged?
* Do we want to say anything about other representations? (e.g. `Box`, slices, fat pointers, ...)
* Do we want an ABI section?

avadacatavra:
I'd say we can keep tuples and structs together at the moment. If it seems like they need to diverge, that can be taken care of when @nikomatsakis does the writeups

PR #30, opened 2018-09-18
lvalue and rvalue should be documented as well

shepmaster:


shepmaster:
I think *whatever* the canonical Rust name becomes, you will want these two documented, even if only as redirects to the proper name. Documenting what they are and picking a name for them can be orthogonal.

steveklabnik:
We explicitly removed `rvalue` and `lvalue` from all Rust docs, in favor of 'place expressions' and 'value expressions' in all documentation a while back: https://doc.rust-lang.org/reference/expressions.html?highlight=place,exp#place-expressions-and-value-expressions

ubsan:
Another important thing is that place expressions can also evaluate to a value, and value expressions can evaluate to a place, if necessary.

First, it's important to note that there are 2 axes on which places can be put - the immutable/mutable axis, the owned/borrowed axis, and the initialized axis (ignored for this comment).

Place-to-value conversion is allowed when, given a place expression `e` of type `T`:
  - `T: Copy`
  - `e` is owned

Value-to-place conversion is allowed for all values (afaik), and results in a mutable, owned place expression.

shepmaster:
To be clear, this is just a *request* that these terms be documented, just like the existing requests for terms. It seems like all of the discussion y'all are having will be lost / not-easily findable when this PR is merged. Perhaps it's appropriate to move that discussion into an issue?

avadacatavra:
we can also just have something like
* lvalue/place
if these are equivalent terms. i'd be in favor of having some disambiguation here (since i have no clue what any of these things mean in any forms :p)

ubsan:
@avadacatavra yeah, the term across the languages is lvalue (C), place (Rust), and glvalue (C++). Basically, it means an expression that refers to a location, as opposed to just a value:

```rust
fn main() {
  0; // value expression - 0 is not a place
  let x = 0;
  x; // place expression - x is a place
}
```

avadacatavra:
i'd be fine if we wanted to add that to this PR 💯

PR #31, opened 2018-09-25
add a stucts-and-tuples chapter

### LINKED ISSUE #11: Representation of structs
Discussion topic about how structs are represented in Rust. Some things to work out:

- Do we ever say *anything* about how a `#[repr(rust)]` struct is laid out
      (and/or treated by the ABI)?
      - e.g., what about different structs with same definition
      - across executions of the same program?
- For example, [rkruppe writes](https://github.com/rust-rfcs/unsafe-code-guidelines/pull/5#discussion_r212776247) that we might "want to guarantee (some subset of) newtype  unpacking and relegate `#[repr(transparent)]` to being the way to guarantee to other crates that a type with private fields is and will remain a newtype?"
- When is interop with `#[rust(C)]` guaranteed and what can we say there?

### LINKED ISSUE #12: Representation of tuples
Discussing how tuples are laid out.

- Is tuple layout equivalent to some corresponding struct? If so, what struct exactly?
    - that struct definition might in turn have undefined layout, of course

Some things that might be useful if they were defined:

- Is it possible to inter-convert a `(T, T, ..., T)` tuple with a `[T; N]` memory representation?

### LINKED ISSUE #17: Effect of `packed` and `align` on representation
Discussion topic for the effect of `#[repr(packed)]` and `#[repr(align)]` on memory layout.

Both of these attributes have RFCs. We should document what current behavior is, what gotchas to watch out for (e.g., `&x.foo` where `foo` is a field of a packed struct may not be aligned), and what we can guarantee going forward.

nikomatsakis:
Writes up what I believe to be the consensus around layout of structs and tuples. Suggestions very welcome!

Here are some of the highlights:

- Tuples are defined to be laid out as if there were a "fresh generic struct" for their arity. So `(T1...Tn)` is laid out the same as `TupleN<T1..Tn>` where `struct TupleN<P0..Pn>(P0..Pn)`.
    - There is an exception for the case where all the types are the same (after lifetime erasure), in which case the layout is guaranteed to be compatible with `[T; N]`.
- Tuple structs are laid out the same as "named fields".
- `#[repr(Rust)]` structs (the default) have no particular layout guarantees. In particular, even if two `#[repr(Rust)]` structs have the fields of the same types, they are not guaranteed to be laid out in a compatible way. This is a conservative position that could be strengthened, though I personally now think it'd be better to "opt-in" to any such guarantees. I cover the pros and cons in the doc.
- `#[repr(C)]`, `#[repr(align)]`, `#[repr(packed)]`, and `#[repr(transparent)]` are all discussed.

Fixes #11
Fixes #12 
Fixes #17

### DIFF
--- a/reference/src/representation/structs-and-tuples.md
+++ b/reference/src/representation/structs-and-tuples.md
@@ -188,6 +188,14 @@ following:
 
 [^aligned]: Aligning an offset O to an alignment A means to round up the offset O until it is a multiple of the alignment A.
 
+One deviation from C comes about with "empty structs". In Rust, a
+struct that contains (transitively) no data members is considered to
+have size zero, which is not something that exists in C. This includes
+a struct like `#[repr(C)] struct Foo { }`. Further, when a
+`#[repr(C)]` struct has a field whose type has zero-size, that field
+may induce padding due to its alignment, but will not otherwise affect
+the offsets of subsequent fields (as it takes up zero space).
+
 The intention is that if one has a set of C struct declarations and a
 corresponding set of Rust struct declarations, all of which are tagged
 with `#[repr(C)]`, then the layout of those structs will all be
--- a/reference/src/representation/structs-and-tuples.md
+++ b/reference/src/representation/structs-and-tuples.md
@@ -2,7 +2,7 @@
 
 **Disclaimer:** This chapter represents the consensus from issues
 [#11] and [#12].  The statements in here are not (yet) "guaranteed"
-not to change.
+not to change until an RFC ratifies them.
 
 [#11]: https://github.com/rust-rfcs/unsafe-code-guidelines/issues/11
 [#12]: https://github.com/rust-rfcs/unsafe-code-guidelines/issues/12
--- a/reference/src/representation/structs-and-tuples.md
+++ b/reference/src/representation/structs-and-tuples.md
@@ -24,15 +24,6 @@ guaranteed layout from a tuple, you are generally advised to create a
 named struct with a `#[repr(C)]` annotation (see [the section on
 structs for more details](#structs)).
 
-There is one exception: if all N fields of the tuple are of the same
-type `T` (with lifetime erased), then the tuple is guaranteed to be
-laid out as the fixed-length array type `[T; N]` (with the numbered
-tuple fields placed in the corresponding indices as expected). This
-permits such tuples to be transmuted and then indexed using an integer
-index.[^exception]
-
-[^exception]: [Proposed in this comment](https://github.com/rust-rfcs/unsafe-code-guidelines/issues/12#issuecomment-417680324).
-
 Note that the final element of a tuple (`Pn`) is marked as `?Sized` to
 permit unsized tuple coercion -- this is implemented on nightly but is
 currently unstable ([tracking issue][#42877]). In the future, we may
@@ -91,35 +82,33 @@ Structs can have various `#[repr]` flags that influence their layout:
   "as if" it were an instance of its field type when passed as an
   argument
 
-### Details *not* relevant to layout
-
-Although the compiler is free to choose the layout of structs as it
-likes, we do put some restrictions on it:
-
-- Layout must be **deterministic** -- if you compile the same program
-  twice (with the same compilation settings), the layout of the
-  structs in the program should not change.
-  - Open question: What sorts of changes can cause layout to change?
-    Do you have to make changes to the types (transitively) included
-    in the struct? Or can any sort of change suffice.
-- **Field names** are not relevant to layout: changing the name of a field
-  in a struct will never affect its layout.
-- **Privacy** is not relevant to layout: changing whether a field is
-  public or private does not affect layout, although it may affect who
-  can observe the layout (e.g., in a `#[repr(C)]` struct, if fields
-  are public, then one may find that client libraries are relying on
-  the order of fields in your struct in some fashion).
-
 ### Default layout ("repr rust")
 
-The default layout of structs is undefined and subject to change
-between individual compilations. We further do not guarantee that two
-structs with different names (but the same field types) will be laid
-out in the same way (for example, the hypothetical struct representing
-tuples ). Finally, the presence or absence of generics can make a
-difference (e.g., `struct Foo { x: u16, y: u32 }` and `struct Foo<T> {
-x: u16, y: T }` where `T = u32` are not guaranteed to be identical),
-owing to the possibility of unsizing coercions.
+The default layout of structs is not specified. Effectively, the
+compiler provdes a deterministic function per struct definition that
+defines its layout. This function may as principle take as input the
+entire input program. Therefore:
+
+- the types of the struct's fields
+- the layout of other structs (including structs not included within this struct)
+- compiler settings
+
+As of this writing, we have not reached a full consensus on what
+limitations should exist on possible field struct layouts. Therefore,
+the default layout of structs is considered undefined and subject to
+change between individual compilations. This implies that (among other
+things) two structs with the same field types may not be laid out in
+the same way (for example, the hypothetical struct representing tuples
+may be laid out differently from user-declared structs). 
+
+Further, the layout of structs is always defined relative to the
+**struct definition** plus a substitution supplying values for each of
+the struct's generic types (in contrast to just considering the fully
+monomorphized field types). This is necessary because the presence or
+absence of generics can make a difference (e.g., `struct Foo { x: u16,
+y: u32 }` and `struct Foo<T> { x: u16, y: T }` where `T = u32` are not
+guaranteed to be identical), owing to the possibility of unsizing
+coercions.
 
 **Compiler's current behavior.** As of the time of this writing, the
 compiler will reorder struct fields to minimize the overall size of
@@ -127,12 +116,45 @@ the struct (and in particular to eliminate padding due to alignment
 restrictions). The final field, however, is not reordered if an
 unsizing coercion may be applied.
 
+#### Unresolved questions
+
+During the course of the discussion in [#11] and [#12], various
+suggestions arose to limit the compiler's flexibility. These questions
+are currently considering **unresolved** and -- for each of them -- an
+issue has been opened for further discussion on the repository. This
+section documents the questions and gives a few light details, but the
+reader is referred to the issues for further discussion.
+
+**Single-field structs.** If you have a struct with single field
+(`struct Foo { x: T }`), should we guarantee that the memory layout of
+`Foo` is identical to the memory layout of `T` (note that ABI details
+around function calls may still draw a distinction, which is why
+`#[repr(transparent)]` is needed). What about zero-sized types like
+`PhantomData`?
+
+**Homogeneous structs.** If you have homogeneous structs, where all
+the `N` fields are of a single type `T`, can we guarantee a mapping to
+the memory layout of `[T; N]`? How do we map between the field names
+and the indices? What about zero-sized types?
+
+**Deterministic layout.** Can we say that layout is some deterministic
+function of a certain, fixed set of inputs? This would allow you to be
+sure that if you do not alter those inputs, your struct layout would
+not change, even if it meant that you can't predict precisely what it
+will be. For example, we might say that struct layout is a function of
+the struct's generic types and its substitutions, full stop -- this
+would imply that any two structs with the same definition are laid out
+the same. This might interfere with our ability to do profile-guided
+layout or to analyze how a struct is used and optimize based on
+that. Some would call that a feature.
+
 ### C-compatible layout ("repr C")
 
 For structs tagged `#[repr(C)]`, the compiler will apply a C-like
 layout scheme. See section 6.7.2.1 of the [C17 specification][C17] for
-a detailed write-up of what such rules entail. For most platforms,
-however, this means the following:
+a detailed write-up of what such rules entail (as well as the relevant
+specs for your platform). For most platforms, however, this means the
+following:
 
 [C17]: http://www.open-std.org/jtc1/sc22/wg14/www/abq/c17_updated_proposed_fdis.pdf
 
--- a/reference/src/representation/structs-and-tuples.md
+++ b/reference/src/representation/structs-and-tuples.md
@@ -51,9 +51,7 @@ Some related discussion:
   forth against some suffix of the struct. This RFC was not accepted,
   however. This lay out requires extra padding and seems somewhat
   surprising: it means that the layout of tuples and tuple structs
-  would diverge significantly from structs with named fields.  This
-  proposal is also incompatible with the guaranteed array layout
-  described above.
+  would diverge significantly from structs with named fields.
 
 <a name="structs"></a>
 
--- a/reference/src/representation/structs-and-tuples.md
+++ b/reference/src/representation/structs-and-tuples.md
@@ -137,7 +137,7 @@ those would have no corresponding C struct declaration -- as
 `#[repr(Rust)]` types have undefined layout, you cannot safely declare
 their layout in a C program.
 
-See also the notes on ABI compatibility under the section on `#[repr(transparent)]`.
+See also the notes on [ABI compatibility](#fnabi) under the section on `#[repr(transparent)]`.
 
 ### Fixed alignment
 
@@ -166,7 +166,9 @@ details][TRR-packed].
 
 [TRR-packed]: https://doc.rust-lang.org/stable/reference/type-layout.html#the-packed-representation
 
-### ABI Compatibility
+<a name="fnabi"> </a>
+
+### Function call ABI compatibility
 
 In general, when invoking functions that use the C ABI, `#[repr(C)]`
 structs are guaranteed to be passed in the same way as their
@@ -272,7 +274,7 @@ deterministic" layout. These areas need future exploration.
 
 ## Counteropinions and other notes
 
-@joshtrippler [argued against reordering struct
+@joshtripplet [argued against reordering struct
 fields](https://github.com/rust-rfcs/unsafe-code-guidelines/issues/11#issuecomment-417953576),
 suggesting instead it would be better if users reordering fields
 themselves. However, there are a number of downsides to such a
--- a/reference/src/representation/structs-and-tuples.md
+++ b/reference/src/representation/structs-and-tuples.md
@@ -81,9 +81,6 @@ struct Foo {
 (In fact, one may use such field names in patterns or in accessor
 expressions like `foo.0`.)
 
-Field names are not relevant to layout: changing the name of a field
-in a struct will never affect its layout.
-
 Structs can have various `#[repr]` flags that influence their layout:
 
 - `#[repr(Rust)]` -- the default.
@@ -93,6 +90,25 @@ Structs can have various `#[repr]` flags that influence their layout:
 - `#[repr(transparent)]` -- request that a "wrapper struct" be treated
   "as if" it were an instance of its field type when passed as an
   argument
+
+### Details *not* relevant to layout
+
+Although the compiler is free to choose the layout of structs as it
+likes, we do put some restrictions on it:
+
+- Layout must be **deterministic** -- if you compile the same program
+  twice (with the same compilation settings), the layout of the
+  structs in the program should not change.
+  - Open question: What sorts of changes can cause layout to change?
+    Do you have to make changes to the types (transitively) included
+    in the struct? Or can any sort of change suffice.
+- **Field names** are not relevant to layout: changing the name of a field
+  in a struct will never affect its layout.
+- **Privacy** is not relevant to layout: changing whether a field is
+  public or private does not affect layout, although it may affect who
+  can observe the layout (e.g., in a `#[repr(C)]` struct, if fields
+  are public, then one may find that client libraries are relying on
+  the order of fields in your struct in some fashion).
 
 ### Default layout ("repr rust")
 
--- a/reference/src/representation/structs-and-tuples.md
+++ b/reference/src/representation/structs-and-tuples.md
@@ -188,14 +188,6 @@ following:
 
 [^aligned]: Aligning an offset O to an alignment A means to round up the offset O until it is a multiple of the alignment A.
 
-One deviation from C comes about with "empty structs". In Rust, a
-struct that contains (transitively) no data members is considered to
-have size zero, which is not something that exists in C. This includes
-a struct like `#[repr(C)] struct Foo { }`. Further, when a
-`#[repr(C)]` struct has a field whose type has zero-size, that field
-may induce padding due to its alignment, but will not otherwise affect
-the offsets of subsequent fields (as it takes up zero space).
-
 The intention is that if one has a set of C struct declarations and a
 corresponding set of Rust struct declarations, all of which are tagged
 with `#[repr(C)]`, then the layout of those structs will all be
@@ -206,6 +198,38 @@ those would have no corresponding C struct declaration -- as
 their layout in a C program.
 
 See also the notes on [ABI compatibility](#fnabi) under the section on `#[repr(transparent)]`.
+
+**Structs with no fields.** One area where Rust layout can deviate
+from C/C++ -- even with `#[repr(C)]` -- comes about with "empty
+structs" that have no fields. In C, an empty struct declaration like
+`struct Foo { }` is illegal. However, both gcc and clang support
+options to enable such structs, and [assign them size
+zero](https://godbolt.org/z/AS2gdC). Rust behaves the same way --
+empty structs have size 0 and alignment 1 (unless an explicit
+`#[repr(align)]` is present). C++, in contrast, gives empty structs a
+size of 1, unless they are inherited from or they are fields that have
+the `[[no_unique_address]]` attribute, in which case they do not
+increase the overall size of the struct.
+
+**Structs of zero-size.** It is also possible to have structs that
+have fields but have non-zero size. In this case, the size of the
+struct would be zero, but its alignment may be greater. For example,
+`#[repr(C)] struct Foo { x: [u16; 0] }` would have an alignment of 2
+bytes by default. ([This matches the behavior in gcc and
+clang](https://godbolt.org/z/5w0gkq).)
+
+**Structs with fields of zero-size.** If a `#[repr(C)]` struct
+containing a field of zero-size, that field does not occupy space in
+the struct; it can affect the offsets of subsequent fields if it
+induces padding due to the alignment on its type. ([This matches the
+behavior in gcc and clang](https://godbolt.org/z/5w0gkq).)
+
+**C++ compatibility hazard.** As noted above when discussing structs
+with no fields, C++ treats empty structs like `struct Foo { }`
+differently from C and Rust. This can introduce subtle compatibility
+hazards. If you have an empty struct in your C++ code and you make the
+"naive" translation into Rust, even tagging with `#[repr(C)]` will not
+produce layout- or ABI-compatible results.
 
 ### Fixed alignment
 
--- a/reference/src/representation/structs-and-tuples.md
+++ b/reference/src/representation/structs-and-tuples.md
@@ -1,0 +1,258 @@
+# Representation of structs and tuples
+
+**Disclaimer:** This chapter represents the consensus from issues
+[#11] and [#12].  The statements in here are not (yet) "guaranteed"
+not to change.
+
+[#11]: https://github.com/rust-rfcs/unsafe-code-guidelines/issues/11
+[#12]: https://github.com/rust-rfcs/unsafe-code-guidelines/issues/12
+
+## Tuple types
+
+In general, an anonymous tuple type `(T1..Tn)` of arity N is laid out
+"as if" there were a corresponding tuple struct declared in libcore:
+
+```rust
+#[repr(Rust)]
+struct TupleN<P1..Pn>(P1..Pn);
+```
+
+In this case, `(T1..Tn)` would be compatible with `TupleN<T1..Tn>`.
+As discussed below, this generally means that the compiler is **free
+to re-order field layout** as it wishes. Thus, if you would like a
+guaranteed layout from a tuple, you are generally advised to create a
+named struct with a `#[repr(C)]` annotation (see [the section on
+structs for more details](#structs)).
+
+There is one exception: if all N fields of the tuple are of the same
+type `T` (with lifetime erased), then the tuple is guaranteed to be
+laid out as the fixed-length array type `[T; N]` (with the numbered
+tuple fields placed in the corresponding indices as expected). This
+permits such tuples to be transmuted and then indexed using an integer
+index.[^exception]
+
+[^exception]: [Proposed in this comment](https://github.com/rust-rfcs/unsafe-code-guidelines/issues/12#issuecomment-417680324).
+
+### Other notes on tuples
+
+Some related discussion:
+
+- [RFC #1582](https://github.com/rust-lang/rfcs/pull/1582) proposed
+  that tuple structs should have a "nested representation", where
+  e.g. `(T1, T2, T3)` would in fact be laid out as `(T1, (T2,
+  T3))`. The purpose of this was to permit variadic matching and so
+  forth against some suffix of the struct. This RFC was not accepted,
+  however. This lay out requires extra padding and seems somewhat
+  surprising: it means that the layout of tuples and tuple structs
+  would diverge significantly from structs with named fields.  This
+  proposal is also incompatible with the guaranteed array layout
+  described above.
+
+<a name="structs"></a>
+
+## Struct types
+
+Structs come in two principle varieties:
+
+```rust
+// Structs with named fields
+struct Foo { f1: T1, .., fn: Tn }
+
+// Tuple structs
+struct Foo(T1, .., Tn)
+```
+
+In general, tuple structs can be understood as equivalent to named
+structs declared with the same order; therefore, the two declarations
+of `Foo` in the previous example are treated equivalently.
+
+Structs can have various `#[repr]` flags that influence their layout:
+
+- `#[repr(Rust)]` -- the default.
+- `#[repr(C)]` -- request C compatibility
+- `#[repr(align(N))]` -- specify the alignment
+- `#[repr(packed)]` -- request packed layout where fields are not internally aligned
+- `#[repr(transparent)]` -- request that a single-field struct be
+  treated "as if" it were an instance of its field type when passed as
+  an argument
+
+### Default layout ("repr rust")
+
+The default layout of structs is undefined and subject to change
+between compiler revisions. We further do not guarantee that two
+structs with different names (but the same field types) will be laid
+out in the same way (for example, the hypothetical struct representing
+tuples ). Finally, the presence or absence of generics can make a
+difference (e.g., `struct Foo { x: u16, y: u32 }` and `struct Foo<T> {
+x: u16, y: T }` where `T = u32` are not guaranteed to be identical),
+owing to the possibility of unsizing coercions.
+
+**Compiler's current behavior.** As of the time of this writing, the
+compiler will reorder struct fields to minimize the overall size of
+the struct (and in particular to eliminate padding due to alignment
+restrictions). The final field, however, is not reordered if an
+unsizing coercion may be applied.
+
+### C-compatible layout ("repr C")
+
+For structs tagged `#[repr(C)]`, the compiler will apply a C-like
+layout scheme (see section 6.7.2.1 of the [C17 specification][C17] for
+a detailed write-up):
+
+[C17]: http://www.open-std.org/jtc1/sc22/wg14/www/abq/c17_updated_proposed_fdis.pdf
+
+- Field order is preserved.
+- The first field begins at offset 0.
+- Assuming the struct is not packed, each field's offset is aligned to
+  the natural alignment for that field's type, possibly creating
+  unused padding bits.
+- The total size of the struct is rounded up to its overall alignemnt.  
+
+The intention is that if one has a set of C struct declarations and a
+corresponding set of Rust struct declarations, all of which are tagged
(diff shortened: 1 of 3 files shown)


reference/src/representation/structs-and-tuples.md
```diff
@@ -0,0 +1,258 @@
...
+structs declared with the same order; therefore, the two declarations
+of `Foo` in the previous example are treated equivalently.
+
+Structs can have various `#[repr]` flags that influence their layout:
+
+- `#[repr(Rust)]` -- the default.
+- `#[repr(C)]` -- request C compatibility
+- `#[repr(align(N))]` -- specify the alignment
+- `#[repr(packed)]` -- request packed layout where fields are not internally aligned
+- `#[repr(transparent)]` -- request that a single-field struct be
```
eddyb: I don't think "single-field" here is not accurate. cc @rkruppe

IIRC, it's more like "single data field and any number of ZSTs".

reference/src/representation/structs-and-tuples.md
```diff
@@ -0,0 +1,258 @@
...
+layout scheme (see section 6.7.2.1 of the [C17 specification][C17] for
+a detailed write-up):
+
+[C17]: http://www.open-std.org/jtc1/sc22/wg14/www/abq/c17_updated_proposed_fdis.pdf
+
+- Field order is preserved.
+- The first field begins at offset 0.
+- Assuming the struct is not packed, each field's offset is aligned to
+  the natural alignment for that field's type, possibly creating
+  unused padding bits.
```
eddyb: I believe this is a misuse of "natural alignment". AFAIK "natural alignment" specifically means `align_of::<T>() == size_of::<T>()` (and mainly makes sense for scalars/vectors, not arbitrary types).


You can probably use "(ABI-mandated) alignment (i.e. what `align_of::<T>` returns)".

reference/src/representation/structs-and-tuples.md
```diff
@@ -0,0 +1,258 @@
...
+
+For structs tagged `#[repr(C)]`, the compiler will apply a C-like
+layout scheme (see section 6.7.2.1 of the [C17 specification][C17] for
+a detailed write-up):
+
+[C17]: http://www.open-std.org/jtc1/sc22/wg14/www/abq/c17_updated_proposed_fdis.pdf
+
+- Field order is preserved.
+- The first field begins at offset 0.
+- Assuming the struct is not packed, each field's offset is aligned to
```
eddyb: Do we define "aligned" (as "multiple of alignment") anywhere else?

reference/src/representation/structs-and-tuples.md
```diff
@@ -0,0 +1,258 @@
...
+
+### Packed layout
+
+The `#[repr(packed)]` attribute may be used to remove all padding from
+the struct. The resulting fields may not fall at properly aligned
+boundaries in memory. This makes it unsafe to create a Rust reference
+(`&T` or `&mut T`) to those fields, as the compiler requires that all
+reference values must always be aligned (so that it can use more
+efficient load/store instructions at runtime). See the [Rust reference
+for more details][TRR-packed].
```
eddyb: Should mention `#[repr(packed(N))]` (or is it `#[repr(pack(N))]`?) - in general, packing doesn't "remove all padding", but imposes a maximum limit on the alignment guaranteed of each fields.

E.g. in a `#[repr(pack(4))]` struct, fields with alignment `1`, `2` or `4` are placed as usual, at an offset multiple of that alignment, but fields with *larger* alignment than `4` don't get anything better than a multiple of `4`.

reference/src/representation/structs-and-tuples.md
```diff
@@ -0,0 +1,258 @@
...
+annotations that could be added
+([1](https://github.com/rust-rfcs/unsafe-code-guidelines/issues/11#issuecomment-420077105),
+[2](https://github.com/rust-rfcs/unsafe-code-guidelines/issues/11#issuecomment-420077105)),
+such that the knowledge of the improved layouts can be recorded in the
+source.
+
+As a more declarative alternative, @alercah [proposed a possible
+extension](https://github.com/rust-rfcs/unsafe-code-guidelines/issues/12#issuecomment-420165155)
+that would permit one to declare that the layout of two structs or
+types are compatible (e.g., `#[repr(as(Foo))] struct Bar { .. }`),
```
eddyb: If you can *name* the struct you want to be compatible with, why not newtype it with `#[repr(transparent)]`?

reference/src/representation/structs-and-tuples.md
```diff
@@ -0,0 +1,258 @@
...
+[#12]: https://github.com/rust-rfcs/unsafe-code-guidelines/issues/12
+
+## Tuple types
+
+In general, an anonymous tuple type `(T1..Tn)` of arity N is laid out
+"as if" there were a corresponding tuple struct declared in libcore:
+
+```rust
+#[repr(Rust)]
+struct TupleN<P1..Pn>(P1..Pn);
```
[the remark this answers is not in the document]
gnzlbg: > Could add Pn: ?Sized

Does this work already? If not, I'd prefer to open an issue so that we don't forget about this particular extension (allowing the last element of a struct / tuple / ... to be unsized), but to leave it out from the document for now.

reference/src/representation/structs-and-tuples.md
```diff
@@ -0,0 +1,258 @@
...
+// Structs with named fields
+struct Foo { f1: T1, .., fn: Tn }
+
+// Tuple structs
+struct Foo(T1, .., Tn)
+```
+
+In general, tuple structs can be understood as equivalent to named
+structs declared with the same order; therefore, the two declarations
+of `Foo` in the previous example are treated equivalently.
```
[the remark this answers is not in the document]
nikomatsakis: Yes, this is what I was trying to convey, essentially.

reference/src/representation/structs-and-tuples.md
nikomatsakis: Thanks, I meant to come back and look into what the various "kinds" of alignment were.

reference/src/representation/structs-and-tuples.md
nikomatsakis: No, do you think we should? I can add a footnote or something. I'm trying not to make this an "unreadably dense" specification.

reference/src/representation/structs-and-tuples.md
nikomatsakis: I imagine because you may not have access to its private fields at this point in the code, but you also control its definition (e.g., maybe it's in another crate).

reference/src/representation/structs-and-tuples.md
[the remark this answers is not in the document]
nikomatsakis: Done

reference/src/representation/structs-and-tuples.md
nikomatsakis: Added a definition in a footnote

reference/src/representation/structs-and-tuples.md
nikomatsakis: OK, I wasn't aware of that generalization.

nikomatsakis:
Pushed an update containing the various suggestions from @eddyb.

reference/src/representation/structs-and-tuples.md
nikomatsakis: I added some text to this effect, thanks.

reference/src/representation/structs-and-tuples.md
```diff
@@ -0,0 +1,293 @@
...
+corresponding set of Rust struct declarations, all of which are tagged
+with `#[repr(C)]`, then the layout of those structs will all be
+identical. Note that this setup implies that none of the structs in
+question can contain any `#[repr(Rust)]` structs (or Rust tuples), as
+those would have no corresponding C struct declaration -- as
+`#[repr(Rust)]` types have undefined layout, you cannot safely declare
+their layout in a C program.
+
+See also the notes on ABI compatibility under the section on `#[repr(transparent)]`.
+
```
gnzlbg: Maybe it might be worth spelling out which properties of `repr(C)` are platform dependent (implementation defined) and cannot be relied on by portable code.

reference/src/representation/structs-and-tuples.md
gnzlbg: Maybe we should open an issue to discuss this on its own merits. This feels like a escape hatch that could be used to subvert privacy, but in a more controlled way than... `#define private public` or similar in C++.

reference/src/representation/structs-and-tuples.md
```diff
@@ -0,0 +1,293 @@
...
+`#[repr(transparent)]` cannot be applied to *any* struct. It is
+limited to structs with a single field whose type `T` has non-zero
+size, along with some number of other fields whose types are all
+zero-sized (typically `std::marker::PhantomData` fields). The struct
+then takes on the "ABI behavior" of the type `T` that has non-zero
+size.
+
+(Note further that the Rust ABI is undefined and theoretically may
+vary from compiler revision to compiler revision.)
+
```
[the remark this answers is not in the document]
ubsan: Yes

reference/src/representation/structs-and-tuples.md
ubsan: What are you asking about? I'm not sure, beyond "the members will be layed out in program order" and "a pointer to a struct has the same value (representation) as a pointer to its first member", one could even ask for in a non-platform-dependent program.

reference/src/representation/structs-and-tuples.md
eddyb: I was asking because it seems *more* useful if there's no dependency-dependent relationship between the two crates.

reference/src/representation/structs-and-tuples.md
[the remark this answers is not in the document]
gnzlbg: Nevermind, this only applies to structs and tuples, not `enum`s. What I was thinking about is that `Option` is not `repr(C)` currently, so I was wondering how this would work for non `repr(C)` types, but that should be discussed in the chapter about `enum`s instead of here.

reference/src/representation/structs-and-tuples.md
gnzlbg: I meant properties like this:

> Assuming the struct is not packed, each field's offset is aligned[^aligned] to
  the ABI-mandated alignment for that field's type, possibly creating
  unused padding bits.

For example, I can hardcode the address offset of the first field w.r.t. to a value of the struct type to `0`, and that will work on all platforms. But doing the same for any other field, even if it is correct for the platform being targeted, is not portable in general.

reference/src/representation/structs-and-tuples.md
```diff
@@ -0,0 +1,293 @@
...
+guaranteed layout from a tuple, you are generally advised to create a
+named struct with a `#[repr(C)]` annotation (see [the section on
+structs for more details](#structs)).
+
+There is one exception: if all N fields of the tuple are of the same
+type `T` (with lifetime erased), then the tuple is guaranteed to be
+laid out as the fixed-length array type `[T; N]` (with the numbered
+tuple fields placed in the corresponding indices as expected). This
+permits such tuples to be transmuted and then indexed using an integer
+index.[^exception]
```
alercah: Is this an exception specific to tuples, or should it apply also to tuple-like structs?

reference/src/representation/structs-and-tuples.md
```diff
@@ -0,0 +1,293 @@
...
+  0: T1,
+  ...
+  n-1: Tn
+}
+```
+
+(In fact, one may use such field names in patterns or in accessor
+expressions like `foo.0`.)
+
+Field names are not relevant to layout: changing the name of a field
```
alercah: This is an interesting assertion that, while I understand it is valuable to make, I am not entirely sure it is a guarantee we do want to make, given the right reserved to the compiler to reorder structs at will.

reference/src/representation/structs-and-tuples.md
```diff
@@ -0,0 +1,293 @@
...
+out in the same way (for example, the hypothetical struct representing
+tuples ). Finally, the presence or absence of generics can make a
+difference (e.g., `struct Foo { x: u16, y: u32 }` and `struct Foo<T> {
+x: u16, y: T }` where `T = u32` are not guaranteed to be identical),
+owing to the possibility of unsizing coercions.
+
+**Compiler's current behavior.** As of the time of this writing, the
+compiler will reorder struct fields to minimize the overall size of
+the struct (and in particular to eliminate padding due to alignment
+restrictions). The final field, however, is not reordered if an
```
alercah: Would it be simpler to say "An unsized field always comes last" or the like?

reference/src/representation/structs-and-tuples.md
```diff
@@ -0,0 +1,293 @@
...
+- `#[repr(align(N))]` -- specify the alignment
+- `#[repr(packed)]` -- request packed layout where fields are not internally aligned
+- `#[repr(transparent)]` -- request that a "wrapper struct" be treated
+  "as if" it were an instance of its field type when passed as an
+  argument
+
+### Default layout ("repr rust")
+
+The default layout of structs is undefined and subject to change
+between compiler revisions. We further do not guarantee that two
```
alercah: Between individual compilations, no? I think that is what we had determined was the line.

reference/src/representation/structs-and-tuples.md
```diff
@@ -0,0 +1,293 @@
...
+
+The `#[repr(align(N))]` attribute may be used to raise the alignment
+of a struct, as described in [The Rust Reference][TRR-align].
+
+[TRR-align]: (https://doc.rust-lang.org/stable/reference/type-layout.html#the-align-representation).
+
+### Packed layout
+
+The `#[repr(packed(N))]` attribute may be used to impose a maximum
+limit on the alignments for individual fields. It is most commonly
```
alercah: This should probably mention that, unless `repr(align)` is also present, the resulting struct will have alignment no greater than `N`.

reference/src/representation/structs-and-tuples.md
alercah: I think the reason I was thinking of that at the time is because `repr(transparent)` adds an indirection overhead through the extra `.0` field, and doesn't let you rename individual fields. But there is a broader space of "newtypes are painful to use" that may be worth doing separate work in that subsumes this.

You may also have individual fields which are not of identical type, but are layout-compatible.

But yeah, `transparent` may well be all we need.

reference/src/representation/structs-and-tuples.md
gnzlbg: The text defines tuples as

> "as if" there were a corresponding tuple struct declared in libcore

so this applies to tuple structs as well.

nikomatsakis: Actually, the text was not meant to imply that. In fact, this was saying that this is **an exception** to tuples being laid out as structs. 

That said, we may want to commit to the same thing more generally. I would however be opposed to making it specific to "tuple structs" but not "named field structs". If we're going to do this thing, let's do it uniformly.

reference/src/representation/structs-and-tuples.md
nikomatsakis: True. Maybe I'll just cut it out. It seems like a random guarantee to give -- in other words, we're not saying what information the compiler **can use**, we're just identifying one random thing it **cannot**...

reference/src/representation/structs-and-tuples.md
nikomatsakis: It wouldn't really be sufficient. The field begins life with a sized type.

nikomatsakis: e.g., when you create the struct, you know the full type, but you later coerce a pointer to that struct into a (fat) pointer that doesn't know the full type.

reference/src/representation/structs-and-tuples.md
nikomatsakis: OK.

reference/src/representation/structs-and-tuples.md
nikomatsakis: I thought I saw that it was an error to combine the two?

reference/src/representation/structs-and-tuples.md
eddyb: Also, we'll likely relax the rules around unsized types in the future.

rkruppe:
Long ago I proposed that we might want to guarantee (some subset of) newtype unpacking for repr(Rust) structs. @nikomatsakis carried this over into #11 as discussion point but it received no further discussion. I like to think that means it's uncontroversial 😄 I've also never heard of any reason why one might not want that to be true.

To make a specific proposal, let's restrict it to structs [1] that contain a single field having the same *memory layout* as the type of the sole field. So `struct Foo<T>(T);` and `struct Foo<T> { x: T }` would be laid out like `T` in memory, though possibly still passed differently in function calls.


[1] The same guarantee for `(T,)` are already covered by the special case of homogeneous tuples being laid out like arrays that is already in this PR.

the8472:
@rkruppe for `(T,)` the `.0` field is public, in `Foo<T>` it is not. The members of the tuple are guaranteed to never change for that type but since Foo has private fields it is a non-breaking change to add new fields which would remove the layout guarantee.

So shouldn't that guarantee be restricted to structs with a single public field? Or at least there should be a lint if such transmutations are used by a crate that does not own `Foo`.

rkruppe:
Visibility should not affect layout, only who can rely on the layout.

the8472:
That might be worth documenting.

Still, it might be better to document the intent of such structs with `repr(transparent)`.

reference/src/representation/structs-and-tuples.md
```diff
@@ -0,0 +1,293 @@
...
+extension](https://github.com/rust-rfcs/unsafe-code-guidelines/issues/12#issuecomment-420165155)
+that would permit one to declare that the layout of two structs or
+types are compatible (e.g., `#[repr(as(Foo))] struct Bar { .. }`),
+thus permitting safe transmutes (and also ABI compatibility).  One
+might also use some weaker form of `#[repr(C)]` to specify a "more
+deterministic" layout. These areas need future exploration.
+
+## Counteropinions and other notes
+
+@joshtrippler [argued against reordering struct
```
RalfJung: `s/joshtrippler/joshtriplett/`

reference/src/representation/structs-and-tuples.md
nikomatsakis: Actually, I would like to push back on this slightly — there is a general desire to ensure that th compiler output is **deterministic**. This is not 100% true but it is very nearly true, and we would like to to be true. This seems to imply that, so long as the input does not change, the layout cannot change. I am not sure why we would need to lose that guarantee.

reference/src/representation/structs-and-tuples.md
nikomatsakis: Updated.

reference/src/representation/structs-and-tuples.md
```diff
@@ -0,0 +1,293 @@
...
+- [RFC #1582](https://github.com/rust-lang/rfcs/pull/1582) proposed
+  that tuple structs should have a "nested representation", where
+  e.g. `(T1, T2, T3)` would in fact be laid out as `(T1, (T2,
+  T3))`. The purpose of this was to permit variadic matching and so
+  forth against some suffix of the struct. This RFC was not accepted,
+  however. This lay out requires extra padding and seems somewhat
+  surprising: it means that the layout of tuples and tuple structs
+  would diverge significantly from structs with named fields.  This
+  proposal is also incompatible with the guaranteed array layout
+  described above.
```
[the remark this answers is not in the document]
nikomatsakis: I suppose that's true. I will revise.

reference/src/representation/structs-and-tuples.md
nikomatsakis: I don't have a strong opinion either way. I have left it, however, because I agree with @RalfJung that people should be able to be confident that changing the names of their fields won't affect their program's memory layout (even if they don't know what that layout is).

reference/src/representation/structs-and-tuples.md
nikomatsakis: extracted to an "unresolved question"

reference/src/representation/structs-and-tuples.md
nikomatsakis: extracted to an "unresolved question"

nikomatsakis:
I have made a number of edits addressing various concerns. The most significant is that I pulled out some controversial topics and marked them as "Unresolved Questions". I plan to open issues for each of them to continue the conversation. I am marking this with the `final-comment-period` label — we plan the merge the PR next meeting unless major controversy arises.

nikomatsakis:
Here is the set of unresolved questions:

**Zero-sized structs ([#37]).** If you have a struct which --
transitively -- contains no data of non-zero size, then the size of
that struct will be zero as well. These zero-sized structs appear
frequently as exceptions in other layout considerations (e.g.,
single-field structs). An example of such a struct is
`std::marker::PhantomData`.

[#37]: https://github.com/rust-rfcs/unsafe-code-guidelines/issues/37

**Single-field structs ([#34]).** If you have a struct with single field
(`struct Foo { x: T }`), should we guarantee that the memory layout of
`Foo` is identical to the memory layout of `T` (note that ABI details
around function calls may still draw a distinction, which is why
`#[repr(transparent)]` is needed). What about zero-sized types like
`PhantomData`?

[#34]: https://github.com/rust-rfcs/unsafe-code-guidelines/issues/34

**Homogeneous structs ([#36]).** If you have homogeneous structs, where all
the `N` fields are of a single type `T`, can we guarantee a mapping to
the memory layout of `[T; N]`? How do we map between the field names
and the indices? What about zero-sized types?

[#36]: https://github.com/rust-rfcs/unsafe-code-guidelines/issues/36

**Deterministic layout ([#35]).** Can we say that layout is some deterministic
function of a certain, fixed set of inputs? This would allow you to be
sure that if you do not alter those inputs, your struct layout would
not change, even if it meant that you can't predict precisely what it
will be. For example, we might say that struct layout is a function of
the struct's generic types and its substitutions, full stop -- this
would imply that any two structs with the same definition are laid out
the same. This might interfere with our ability to do profile-guided
layout or to analyze how a struct is used and optimize based on
that. Some would call that a feature.

reference/src/representation/structs-and-tuples.md
```diff
@@ -0,0 +1,349 @@
...
+- `#[repr(transparent)]` -- request that a "wrapper struct" be treated
+  "as if" it were an instance of its field type when passed as an
+  argument
+
+### Default layout ("repr rust")
+
+The default layout of structs is not specified. Effectively, the
+compiler provdes a deterministic function per struct definition that
+defines its layout. This function may as principle take as input the
+entire input program. Therefore:
```
rkruppe: I am not sure this wording reserves the freedoms it want to reserve, and even if someone argued it does I would like it to be clarified. Since we did not get consensus to rule out profile-guided layout, the *program source code* is not all that informs layout. Not even if that includes build scripts, input data files, etc. that are used during the profiling run, since the program may not be deterministic w.r.t. these (e.g. it might have race conditions or depend on the system time or ...). So I don't think we can guarantee anything involving the words "deterministic function" and have to stick to something like "every time you invoke the compiler you may get a completely different layout".

gnzlbg: > Since we did not get consensus to rule out profile-guided layout

I might have missed this, but did you managed to write your thoughts about this down?

rkruppe: No, not yet.

reference/src/representation/structs-and-tuples.md
```diff
@@ -130,8 +130,9 @@ unsizing coercion may be applied.
 ### C-compatible layout ("repr C")
 
 For structs tagged `#[repr(C)]`, the compiler will apply a C-like
-layout scheme (see section 6.7.2.1 of the [C17 specification][C17] for
-a detailed write-up):
+layout scheme. See section 6.7.2.1 of the [C17 specification][C17] for
+a detailed write-up of what such rules entail. For most platforms,
+however, this means the following:
```
gnzlbg: Maybe we should just go one step further here and just state that the "most recent" C specification applies. It would be a pain to have to update this every N years, history has shown that the newer C specs are backwards compatible with the old ones (e.g. the C99, C11, and C17 specs have never introduced breaking changes here - they just have defined behavior that was undefined before), and we probably want to be as compatible as possible with C anyways. 

This kinds of tides Rust with the C spec, but that just means that the Rust language would be just another stakeholder in the C standardization process, just like C++ is. They don't really meat that frequently, but we should probably be sending someone there anyways to make sure that the interest of Rust are represented.

reference/src/representation/structs-and-tuples.md
```diff
@@ -0,0 +1,349 @@
...
+### Default layout ("repr rust")
+
+The default layout of structs is not specified. Effectively, the
+compiler provdes a deterministic function per struct definition that
+defines its layout. This function may as principle take as input the
+entire input program. Therefore:
+
+- the types of the struct's fields
+- the layout of other structs (including structs not included within this struct)
+- compiler settings
```
RalfJung: The enumeration starts with "Therefore" but doesn't go on matching that.

reference/src/representation/structs-and-tuples.md
```diff
@@ -0,0 +1,349 @@
...
+
+- In a generic struct, the [best ordering of fields may not be known
+  ahead of
+  time](https://github.com/rust-rfcs/unsafe-code-guidelines/issues/11#issuecomment-420659840),
+  so the user cannot do it manually.
+- If layout is defined, then it becomes part of your API, such taht
+  reordering fields is a breaking change for your clients (if we
+  consider unsafe code that may rely on the layout, then this applies
+  [even to structs with named
+  fields](https://github.com/rust-rfcs/unsafe-code-guidelines/issues/11#issuecomment-420117856).
```
RalfJung: Implementation details like struct layout are not part of your API and don't have to be semver-maintained *unless you say so*. unsafe code can observe all sorts of things, as there is no way to guarantee compatibility with that.

So, I disagree with the implication stated here.

reference/src/representation/structs-and-tuples.md
```diff
@@ -0,0 +1,349 @@
...
+fields](https://github.com/rust-rfcs/unsafe-code-guidelines/issues/11#issuecomment-417953576),
+suggesting instead it would be better if users reordering fields
+themselves. However, there are a number of downsides to such a
+proposal (and -- further -- it does not match our existing behavior):
+
+- In a generic struct, the [best ordering of fields may not be known
+  ahead of
+  time](https://github.com/rust-rfcs/unsafe-code-guidelines/issues/11#issuecomment-420659840),
+  so the user cannot do it manually.
+- If layout is defined, then it becomes part of your API, such taht
```
RalfJung: Typo: s/taht/that/

reference/src/representation/structs-and-tuples.md
```diff
@@ -0,0 +1,349 @@
...
+
+some_function(Foo { x: 22 }); // Bad!
+```
+
+Instead, you should declare the struct with `#[repr(transparent)]`,
+which specifies that `Foo` should use the ABI rules for its field
+type, `u32`. This is useful when using "wrapper structs" in Rust to
+give stronger typing guarantees.
+
+`#[repr(transparent)]` cannot be applied to *any* struct. It is
```
RalfJung: I might be reading this wrong (English being a foreign language and all), but to me "cannot be applied to any struct" means "cannot ever be applied, no matter the struct". Shouldn't this say "cannot be applied to *every* struct"?

reference/src/representation/structs-and-tuples.md
rkruppe: Fields are listed in declaration order in rustdoc, so *if all fields are public* and the struct isn't `non_exhaustive` it seems rather reasonable for a client to assume that their relative order is intended to be exposed (just as the number and names and types of public fields is already implicitly part of the API). One can certainly argue that they shouldn't but I don't think it's realistic to establish such a convention at this point, and since the threat here is UB rather than just code no longer compiling after a dependency bump, it seems risky to try.

gnzlbg: > One can certainly argue that they shouldn't but I don't think it's realistic to establish such a convention at this point,

Are you talking here about `repr(Rust)` structs as well? As in, the compiler is able to reorder their fields, except if all fields are `pub` ? That would be a bit of subtle knowledge that one would need to know about the usage of `pub` on struct fields. I think I would prefer a more explicit way of telling the consumers of a struct that they can rely on the relative order of the fields, like `repr(C)`.

rkruppe: This is in the context of the part of the text we're commenting on here, i.e., we are talking about a scenario where we guarantee the layout of all repr(Rust) structs. The question is then, when will consumers of another library's type be allowed to rely on the declaration order of the fields: implicitly whenever the fields are public, or only if it is explicitly guaranteed in the documentation? Certainly if fields are private, there's no reasonable expectation about their presence let alone relative order, but for reasons outlined above I think it would be user-hostile to say "the docs tell you which order these public fields are declared in but you need to forget about this and not use that knowledge".

reference/src/representation/structs-and-tuples.md
```diff
@@ -0,0 +1,349 @@
...
+
+[#35]: https://github.com/rust-rfcs/unsafe-code-guidelines/issues/35
+
+### C-compatible layout ("repr C")
+
+For structs tagged `#[repr(C)]`, the compiler will apply a C-like
+layout scheme. See section 6.7.2.1 of the [C17 specification][C17] for
+a detailed write-up of what such rules entail (as well as the relevant
+specs for your platform). For most platforms, however, this means the
+following:
```
RalfJung: AFAIK C does not have zero-sized types, so the standard doesn't say anything about them. Should we say explicitly that zero-sized types never affect the offsets of the other fields, and if moreover their alignment is 1 then they will not affect alignment and size of the entire struct either?

gnzlbg: Sounds reasonable to me.

rkruppe: A ZST with alignment > 1 can affect the offset of following fields, assuming we don't reorder the ZST fields (which seems like the most faithful extension of C layout), e.g. `{ u8, ZSTWithAlign4, u8 }` should have the second u8 at offset 4.

reference/src/representation/structs-and-tuples.md
nikomatsakis: I will rewrite to clarify.

reference/src/representation/structs-and-tuples.md
rkruppe: To be clear: it is certainly possible to define PGO traces as part of the compiler input (beyond just sources and compiler flags), though it's a bit of a stretch IMO. That should be called out explicit in the text, though. What's currently written here could reasonably be read as saying the layout depends just on the source code and flags such as `-C opt-level`.

reference/src/representation/structs-and-tuples.md
nikomatsakis: Yes, my phrasing was poor. There is a way of reading the text that makes it correct, but it relies on intonation to get the meaning across -- sort of "cannot be applied to *just any* struct".

reference/src/representation/structs-and-tuples.md
rkruppe: Ah, the updated version is good in that regard, thanks!

reference/src/representation/structs-and-tuples.md
nikomatsakis: Good point, I have forgotten about ZST and C. It does seem a bit strange, doesn't it, that [this struct has size 0](https://play.rust-lang.org/?version=stable&mode=debug&edition=2015&gist=6abbb8dd0813670b5c465e16d5ccc825)?

```rust
#[repr(C)]
struct Foo { }
```

Probably not a major issue in practice.

nikomatsakis:
I pushed some updates. The most interesting thing has to do with zero-sized structs. I added the following text to the `#[repr(C)]` section:

> One deviation from C comes about with "empty structs". In Rust, a struct that contains (transitively) no data members is considered to have size zero, which is not something that exists in C. This includes a struct like `#[repr(C)] struct Foo { }`. Further, when a `#[repr(C)]` struct has a field whose type has zero-size, that field may induce padding due to its alignment, but will not otherwise affect the offsets of subsequent fields (as it takes up zero space).

reference/src/representation/structs-and-tuples.md
nikomatsakis: I pushed a commit documenting current behavior, so I'll mark this conversation as **resolved**.

reference/src/representation/structs-and-tuples.md
```diff
@@ -0,0 +1,370 @@
...
+
+[^aligned]: Aligning an offset O to an alignment A means to round up the offset O until it is a multiple of the alignment A.
+
+One deviation from C comes about with "empty structs". In Rust, a
+struct that contains (transitively) no data members is considered to
+have size zero, which is not something that exists in C. This includes
+a struct like `#[repr(C)] struct Foo { }`. Further, when a
+`#[repr(C)]` struct has a field whose type has zero-size, that field
+may induce padding due to its alignment, but will not otherwise affect
+the offsets of subsequent fields (as it takes up zero space).
```
RalfJung: As @rkruppe noted on Zulip, this seems like a problem. It means that "copy-pasting `struct` definitions and adding `repr(C)` everywhere" does not give you C compatibility, because your `Foo` would actually take space when put in a larger struct in `C`.

This seems like a bug, TBH. I am not sure if it is a bug that we can still fix. Might be worth having at least a warning.

nikomatsakis: Yes — I was thinking the same in that conversation. That is, "bug and not clearly a bug we can fix", which does suggest that at least a lint is warranted.

nikomatsakis: For the benefit of others, the [Zulip conversation was here](https://rust-lang.zulipchat.com/#narrow/stream/136281-wg-unsafe-code-guidelines/subject/how.20much.20determinism/near/136350308).

ubsan: That's not actually how C works - C does not allow empty structs. The godbolt link was from C++, which does allow empty structs, and in which empty structs have size one.

rkruppe: Transcribing another thing @ubsan noted on Zulip: empty structs are accepted as an extension by some C compilers, but (at least) GCC and Clang make them have size zero, unlike C++. Example: https://godbolt.org/z/AS2gdC

RalfJung: Oh wow. So this same data structure behaves different in C and C++.

Seems like the most sensible choice then is to keep size 0 in Rust (consistent with the rest of Rust, and with C in at least GCC and Clang), but still have a lint about the difference to C++?

ubsan: Allow-by-default at best - 0 size structures are weird in C++, and usually you'd use either EBO or `[[no_unique_address]]` with them.

RalfJung: Isn't them being weird another argument for making this a warn-by-default lint?  I expect many people will not know this.  I am not a C++ expert, but I have programmed in C++ for many years and never heard about this; I don't think we can expect everybody doing C++ FFI to know about these issues.

gnzlbg: So to summarize: C does not allow empty structs, some C language extensions allow empty structs with `sizeof == 0`, C++ does allow empty structs but these have a `sizeof == 1` unless they are inherited from (in which case EBO guarantees that they won't increase the size of the struct) or they are fields that have the `[[no_unique_address]]` attribute. 

I think `#[repr(C)]` should warn on this when it makes a difference by default, that is, when the ZST would change the layout. We could have an opt-in warning that always warns on ZST being used in `#[repr(C)]` but I fear that would be extremely noisy for little win. 

About the situation with the C-language extension and C++ it appears that `#[repr(C)] != #[repr(C++)]`, so maybe we just need to add new `repr`s to deal with those.

RalfJung: I had to google EBO: [Empty base optimization](https://en.cppreference.com/w/cpp/language/ebo).

gnzlbg: @RalfJung note that EBO is guaranteed by the C++>=11 standard for types with [standard layout](https://en.cppreference.com/w/cpp/named_req/StandardLayoutType), like empty structs: `struct T {};`, so it is a required layout optimization.

nikomatsakis:
@avadacatavra my hope was that we will merge it today

nikomatsakis:
OK, I attempted to summarize [the conversation from here](https://github.com/rust-rfcs/unsafe-code-guidelines/pull/31#discussion_r227808315) in the latest commit and added some appropriate warning language.

reference/src/representation/structs-and-tuples.md
```diff
@@ -207,6 +199,38 @@ their layout in a C program.
...
+options to enable such structs, and [assign them size
+zero](https://godbolt.org/z/AS2gdC). Rust behaves the same way --
+empty structs have size 0 and alignment 1 (unless an explicit
+`#[repr(align)]` is present). C++, in contrast, gives empty structs a
+size of 1, unless they are inherited from or they are fields that have
+the `[[no_unique_address]]` attribute, in which case they do not
+increase the overall size of the struct.
+
+**Structs of zero-size.** It is also possible to have structs that
+have fields but have non-zero size. In this case, the size of the
```
gnzlbg: typo: non-zero => zero

ubsan:
It looks like gcc (the only compiler which has implemented `[[no_unique_address]]`), has chosen:
  - to put `[[no_unique_address]]` members at the top of the class holding them
    - (see https://gcc.godbolt.org/z/Q4oFIR - note that this is technically UB).
    - Alignment does still matter tho: https://gcc.godbolt.org/z/Pn1eqo

avadacatavra:
I'm going to merge this, and any additions can be filed as followup issues/PRs (cc @nikomatsakis)

PR #39, opened 2018-10-24
basic chapter for representation of unions

### LINKED ISSUE #13: Representation of unions
Discussing how unions are laid out.

- Is `#[repr(C)]` meaningful when applie to a union?
- When (if ever) do we guarantee that all fields start at offset 0?
- When (if ever) do we guarantee that all fields have the same address?
- Any key things to note re: FFI interop?

RalfJung:
Fixes #13


reference/src/representation/unions.md~
```diff
@@ -0,0 +1,2 @@
+# Representation of unions
```
joshtriplett: This backup file shouldn't exist.

reference/src/representation/unions.md
```diff
@@ -0,0 +1,36 @@
...
+
+```rust
+union U { f1: T1, f2: T2 }
+```
+
+is to determine the offset of the fields.  The layout of these fields themselves
+is already entirely determined by their types, and since we intend to allow
+creating references to fields (`&u.f1`), unions do not have any wiggle-room
+there.
+
```
joshtriplett: Please include a note that unions may contain values that do not correspond to a valid value of any of their fields.

RalfJung: That's not a layout question ("where do which fields go, what about alignment and size?"), but a validity question ("which bit patterns are allowed?"). It seems in the wrong place here.

joshtriplett: It seems like a "representation" question. It doesn't have to go in *this* paragraph, but it seems like it should go somewhere in this document.

I'm open to suggestions if it should appear elsewhere instead, though.

nikomatsakis:
@RalfJung maybe it's worth noting that `#[repr(C)]` is probably also needed for ABI compatibility when invoking functions (although at present I expect we are de facto compatible).

reference/src/representation/unions.md
nikomatsakis: I agree with @RalfJung that seems like an *invariant* question more than layout. I view layout as answering "when you go to write/read this field, at what offset will you find it?" whereas this question is sort of "what might you find there". 

In particular, I think layout+representation are purely static properties.

avadacatavra: This can be resolved at a later date during editing--for the purposes of this PR, validity is out of scope.

PR #40, opened 2018-10-25
lvalues are called places in Rust/MIR

RalfJung:
At least for anyone with a background in functional programming, calling these a "\*value" is very confusing. *Values* are the result of a computation, e.g. "5" is a value. "2+3" is not a value, it is an *expression* that evaluates/computes to the value "5".

The lvalue/rvalue distinction is on a different axis though, I think. Composing the two terminologies, we have "place expressions" like `x[3].foo`, and "place values" like (memory address) `0x00803240` -- where again "place values" are the thing that "place expressions" evaluate to. This all makes a lot of sense. If we'd stick to C(++) terminology, we would have to distinguish "lvalue expressions" and "lvalue values", and that seems just absurd.

This distinction, btw, manifests in the codebase as well: `rustc::mir::Place` is a place expression, `rustc_mir::interpret::Place` is a place value.

---

I think we should find matching terminology to replace "rvalue". Some "X" referring to the result of a computation like `2 + 3` or `*foo.bar` or `&(*x).field`. The result can be pretty much any sequence of bytes, the key factor being that you can read them (copy them into a place), but you cannot change them.

Right now, their expression form is `rustc::mir::Rvalue`, and their value form is `rustc_mir::interpret::Operand`. We could call these things "operand expression/value", if `rustc::mir::Operand` wasn't already taken to be something else ("operand expressions" are a subset of "rvalue expressions", namely they are constants or a place expression, they don't actually compute anything on the data). Any suggestions for a good noun to use here?

### DIFF
--- a/reference/src/introduction.md
+++ b/reference/src/introduction.md
@@ -8,7 +8,7 @@
 * *niche*
 * *layout*
 * *tag*
-* *rvalue*
-* *lvalue*
+* *place* (or *lvalue* in C/C++ speak)
+* *rvalue* (maybe we can come up with a Rust term for this as well?)
 
 ## Unsafe abstraction


ubsan:
I still like "value" and "place"; I do think that the "place value" thing doesn't make a lot of sense. They're called "addresses" or "reference values", or "values of reference type".

reference/src/introduction.md
```diff
@@ -8,7 +8,7 @@
 * *niche*
 * *layout*
 * *tag*
-* *rvalue*
-* *lvalue*
+* *place* (or *lvalue* in C/C++ speak)
```
ubsan: These are called "lvalues" in C, and "glvalues" in C++.

reference/src/introduction.md
```diff
@@ -8,7 +8,7 @@
...
+* *rvalue* (maybe we can come up with a Rust term for this as well?)
```
ubsan: These are called "values of the expression" in C, "prvalue" in C++.

ubsan: (and we should just call them "values")

RalfJung:
> "addresses"

A `Place` is not the same as a `Pointer` though (speaking in miri types).

> "reference values", or "values of reference type"

These certainly do not work, there is a distinction between a `Place` and an "`Operand` of reference type" (again using the miri types, for lack of a better term).  We can convert between the two using `&` and `*`, but I think it is crucial that we have clear terminology to distinguish the two -- exactly *because* they are so similar, but distinct. It took me years to understand this stuff properly, and I think one reason is that the terminology is so messed up and confusing.

Both places and what you call "values" have the distinction between "what the programmer writes down" and "what this eventually denotes/evaluates to", so we should use consistent terminology. "Compile-time place" vs "run-time place" would also work, but I think using the well-established expression vs value terminology from functional languages makes most sense here. And either way we should not use "value" in another way, that will just be confusing to the part of our community coming from functional languages where that term has a very clear meaning already. It might be better to avoid "value" entirely.

reference/src/introduction.md
RalfJung: Isn't there a term to distinguish expressions like "3+2" from expressions like "x", where the latter can be used as lvalues but the former cannot?

ubsan: Yes, the latter is a (Rust) place expression, (C) lvalue expression, or (C++) glvalue expression.

ubsan:
@RalfJung oh, I missed what you meant by "place value", just ignore that part.

RalfJung:
The best term I could come up with so far for `rustc::mir::Rvalue` and `rustc_mir::interpret::Operand` is to call them an "object" -- so the pattern would be that "an object is what you put into a place".  But that clashes with trait objects and OOP.

ubsan:
@RalfJung why not "Operand"?

RalfJung:
FWIW, I am going to call them "robjects" for now just so that I can consistently refer to them some way; that term should be sufficiently bad that nobody should like it enough to grow attached to it. ;)

["Operand"](https://doc.rust-lang.org/nightly/nightly-rustc/rustc/mir/enum.Operand.html) is already used in MIR to refer to the LHS and RHS of `+`. In MIR, robjects do not nest (to keep the IR simpler). So lots of confusion would sure ensue in the compiler if we called robjects "operands".
That said, I do like the term. We'd have to rename the current use of "Operand" though. "PrimOperand" or "SimpleOperand" or so come to my mind.

ubsan:
So, I'm going to recommend we call them immediates. It's a reasonable way of referring to them, and evokes their ephemeral nature.

rkruppe:
For the record "immediate" it already used for other compiler-y things

- small constants stored directly in instructions (in most architectures)
- concepts derived from the above, such as inline asm constraints
- as of very recently, small primitive values in rustc CTFE and in miri (and I'm sure it's not the first one to use the word in that way, but I can't immediately name others)

 We're not going to find any fitting term that's completely unused, but personally I have to deal with the CPU architecture kind of immediate multiple times per day on average so I wouldn't be super happy if it became overloaded like that.

ubsan:
@rkruppe There's not a lot of words that can reasonably be used without overloading somebody.

rkruppe:
Sure, I said as much myself. I just want to spell out for each option who that is going to be. For "immediate" it's going to be everyone who interacts with assembly and (unless that name gets changed again) everyone who works on miri or on certain parts of rustc.

ubsan:
I still personally think overloading `value` to mean `result of an expression which is not a place`, and using `result` to mean `result of the evaluation of an expression`, is my favorite way to go, especially as it's the least confusing to those who aren't experts in the field, imo.

RalfJung:
@ubsan Even then we still lack a term for "non-place expressions".  Expressing that as a negative (non-place) is far from optimal.

I consider "Operand" to be the best option so far.

ubsan:
@RalfJung this is what we're discussing.

The options so far for "result of expression:
  - value
  - result

The options so far for "non-place result of expression":
  - value
  - operand
  - immediate
  - rvalue

if you want to discuss an expression, then you can add the category of the result to "expression", i.e.,
  - value expression
  - operand expression
  - immediate expression
  - rvalue expression

this is contrasting to "place" and "place expression", respectively.

RalfJung:
> "non-place result of expression"

That's not what I intended "operand", "robject" or so to mean. I was originally asking for (and I understand now I wasn't clear enough, so here you go):

* a term for "non-place" (no matter whether in expression form or as the result of an evaluation), `$NON_PLACE`, and
* a term for the result of evaluating an expression, `$EVAL_RESULT`.

The constraint I hope to satisfy is that we can say "place expression" and "`$NON_PLACE` expression" if we have to emphasize that this entity has *not* been evaluated to a result yet, and that we could use "place `$EVAL_RESULT`" and "`$NON_PLACE` `$EVAL_RESULT`" if we have to emphasize that this entity *has* been evaluated to a result. So we get a nice two-by-two grid for all four possible combinations of place/non-place and expression/evaluated-to-result, and we have that grid filled uniformly.

For example:

|    |      Place      |  Not a place |
|----------|-------------|------|
| **Expression** |  place expression | operand expression |
| **Completely evaluated** |    place value   |   operand value |

This is my current favorite (after scratching "object").

Or: place expression, rvalue expression, place value, rvalue value. I find this really awkward, "rvalue value" is just not a good term.

Or: place expression, operand expression, place result, operand result. I guess I could live with this.

Or: place expression, rvalue expression, place result, rvalue result. I would dislike this because it goes against the common (but admittedly not omnipresent) notion that a "value" is some kind of normal form, so "rvalue expression" is both saying it is already in normal form and yet it needs evaluating (being an expression)...?

---

As for "immediate", I agree with @rkruppe. I think "operand" is likely going to cause the least confusion. It is certainly my favorite from @ubsan's list. If we accept that I would propose we rename:

* `rustc::mir::Operand` -> `rustc::mir::AtomicOperand` (bikeshed still open)
* `rustc::mir::Rvalue` -> `rustc::mir::Operand`

No renaming would be necessary in miri as it already uses the "operand" terminology.

ubsan:
I think, personally, that it will cause the least confusion if we use

- place expression/place result
- value expression/value result

operand makes me think too much of "argument to a function/operator".

RalfJung:
I am somewhat surprised that you want to make that distinction. It does not arise in surface Rust, does it? Everything you can use on the RHS of an assignment, you can also use as a function argument.

The distinction does arise in MIR, and it is precise the difference between what I proposed to be called "atomic operand" ("primitive operand", "simple operand") and a general operand.

---

Also, I thought we had agreed to not use terminology that will confuse the heck out of a sizeable fraction of people caring about Rust. :/  In particular, the whole point of `$NON_PLACE` was to be able to say things like "`$NON_PLACE`s come in two forms: `$NON_PLACE` expressions and `$NON_PLACE` `$EVAL_RESULT`s". Saying that a "value expression" is a form of a "value" is just nonsense, as far as I am concerned.

I begrudgingly accepted to drop "object" because you said it would be very confusing to people that are deep into C/C++ terminology. I am still wondering how this clashes with C's use of the term. "The content of a place is an object" seems perfectly consistent, and there also is a notion of "temporary object" that seems to fit well with what miri calls "immediate operands" (would then be "immediate objects"), namely constants like in `x + 1`.
If that is sufficient grounds for dropping "object", we ought to drop "value" for the same reason.

---

Another term that just came to my mind is "intermediate" (the noun).   The idea being that `$NON_PLACE`s actually are not very long-lived, they are computed only to be immediately stored into a place. Unfortunately in English this is also an adjective, because otherwise I think it'd be rather well-suited. :/ But this way, "intermediate expression/result/value" do not work very well.

And to mention yet another alternative, we could call `$NON_PLACE` expressions "r-expressions" and `$NON_PLACE` values/results "r-values/results". One problem with this is that there is no collective term that encompasses both of them.

gnzlbg:
> I begrudgingly accepted to drop "object" because you said it would be very confusing to people that are deep into C/C++ terminology. I am still wondering how this clashes with C's use of the term. "The content of a place is an object" seems perfectly consistent,

For what is worth, having some background in C++ terminology, I do not think using object in this way is an issue. I am not as deep in as @ubsan and I agree with them that `object` is an overloaded term. But whichever terms we choose are going to have to be very precisely defined, and I don't think it will take long till we get used to them.

> For example
> [..table..]
> This is my current favorite (after scratching "object").

Could you show the same table using object ?

Independently of the terms we choose (object, result, operand, ...), I think this type of table representation is much easier to understand than than the trees used to represent "value" categories in C++.

RalfJung:
> Could you show the same table using object ?

Sure, it is literally `s/operand/object/g`:

|    |      Place      |  Not a place |
|----------|-------------|------|
| **Expression** |  place expression | object expression |
| **Completely evaluated** |    place value   |   object value |

Or with "result":

|    |      Place      |  Not a place |
|----------|-------------|------|
| **Expression** |  place expression | object expression |
| **Completely evaluated** |    place result   |   object result |

In both columns, I assume we'd sometimes just talk of "places" or "objects/operands/..." when it is clear from context whether we are talking about an expression or a value/result.

> Independently of the terms we choose (object, result, operand, ...)

Note that "result" was only ever proposed for `$EVAL_RESULT`, not for `$NON_PLACE`.

gnzlbg:
So @ubsan , is this what you meant?  

| | Expression | Completely Evaluated |
|---|----|---|
**Place** | place expression | place result
**Not a Place** | value expression | value result

RalfJung:
@gnzlbg You flipped the table at the diagonal.^^ Makes it rather confusing to compare them...

gnzlbg:
Sorry, I just copied @ubsan lines (that confused me a bit) :D 

Here it is unflipped:

| | Place | Not a Place |
|---|----|---|
**Expression** | place expression | value expression
**Completely Evaluated** | place result | value result

ubsan:
@gnzlbg `object` is an important term of art, one which I will continue to use within Rust.

gnzlbg:
So could we nail this down (and summarize this) to a couple of concrete proposals that can be discussed in the next meeting ?

Ixrec:
Before I over-bikeshed the rvalue question, I'd like to clarify that I am a huge fan of using "place" for lvalues in Rust.

> In particular, the whole point of $NON_PLACE was to be able to say things like "$NON_PLACEs come in two forms: $NON_PLACE expressions and $NON_PLACE $EVAL_RESULTs". Saying that a "value expression" is a form of a "value" is just nonsense, as far as I am concerned.

Maybe I'm missing something, but it also seems like nonsense to say that a "place expression" is a form of a "place"? Surely "place/value/object/etc expression" is _an expression that evaluates to_ a place/value/object/whatever, not _a form of_ a place/value/object/etc, right? Though I'm not a huge fan of "value" for this anyway since it's not at all intuitive that "places" don't qualify as "values" (if nothing else, C++'s "lvalue" set a pretty strong precedent that places are a kind of value). In general, I think "value" is just far too vague to be helpful in precise definitions, even more so than all of the other options.

-----

I don't think using "object" for this purpose would be confusing just because C++ objects are a totally different thing; everyone knows Rust doesn't have traditional OOP classes. But there is another reason why I'm not a fan of "object" in particular: C++ gives "object" two completely different meanings, effectively one for normal programmers and one for language lawyers, which makes any use of it in a formal/spec-y/UB-defining context come across as an obfuscating term that immediately makes the document gibberish to anyone that isn't already familiar with what definition is implied and all the unstated implications of that definition (I still have no idea which primitives types are and are not "lawyer objects" in C++). In other words, if "object" was used, it'd give me the impression that, just like the C and C++ specs, being comprehensible for non-wizards/non-lawyers is a non-goal of these documents. Which is probably completely unfair to you guys, but that is the connotation C++ has given it for me.

@ubsan I've never heard of "object" having a defined meaning in Rust, unless it's supposed to be short for "trait object". What meaning are you talking about?

-----

Using "operand" for this would also be super weird because it already has the "normal people meaning" of things you pass to operators in Rust and pretty much every other language. So giving that term a double meaning no one would ever expect also seems like it'd create the "wizards only" connotation.

In particular, the popular oversimplification that that a C++ lvalue is "anything you can put on the left of the assignment operator" immediately implies that _places are operands_ for assignment operators, so I think that'd be just as bizarre to C++ users as "object".

-----

On the constructive side, I don't think we've exhausted all the candidates yet. I can think of at least two:
- temporary
- data / datum

They aren't amazing options, but I do think "data expression" causes nowhere near as many problems as "object/value/operand expression" would.

nikomatsakis:
Bah. I wanted to write a big opus here but I've not had time to re-read the comments and really form an opinion. One thing I did want to say though:

I've generally found that "place expressions" is a bit more awkward in practice than I would like. I've found myself gravitating more and more to the term *path* to represent an "lvalue" -- basically, an expression that leads one to some *place* in memory (i.e., `place` as the runtime value, and `path` as the static equivalent). I don't know that this fits the matrix @RalfJung was promoting, but it's just  a thought I've had recently. 

(I believe there is some precedent for using the term *path* to refer to expressions like `a.b.c`, in any case, outside of Rust, though I can't be bothered to go looking for it.)

RalfJung:
> I've found myself gravitating more and more to the term path to represent an "lvalue" -- basically, an expression that leads one to some place in memory (i.e., place as the runtime value, and path as the static equivalent).

So you think `rustc::mir::Place` should be renamed to `rustc::mir::Path`?

Main main issue with that is that the fact that "place" and "path" are the dynamic and static representation of each other is not at all obvious in the terms.

> I believe there is some precedent for using the term path to refer to expressions like a.b.c, in any case, outside of Rust, though I can't be bothered to go looking for it.

AFAIK this does not usually include derefs though. Actually, my understanding was that `.b.c` is the path, and that it is then applied to `a`, i.e., a path is just syntax for an integer used as pointer offset -- but that may be wrong.

RalfJung:
As part of our ongoing efforts to formalize a language that's reasonably close to MIR (and maybe extends a bit to slightly higher levels of abstractions, we'll see), I came to realize one thing: If I now had to define "rvalue", I'd say it is "something that can be put into a place".  Putting it into a place is the only "elimination-style" operation that is ever performed on rvalues.  The candidate model we are considering here is to literally model an rvalue as a function (closure) that takes a place, and then has the effect of storing itself into that place.  Miri's "operands" (which are either a place or an immediate constant) cover two special cases of "things you can copy into a place".

I wonder if there is a good name one could derive from this.

gnzlbg:
I think I like:

|    |      Place      |  Not a place |
|----------|-------------|------|
| **Expression** |  place expression | value expression |
| **Completely evaluated** |   (evaluated) place  |   (evaluated) value |

where we just use "expression" to denote places and values that are not completely evaluated, and just "place" and "value" to mean evaluated ones. If the context requires disambiguation or being extremely pedantic, one can use "evaluated" to emphasize that, resulting in "evaluated place" and "evaluated value".

Centril:
Like @RalfJung (https://github.com/rust-lang/unsafe-code-guidelines/pull/40#issuecomment-437838293), I think it is natural to say that *"An __object__ can be put into a __place__."* which fits with https://github.com/rust-lang/unsafe-code-guidelines/pull/40#issuecomment-453029066. However, as @Ixrec noted it might not work well so *datum* seems fine to me.

|                      | **Place**  | **Not a place** |
| ------------- | ------------- | ------------- |
| **Expression**  | place expression  | datum expression |
| **Completely evaluated** | place result? | datum result? |

When you don't care about the expression / result distinction you can just drop that part and just talk about "place" and "datum".

Also, we can say:

+ "if the result of an expression is a datum, it's called a datum expression"
+ "if the result of an expression is a place, it's called a place expression"
+ "if the result of an expression is a type, it's called a type expression"

ubsan:
I don't think there's a reasonable argument against `value`.

I have taught the intuition of this stuff to many people across the experience spectrum, and `value` is a useful word in this area. `datum` is not even close to that. I would love to be able to use the same term between explanations and the standard - otherwise (as I do in C++), I have to explain "yeah so a value is called a `prvalue` in the standard".

rkruppe:
The argument against using "value" for this purpose seems to be that it could instead be used as specific term for the result that place/value expressions evaluate to, to explicitly distinguish them from the expressions. I for one can understand why some people might want that, but there's multiple reasonable alternatives to that, and the usefulness for teaching is significant -- @ubsan is right, *value* is by and large the predominant and known term for the things we put into places.

As far as I can tell, the conflict here is not that anyone disputes that `5` or `(2, Some("foo"))` are "values" in any sense of the word but just that some people would prefer to use value in a different way (for evaluation results specifically, regardless of place or thing-put-into-place). However, after exploring all the other terms above, it's becoming clear to me that there is no good alternative to value for "things that are put into places", so even though it's good we had the discussion about those alternatives, I think it's a failed project.

So I'd like to raise the banner again for simply having place expressions and value expressions result in places and values respectively. That is, demarcating the evaluation result from expressions by simply not including "expression", and not caring that informal discussion will sometimes use the shorer terms to include the expression forms as well. This is [what the reference does right now](https://doc.rust-lang.org/reference/expressions.html#place-expressions-and-value-expressions) and personally, despite having some formal/PL background and often nit picking about the difference between expressions and evaluation results, I've never had the slightest problem with it.

---

One other position I find interesting is @nikomatsakis saying "place expressions" is a bit awkward in practice. I can see that, though the alternative "path" massively collides with the existing use in the module/namespace system (including e.g. corresponding macro_rules fragments -- I could totally see declarative macros wanting to accept place expressions specifically).

RalfJung:
@ubsan

> I don't think there's a reasonable argument against value.

Are you referring to the entire right column in [my table](https://github.com/rust-lang/unsafe-code-guidelines/pull/40#issuecomment-437841297) or just the bottom-right corner?  If you include the top-right corner, then... let me just say that I'd be fine with you putting different weight on things, but outright calling my arguments unreasonable is not a productive way to communicate.

@rkruppe 

> value is by and large the predominant and known term for the things we put into places.

That would be the bottom-right corner, though? We don't put "2+2" into a place, but "4". I even found myself tempted to call the things in the bottom-right corner "values". I could live with that, provided we find a reasonable term for the top-right corner.

"value expression" is... odd.

> This is what the reference does right now

"Place Expressions and Value Expressions", oh wow. I had no idea.

---

FWIW, in an ongoing formalization where "value" would heavily conflict with existing terminology, we call these things "temporaries" since they only exist ephemerally during an assignment (in our model). Once stored in a place, they become a sequence of "things that can be stored in a heap cell", which we call "immediates" (this matches Miri's use of the term "immediate"; our heap cells are just a bit weird -- in a more realistic model this would be just "byte").

In a more C-style model with an idea of "abstract values", the term "temporary" admittedly works less well.

rkruppe:
> That would be the bottom-right corner, though? We don't put "2+2" into a place, but "4". I even found myself tempted to call the things in the bottom-right corner "values". I could live with that, provided we find a reasonable term for the top-right corner.
> 
> "value expression" is... odd.

It's not perfect IMO but "a foo expression evaluates to a foo" is a nice mnemonic and I don't see a better overall alternative.

re: temporaries, it would be a fine term on its own, but besides the downside you mention it also collides with the idea of compiler-generated temporary *places* that occur in expressions such as `foo(&bar())`. And calling those things just "temporaries" has a big precedent across languages, so I don't think we should try to change that terminology.

gnzlbg:
This needs rebasing.

gnzlbg:
I'd like to propose revisiting this in the next meeting (next thursday) with the goal of making progress here. Ideally, we would agree on the "least bad" option, temporarily pick it, and open an issue to keep discussing this further. Picking the "least bad" option and start using it would give us experience with its downsides, which might allow us to come up with better terminology in the future.

RalfJung:
I won't be able to join next Thursday, can we do this in two weeks?

gnzlbg:
> I won't be able to join next Thursday, can we do this in two weeks?

Sure!

RalfJung:
Btw, turns out the reference used to speak about lvalues and rvalues but was [changed in December 2017](https://github.com/rust-lang-nursery/reference/pull/167) to talk about "place expressions" and "value expressions" instead, without much of a discussion. Still almost one year earlier than this discussion started.

gnzlbg:
It's unclear whether we will tackle this on the next meeting. @RalfJung do you think you would be able to write a summary comment here with the alternatives that got the most traction, and maybe include at the end a non-binding poll to see where everyone stands ?

RalfJung:
Closing in favor of https://github.com/rust-lang/unsafe-code-guidelines/pull/175, where I concede to the majority opinion that the 4 terms in question are "value", "place", "value expression" and "place expression".

RalfJung:
@gnzlbg sorry, I was that too late. But with @rkruppe, @nikomatsakis, @gnzlbg  and @ubsan in agreement, I think this is pretty much settled. 

We adapted this terminology for our Coq formalization of Stacked Borrows. We also introduced the term "result" to be able to talk about "evaluated place or evaluated value", i.e., "something that does not compute further". It's been a bit confusing at first, and there is friction because all our other developments use different terminology, but it'll work.

PR #45, opened 2018-11-05
initial pass at function pointers

ubsan:


reference/src/representation/function-pointers.md
```diff
@@ -0,0 +1,114 @@
+# Representation of Function Pointers
+
+### Terminology
+
+In Rust, a function pointer type, is either `fn(Args...) -> Ret`, or
+`extern "ABI" fn(Args...) -> Ret`. A function pointer is the address of a
+function, and have function pointer type.
+The pointer is implicit in the `fn` type,
+and `fn` types are implicitly `'static`.
```
sfackler: There's also `unsafe fn()` and `unsafe exterN "ABI" fn()`.

ubsan: ah, true

reference/src/representation/function-pointers.md
```diff
@@ -0,0 +1,116 @@
...
+and the opposite is true of `dlsym`, which requires you to convert from
+`void*` to `void (*)()` in order to get the address of functions.
+This conversion is _not_ guaranteed by Rust itself, however;
+simply the implementation. If the underlying platform allows this conversion,
+so will Rust.
+
+However, null values are not supported by the Rust function pointer types --
+just like references, the expectation is that you use `Option` to create
+nullable pointers. `Option<fn(Args...) -> Ret>` will have the exact same ABI
+as `fn(Args...) -> Ret`, but additionally allows null pointer values.
```
RalfJung: This should likely reference the `enum` chapter once one exists; I guess the rules here will be exactly the same as the ones for reference types?

ubsan: Yep!

reference/src/representation/function-pointers.md
```diff
@@ -0,0 +1,116 @@
...
+
+### Terminology
+
+In Rust, a function pointer type, is either `fn(Args...) -> Ret`,
+`extern "ABI" fn(Args...) -> Ret`, `unsafe fn(Args...) -> Ret`, and
+`unsafe extern "ABI" fn(Args...) -> Ret`.
+A function pointer is the address of a function,
+and has function pointer type.
+The pointer is implicit in the `fn` type,
+and `fn` types are implicitly `'static`.
```
nikomatsakis: I'm not sure what it means to be `'static`. If "`T` is 'static" means `T: 'static`, then this is not *strictly* true. For example,

```rust
fn(&'a u32): 'static
```

does not hold. This is fallout from the "syntactic definition of outlives". Such a type is not entirely trivial to make though.

I think the point is that fn pointer types do not carry their own lifetime, so they are always assumed to point to a statically defined function with no state of its own, right? Maybe we can word that more directly.

RalfJung:
On Zulip, @ubsan was asking about function pointer transmutes:

> Can you `transmute<fn(*const T), fn(*const ())>(x)(p)`?

(I assume we have `T: Sized` here.)

I would say we can modularize the problem (but this goes deep into validity invariant territory):

* The actual cast is always okay, because the validity invariant for function pointers does not talk about the signature. Maybe it just says "non-NULL", maybe it also says "points to executable memory", but I think it should not take the signature into account.
* The call, then, is when things may become UB if the signatures are wrong. We probably need some definition of "ABI compatible" here: Which "mismatches" we allow between the caller and callee signature will depend not just on layout but also on ABI concerns. For example, CTFE/miri currently say that either the type must be exactly the same, or else both types must have [`Scalar`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc/ty/layout/struct.Scalar.html) ABI that may differ only in its `valid_range`. In your example, the second condition is satisfied.
* The behavior of allowed argument mismatches, then, is the same as that of a `transmute`, of union-based type punning, and of type punning through changing pointer types: In your case, I would say your code is equivalent to `x(mem::transmute<*const T, *const ()>(p))`.

rkruppe:
> The behavior of allowed argument mismatches, then, is the same as that of a transmute, of union-based type punning, and of type punning through changing pointer types: In your case, I would say your code is equivalent to x(mem::transmute<*const T, *const ()>(p)).

That doesn't work because type punning is just about in-memory representation (and notions that build on it, like validity) but here the call ABI matters as well. For example, one can easily transmute between integers and floats, but trying `transmute<fn(f32) -> f32, fn(i32) -> i32>(f32::sin)(42)` needs to be UB: on many targets the caller would pass the parameter in an integer register but the callee expects it in a floating point register.

RalfJung:
> here the call ABI matters as well.

That's why there is an ABI compat check first, declaring full UB unless the types are ABI-compatible "enough".

rkruppe:
Oops, don't know how I missed that 🤐

RalfJung:
@ubsan will you have time to do another pass over this, incorporating comments, until our next meeting in two weeks?

ubsan:
@RalfJung yeah! I'll do it this weekend.

nikomatsakis:
Things remaining to be done before merging:

- Address the question of `'static` [raised here](https://github.com/rust-rfcs/unsafe-code-guidelines/pull/45#discussion_r233911726)
- Open an issue on the topic of "when can fn pointers be transmured" and [copy over and/or link to Ralf's comment here](https://github.com/rust-rfcs/unsafe-code-guidelines/pull/45#issuecomment-439604001)
    - we can reference the issue in the text

reference/src/representation/function-pointers.md
```diff
@@ -8,7 +8,16 @@ In Rust, a function pointer type, is either `fn(Args...) -> Ret`,
...
-and `fn` types are implicitly `'static`.
+and they have no lifetime of their own;
+therefore, function pointers are assumed to point to
+a block of code with static lifetime.
+This is not necessarily always true,
+since, for example, you can unload a dynamic library.
+Therefore, this is _only_ a safety invariant,
+not a validity invariant;
+as long as one doesn't call a function pointer which points to freed memory,
+it is not undefined behavior.
```
RalfJung: I'd just say something like: "Function pointers have no lifetime. They must point to an allocated block of code when being called."

The rest of this belongs to https://github.com/rust-rfcs/unsafe-code-guidelines/issues/72, doesn't it?

RalfJung:
> Open an issue on the topic of "when can fn pointers be transmured"

Done: https://github.com/rust-rfcs/unsafe-code-guidelines/issues/81

reference/src/representation/function-pointers.md
```diff
@@ -0,0 +1,125 @@
...
+    - thunk may be null, and shall be passed unchanged to func
+    - self may be null, in which case no iteration is done
+*/
+
+void iterate(struct Cons const *self, void (*func)(int, void *), void *thunk);
+bool for_all(struct Cons const *self, bool (*func)(int, void *), void *thunk);
+```
+
+```rust
+pub struct Cons {
```
vi, line 84: Is the lack of `#[repr(C)]` accidental or intentional?

Lokathor: probably accidental, however you should probably open a new issue about this, because a years old PR is a little too out off the way to get much attention.

vi: Shall I stop using this algorithm (spot suspicious place in a git-based book -> go to repo -> go to that file -> blame -> comment) in general?

Github UI seems to provide nice way of attaching comments to specific lines in files, so I expected those comments would be delivered precisely to those who deal with that particular file.

Lokathor: Well as I said this PR is years old. The person who wrote it doesn't even really work on Rust any more.

We are happy to have people taking an interest in fixing anything amiss, it's just that a new issue is generally preferred over continued discussion on a merged or closed PR, particularly an older one. That way the question and/or comment will stay on the list of things to resolve rather than be potentially forgotten by accident. Many people working on Rust have so many watched issues and projects that individual notifications/emails can easily get lost among them all.

PR #47, opened 2018-11-18
Layout of vector types

### LINKED ISSUE #38: Layout of vector types
The layout of Rust vector types (e.g. `core::arch::x86_64::__m128` and friends) is currently unspecified but we probably want to make it implementation defined. That requires the implementation to specify their layout in "unsafe-code-guidelines speak" so we might just as well do it here for the canonical implementation at least for the stable types on the tier 1 platforms (maybe there is a way to do this generally). Some things to consider: 

* In C, `__m256` is only available when `AVX` is available such that its layout is always a 256-bit wide register. 

*  In Rust, `__m256` is always available, even when 256-bit or 128-bit wide registers are not available. That is, `repr(C)` and `repr(Rust)` vector types do not necessarily have the same layout. 

* The Rust function abi for vector types transparently handles this by passing vector types through memory, but the x64 SysV ABI passes them in registers (this leads to subtleties when using them in C FFI: https://github.com/rust-lang/rust/issues/53346).

gnzlbg:
Closes #38 .

### DIFF
--- a/reference/src/representation/vectors.md
+++ b/reference/src/representation/vectors.md
@@ -14,7 +14,8 @@ currently different for each architecture.
 
 ## Vector types
 
-Vector types are `repr(simd)` homogeneous tuple-structs containing `N` elements of type `T`:
+Vector types are `repr(simd)` homogeneous tuple-structs containing `N` elements
+of type `T`:
 
 ```rust
 #[repr(simd)]
@@ -84,10 +85,8 @@ unsafe {
   
   [#36]: https://github.com/rust-rfcs/unsafe-code-guidelines/issues/36
   
-* Should unions containing a vector type and any number of ZSTs also be vector
-  types? For example, should `MaybeUninit<Vector<T, N>>` also be a vector type?
-  This should be documented for `unions` in general, and the rules for vectors
-  should follow from that. This document should, at most, mention that in the
-  context of vectors.
-
 * Should we require `N` to be a power-of-two?
+
+* `MaybeUninit<T>` does not have the same `repr` as `T`, so
+  `MaybeUninit<Vector<T, N>>` are not `repr(simd)`, which has performance
+  consequences and means that `MaybeUninit<Vector<T, N>>` is not C-FFI safe.
--- a/reference/src/representation/vectors.md
+++ b/reference/src/representation/vectors.md
@@ -84,8 +84,10 @@ unsafe {
   
   [#36]: https://github.com/rust-rfcs/unsafe-code-guidelines/issues/36
   
-  
 * Should unions containing a vector type and any number of ZSTs also be vector
   types? For example, should `MaybeUninit<Vector<T, N>>` also be a vector type?
+  This should be documented for `unions` in general, and the rules for vectors
+  should follow from that. This document should, at most, mention that in the
+  context of vectors.
 
-* Should we guarantee that `N` is a power-of-two ?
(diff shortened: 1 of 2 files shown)


reference/src/representation/vectors.md
```diff
@@ -0,0 +1,91 @@
...
+  ```
+  
+  This is blocked on the resolution of issue [#36] about the layout of
+  homogeneous structs and tuples.
+  
+  [#36]: https://github.com/rust-rfcs/unsafe-code-guidelines/issues/36
+  
+  
+* Should unions containing a vector type and any number of ZSTs also be vector
+  types? For example, should `MaybeUninit<Vector<T, N>>` also be a vector type?
```
RalfJung: Which consequences does it have to answer this with "no" or "yes"? I don't understand what "being a vector type" entails.

I would find it strange if this document said anything about unions; that should all fall out of the general union rules. In terms of layout, `repr(C)` unions have all their fields starting at offset 0 and hence if all but one variant are size-0-align-1, their layout (at least size and alignment) is identical to that of the only non-ZST field.

reference/src/representation/vectors.md
```diff
@@ -0,0 +1,91 @@
...
+
+* Should we guarantee that `N` is a power-of-two ?
```
RalfJung: If anything, shouldn't we rather *require* it, at least once people can define their own vector types?

reference/src/representation/vectors.md
gnzlbg: > I don't understand what "being a vector type" entails.

It affects the calling convention, e.g., when interfacing with C (vectors are not aggregates). So depending on what we decide here, one might be able to use `MaybeUninit<Vector>` when interfacing with C or not. Calling conventions are neither representation nor validity AFAICT, so I don't know when and where to discuss this, but we should not forget about it. 

> I would find it strange if this document said anything about unions; that should all fall out of the general union rules.

Yes, this should be resolved more generally in the context of unions. `MaybeUninit<T>` is a general construct, and we might want it to be an aggregate only if `T` is an aggregate, but not if e.g. `T` is `c_int`. 

This document should, at most, repeat what's mentioned about this somewhere else.

reference/src/representation/vectors.md
```diff
@@ -0,0 +1,93 @@
...
+```rust
+assert_eq!(size_of::<Vector<T, N>>(), size_of::<T>() * N);
+assert!(align_of::<Vector<T, N>>() >= align_of::<T>());
+```
+
+That is, two distinct `repr(simd)` vector types that have the same `T` and the
+same `N` have the same size and alignment.
+
+The layout of elements within a vector is the same as that of `[T; N]` such that
+there is a a 1:1 correspondence between the indices of the vector and arrays:
```
rkruppe: Nit: ABI alignment is often considered to be a part of "layout", but alignment can be different here (which affects type punning through pointers, though unions are fine). So for clarify I'd prefer to mention that problem explicitly and word the element order issue a bit differently, e.g., "elements are laid out in source field order, enabling random access to vector elements by reinterpreting the vector as an array: [example]".

gnzlbg: I've pushed a commit with your wording.

reference/src/representation/vectors.md
RalfJung: Ah so this is about ABI then. Yeah I have no idea.^^

PR #49, opened 2018-11-27
added floating point/int summary

avadacatavra:
It didn't seem like there was a lot to say about representation without getting into validity

### DIFF
--- a/reference/src/representation/integers-floatingpoint.md
+++ b/reference/src/representation/integers-floatingpoint.md
@@ -1,0 +1,47 @@
+# Representation of Boolean, Floating Point, and Integral Types
+This chapter represents the consensus from issue [#9]. It documents the memory layout and considerations for `bool`, `usize`, `isize`, floating point types, and integral types.
+[#9]: https://github.com/rust-rfcs/unsafe-code-guidelines/issues/9
+
+## Overview
+These are all scalar types, representing a single value. These types have no representation variants (no #[repr(C)] or #[repr(Rust)]. Their size is fixed and well-defined across FFI boundaries and map to their corresponding integral types in the C ABI.
+- `bool`: 1 byte
+  - any `bool` can be cast into an integer, taking on the values 1 (true) or 0 (false)
+- `usize`, `isize`: pointer-sized unsigned/signed integer type
+- `u8` .. `u128`, `i8` .. `i128`
+  - {8, 16, 32, 64, 128}-bit unsigned integer
+  - {8, 16, 32, 64, 128}-bit signed integer
+- `f32`, `f64`
+  - IEEE floats
+  - 32-bit or 64-bit
+- `char`
+  - C++ char: equivalent to either `i8`/`u8`
+  - Rust char: 32-bit
+  - not ABI compatible
+  - represents [Unicode scalar value](http://www.unicode.org/glossary/#unicode_scalar_value)
+
+##`usize`/`isize`
+Types `usize` and `isize` are committed to having the same size as a native pointer on the platform. The representation of `usize` determines the following:
+- how much a pointer of a certain type can be offseted,
+- the maximum size of Rust objects (because size_of/size_of_val return `usize`),
+- the maximum number of elements in an array ([T; N: usize]),
+- `usize`/`isize` in C FFI are compatible with C's `uintptr_t` / `intptr_t`.
+
+The maximum size of any single value must fit within `isize` to [ensure that pointer diff is representable](https://github.com/rust-rfcs/unsafe-code-guidelines/pull/5#discussion_r212703192).
+
+`usize` and C’s `unsized` are *not* equivalent.
+
+## Booleans
+The `bool` type has size 1 byte, but is not specified to be C-compatible. This [discussion](https://github.com/rust-lang/rust/pull/46176#issuecomment-359593446) has more details. For Rust to support a platform, the platform’s C must have the boolean type be a byte, where `true = 1` and `false = 0`.
+
+For full ABI compatibility details, see [Gankro’s post] (https://gankro.github.io/blah/rust-layouts-and-abis/#the-layoutsabis-of-builtins).
+## Relationship to C integer hierarchy
+C integers:
+- char: at least 8 bits
+- short: at least 16 bits (also at least a char)
+- int: at least a short (intended to be a native integer size)
+- long: at least 32 bits (also at least an int)
+- long long: at least 64 bits (also at least a long)
+The C integer types specify a minimum size, but not the exact size. For this reason, Rust integer types are not necessarily compatible with the “corresponding” C integer type. Instead, use the corresponding fixed size data types (e.g. `i64` in Rust would correspond to `int64_t` in C).
+
+## Controversies
+None
--- a/reference/src/representation/integers-floatingpoint.md
+++ b/reference/src/representation/integers-floatingpoint.md
@@ -24,16 +24,19 @@ Types `usize` and `isize` are committed to having the same size as a native poin
 - how much a pointer of a certain type can be offseted,
 - the maximum size of Rust objects (because size_of/size_of_val return `usize`),
 - the maximum number of elements in an array ([T; N: usize]),
-- `usize`/`isize` in C FFI are compatible with C's `uintptr_t` / `intptr_t`.
+- `usize`/`isize` in C FFI are compatible with C's `uintptr_t` / `intptr_t` (and have the same size and alignment).
 
-The maximum size of any single value must fit within `isize` to [ensure that pointer diff is representable](https://github.com/rust-rfcs/unsafe-code-guidelines/pull/5#discussion_r212703192).
+The maximum size of any single value must fit within `usize` to [ensure that pointer diff is representable](https://github.com/rust-rfcs/unsafe-code-guidelines/pull/5#discussion_r212703192).
 
 `usize` and C’s `unsized` are *not* equivalent.
 
 ## Booleans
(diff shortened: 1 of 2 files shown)


reference/src/representation/integers-floatingpoint.md
```diff
@@ -0,0 +1,47 @@
+# Representation of Boolean, Floating Point, and Integral Types
+This chapter represents the consensus from issue [#9]. It documents the memory layout and considerations for `bool`, `usize`, `isize`, floating point types, and integral types.
+[#9]: https://github.com/rust-rfcs/unsafe-code-guidelines/issues/9
+
+## Overview
+These are all scalar types, representing a single value. These types have no representation variants (no #[repr(C)] or #[repr(Rust)]. Their size is fixed and well-defined across FFI boundaries and map to their corresponding integral types in the C ABI.
+- `bool`: 1 byte
+  - any `bool` can be cast into an integer, taking on the values 1 (true) or 0 (false)
+- `usize`, `isize`: pointer-sized unsigned/signed integer type
```
gnzlbg: I think we do wanted to guarantee that `usize` has the same layout (size and alignment) as `uintptr_t` and `isize` as `intptr_t`.

reference/src/representation/integers-floatingpoint.md
```diff
@@ -0,0 +1,47 @@
+# Representation of Boolean, Floating Point, and Integral Types
+This chapter represents the consensus from issue [#9]. It documents the memory layout and considerations for `bool`, `usize`, `isize`, floating point types, and integral types.
+[#9]: https://github.com/rust-rfcs/unsafe-code-guidelines/issues/9
+
+## Overview
+These are all scalar types, representing a single value. These types have no representation variants (no #[repr(C)] or #[repr(Rust)]. Their size is fixed and well-defined across FFI boundaries and map to their corresponding integral types in the C ABI.
+- `bool`: 1 byte
```
gnzlbg: I think we do want to guarantee here that it has the same layout as `_Bool` as well as a size of 1 byte.

reference/src/representation/integers-floatingpoint.md
```diff
@@ -0,0 +1,47 @@
...
+  - Rust char: 32-bit
+  - not ABI compatible
+  - represents [Unicode scalar value](http://www.unicode.org/glossary/#unicode_scalar_value)
+
+##`usize`/`isize`
+Types `usize` and `isize` are committed to having the same size as a native pointer on the platform. The representation of `usize` determines the following:
+- how much a pointer of a certain type can be offseted,
+- the maximum size of Rust objects (because size_of/size_of_val return `usize`),
+- the maximum number of elements in an array ([T; N: usize]),
+- `usize`/`isize` in C FFI are compatible with C's `uintptr_t` / `intptr_t`.
```
gnzlbg: This means that `usize`/`isize` have the same size and alignment as `uintptr_t` and `intptr_t`, it might be worth it to spell this out.

reference/src/representation/integers-floatingpoint.md
```diff
@@ -0,0 +1,47 @@
...
+
+The maximum size of any single value must fit within `isize` to [ensure that pointer diff is representable](https://github.com/rust-rfcs/unsafe-code-guidelines/pull/5#discussion_r212703192).
```
gnzlbg: Above the second point states that "the maximum size of Rust objects" is determined by `usize`, but here it is stated that "The maximum size of any single value must fit within `isize`". 

It is a bit unclear to me what the difference between value and object is here. We probably want to use consistent terminology here but sadly #40 is not merged yet.

reference/src/representation/integers-floatingpoint.md
```diff
@@ -0,0 +1,47 @@
...
+
+`usize` and C’s `unsized` are *not* equivalent.
```
gnzlbg: What does this statement mean ?

reference/src/representation/integers-floatingpoint.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Booleans
+The `bool` type has size 1 byte, but is not specified to be C-compatible. This [discussion](https://github.com/rust-lang/rust/pull/46176#issuecomment-359593446) has more details. For Rust to support a platform, the platform’s C must have the boolean type be a byte, where `true = 1` and `false = 0`.
```
gnzlbg: I thought the consensus on that PR was a bit different. 

IIUC we had consensus that:

* `bool` has a size of `1`
* `true as i8 == 1` and `false as i8 == 0`
* Iff the C platform has a `_Bool`, this `_Bool` has a size of `1`, and the integer representation of `_Bool` is `0 == (int)false` and `1 == (int)true`:
  * then (note: currently, this is the case for all platforms that Rust supports):
    * `bool` is usable on C FFI
    * `bool` has the same layout as `_Bool`
  * otherwise:
    * `bool` is not a proper ctype (not C FFI safe, etc.). This does not mean that it is impossible to do C FFI with the platform. It only means that one cannot use `bool` when doing so.

rkruppe: The team decision in the linked comment treats the possibility of a C platform where `_Bool` is weird in the opposite fashion. Quoting:

> bool has the same representation as the platform's _Bool type.

> on every platform we currently support, this means [...]

It's not clear to me whether the third possibility of simply not supporting such weird C platforms (in the same way we don't support `CHAR_BITS != 8`) was explicitly considered, but I really don't see how what you wrote here can be interpreted into this decision.

gnzlbg: > It's not clear to me whether the third possibility of simply not supporting such weird C platforms (in the same way we don't support CHAR_BITS != 8) was explicitly considered, 

I think it was tangentially addressed here: https://github.com/rust-lang/rust/pull/46176#issuecomment-358800590

The argument was that we don't want people to create and use a `c_bool` out of portability concerns for platforms we might never support. I don't think this means that it is impossible to do C FFI with such a platform, so requiring this would be too strong. We had a tangential discussion (here: https://github.com/rust-rfcs/unsafe-code-guidelines/pull/46) and on Zulip and IIRC the consensus was that it would just mean that `bool` would not do what you intend it to do when used on C FFI - warning about an `improper_ctype` on such platforms would be a potential way forward. 

IIRC @briansmith argued in that PR that they often use Rust C FFI to interface with binary objects that were not produced by C, and that requiring this from the C platform was too strong of a requirement.

rkruppe: I remember all these discussions, but at the end of the day we have an explicit T-lang T-compiler decision (_Bool == Rust bool, so it always does the right thing in C FFI), so I don't know how you can claim there's consensus for something incompatible with that decision.

gnzlbg: > so I don't know how you can claim there's consensus for something incompatible with that decision.

If this should exclusively stick to that particular discussion and ignore all discussions that have happened since, then this document should probably only say something like:

> Rust `bool` has the same layout as `_Bool`. 
> 
> Note: on all platforms that Rust currently supports, `sizeof(_Bool) == 1`, `1 == (int)true`, and `0 == (int)false`.
> 
> Note: if the platform does not have a C ABI, or this C ABI does not have a `_Bool`, the layout of `bool` as well as the result of casting a `bool` to an integer are unspecified.

rkruppe: It's not so much "stick to just that old discussion", if other aspects of the matter have been resolved to our satisfaction since then, we can documented them to. I can also imagine bringing new arguments to the attention of the teams and asking them to re-evaluate. But I do not think it is at all acceptable to unilaterally decide that that decision was wrong and proceed to document something incompatible with that decision.

rkruppe: Keep in mind that saying "size_of bool == 1" or specifying the value of true/false as byte would not be guaranteeing *more*, but rather guaranteeing something *different*. That's why I keep saying "incompatible".

gnzlbg: That issue was closed in favor of https://github.com/rust-lang/rust/pull/46156

rkruppe: Then either there is a contradiction somewhere and T-lang T-compiler need to clear it up, or the docs modified in https://github.com/rust-lang/rust/pull/46156 are only intended to cover the platforms we currently support, while the UCG documents take a broader view.

gnzlbg: I think this is worth discussing in the next meeting.

reference/src/representation/integers-floatingpoint.md
nikomatsakis: Presumably that on most 64-bit platforms, `unsized` is 32 bits. Worth spelling out though.

reference/src/representation/integers-floatingpoint.md
nikomatsakis: Argh, there is something of a contradiction, isn't there? Still, I think the intent was pretty clearly documented in the comments on https://github.com/rust-lang/rust/pull/46176, and it's more that the docs in #46156 are incomplete and should be clarified. I don't recall, did we discuss this in our last meeting?

RalfJung: Following meeting consensus, this should be updated as described in https://github.com/rust-rfcs/unsafe-code-guidelines/issues/53#issuecomment-447050232, right?

reference/src/representation/integers-floatingpoint.md
avadacatavra: I grabbed this from https://github.com/rust-rfcs/unsafe-code-guidelines/issues/9#issuecomment-417406979

gnzlbg: @nikomatsakis that looks like a typo.

PR #51, opened 2018-11-29
Added representation of pointer types.

### LINKED ISSUE #16: Representation of Rust references (`&T`, `&mut T`) and raw pointers (`*const T, `*mut T`)
Representation of Rust references:

- Are `&T` and `&mut T` guaranteed to be a pointer?
- Must always be aligned, non-null
- Guaranteed to be ABI compatible with C pointer types ("in every way?")
    - presuming that the referent is compatible, of course

Representation of raw pointers:

- Guaranteed to be ABI compatible with C pointer types
    - presuming that the referent is compatible, of course

Other factors:

- Considerations for storing things in the low bits of pointers
    - safe with raw pointers, not safe with references (from [this comment](https://github.com/rust-rfcs/unsafe-code-guidelines/issues/16#issuecomment-417406207))

asajeffrey:
Fixes #16

### DIFF
--- a/reference/src/representation/pointers.md
+++ b/reference/src/representation/pointers.md
@@ -2,29 +2,35 @@
 
 ### Terminology
 
-Reference types are types of the form `&T` or `&mut T`.
+Reference types are types of the form `&T`, `&mut T` or `&dyn T`.
 
 Raw pointer types are types of the form `*const T` or `*mut T`.
 
 ### Representation
 
-The alignment of reference and raw pointer types is the word size.
+The alignment of `&T`, `&mut T`, `*const T` and `*mut T` are the same,
+and are at least the word size.
+
+* If `T` is a trait, then the alignment of `&dyn T` is the word size.
+* If `T` is a sized type then the alignment of `&T` is the word size.
+* The alignment of `&[T]` is the word size.
 
 The sizes of `&T`, `&mut T`, `*const T` and `*mut T` are the same,
 and are at least one word.
 
-* If `T` is a trait, then the size of `&T` is two words.
+* If `T` is a trait, then the size of `&dyn T` is two words.
 * If `T` is a sized type then the size of `&T` is one word.
 * The size of `&[T]` is two words.
-
-Note that we do not make any guarantees about the sizes of
-multi-trait objects `&(dyn T + U)` or references to other dynamically sized types.
 
 ### Notes
 
 The representations of `&T` and `&mut T` are the same.
 
-The representation of `&T` when `T` is a trait is the same as that of:
+We do not make any guarantees about the representation of
+multi-trait objects `&(dyn T + U)` or references to other dynamically sized types,
+other than that they are at least word-aligned, and have size at least one word.
+
+The representation of `&dyn T` when `T` is a trait is the same as that of:
 ```rust
 #[repr(C)]
 struct DynObject {
--- a/reference/src/representation/pointers.md
+++ b/reference/src/representation/pointers.md
@@ -15,6 +15,7 @@ and are at least the word size.
 * If `T` is a sized type then the alignment of `&T` is the word size.
 * The alignment of `&[T]` is the word size.
 * The alignment of `&str` is the word size.
+* Alignment in other cases may be more than the word size (e.g., for other dynamically sized types).
 
 The sizes of `&T`, `&mut T`, `*const T` and `*mut T` are the same,
 and are at least one word.
@@ -23,6 +24,7 @@ and are at least one word.
 * If `T` is a sized type then the size of `&T` is one word.
 * The size of `&[T]` is two words.
 * The size of `&str` is two words.
+* Size in other cases may be more than one word (e.g., for other dynamically sized types).
 
 ### Notes
 
--- a/reference/src/representation/pointers.md
+++ b/reference/src/representation/pointers.md
@@ -17,6 +17,9 @@ and are at least one word.
 * If `T` is a sized type then the size of `&T` is one word.
 * The size of `&[T]` is two words.
 
+Note that we do not make any guarantees about the sizes of
+multi-trait objects `&(dyn T + U)`or references to dynamically sized types.
+
 ### Notes
 
 The representations of `&T` and `&mut T` are the same.
@@ -38,7 +41,3 @@ struct Slice<T> {
   len: usize,
 }
 ```
-
-The validity requirements of `&T` include that all values are non-null, which
-impacts niche optimizations and hence representation of types which include `&T`.
-In particular, `Option<&T>` is one word.
--- a/reference/src/representation/pointers.md
+++ b/reference/src/representation/pointers.md
@@ -14,6 +14,7 @@ and are at least the word size.
 * If `T` is a trait, then the alignment of `&dyn T` is the word size.
 * If `T` is a sized type then the alignment of `&T` is the word size.
 * The alignment of `&[T]` is the word size.
+* The alignment of `&str` is the word size.
 
 The sizes of `&T`, `&mut T`, `*const T` and `*mut T` are the same,
 and are at least one word.
@@ -21,6 +22,7 @@ and are at least one word.
 * If `T` is a trait, then the size of `&dyn T` is two words.
 * If `T` is a sized type then the size of `&T` is one word.
 * The size of `&[T]` is two words.
+* The size of `&str` is two words.
 
 ### Notes
 
@@ -47,3 +49,5 @@ struct Slice<T> {
   len: usize,
 }
 ```
+
+The representation of `&str` is the same as that of `&[u8]`.
--- a/reference/src/representation/pointers.md
+++ b/reference/src/representation/pointers.md
@@ -29,7 +29,7 @@ The representation of `&T` when `T` is a trait is the same as that of:
 #[repr(C)]
 struct DynObject {
   data: &u8,
-  vtable: &usize,
+  vtable: &u8,
 }
 ```
 
--- a/reference/src/representation/pointers.md
+++ b/reference/src/representation/pointers.md
@@ -17,7 +17,7 @@ and are at least one word.
 * If `T` is a sized type then the size of `&T` is one word.
 * The size of `&[T]` is two words.
 
-Raw pointer types have no requirements of their representation.
+### Notes
 
 The representations of `&T` and `&mut T` are the same.
 
@@ -38,8 +38,6 @@ struct Slice<T> {
   len: usize,
 }
 ```
-
-### Notes
 
 The validity requirements of `&T` include that all values are non-null, which
 impacts niche optimizations and hence representation of types which include `&T`.
--- a/reference/src/representation/pointers.md
+++ b/reference/src/representation/pointers.md
@@ -5,6 +5,7 @@
 Reference types are types of the form `&T`, `&mut T` or `&dyn T`.
 
 Raw pointer types are types of the form `*const T` or `*mut T`.
+We write `*T` when the mutability attribute is unimportant.
 
 ### Representation
 
@@ -28,7 +29,7 @@ and are at least one word.
 
 ### Notes
 
-The representations of `&T` and `&mut T` are the same.
+The representations of `&T`, `&mut T` and `*T` are the same.
 
 We do not make any guarantees about the representation of
 multi-trait objects `&(dyn T + U)` or references to other dynamically sized types,
@@ -38,8 +39,8 @@ The representation of `&dyn T` when `T` is a trait is the same as that of:
 ```rust
 #[repr(C)]
 struct DynObject {
-  data: &u8,
-  vtable: &u8,
+  data: *u8,
+  vtable: *u8,
 }
 ```
 
@@ -47,7 +48,7 @@ The representation of `&[T]` is the same as that of:
 ```rust
 #[repr(C)]
 struct Slice<T> {
-  ptr: &T,
+  ptr: *T,
   len: usize,
 }
 ```
--- a/reference/src/representation/pointers.md
+++ b/reference/src/representation/pointers.md
@@ -18,7 +18,7 @@ and are at least one word.
 * The size of `&[T]` is two words.
 
 Note that we do not make any guarantees about the sizes of
-multi-trait objects `&(dyn T + U)`or references to other dynamically sized types.
+multi-trait objects `&(dyn T + U)` or references to other dynamically sized types.
 
 ### Notes
 
--- a/reference/src/representation/pointers.md
+++ b/reference/src/representation/pointers.md
@@ -1,0 +1,46 @@
+# Representation of reference and pointer types
+
+### Terminology
+
+Reference types are types of the form `&T` or `&mut T`.
+
+Raw pointer types are types of the form `*const T` or `*mut T`.
+
+### Representation
+
+The alignment of reference and raw pointer types is the word size.
+
+The sizes of `&T`, `&mut T`, `*const T` and `*mut T` are the same,
+and are at least one word.
+
+* If `T` is a trait, then the size of `&T` is two words.
+* If `T` is a sized type then the size of `&T` is one word.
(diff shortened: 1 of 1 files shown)


asajeffrey:
On thinking about it, now that the non-zero and alignment requirements are punted to validity, we're just left with size and alignment. Everything else can be part of the notes.

asajeffrey:
Do we need to add text about the ABI?

reference/src/representation/pointers.md
```diff
@@ -0,0 +1,44 @@
...
+Raw pointer types are types of the form `*const T` or `*mut T`.
+
+### Representation
+
+The alignment of reference and raw pointer types is the word size.
+
+The sizes of `&T`, `&mut T`, `*const T` and `*mut T` are the same,
+and are at least one word.
+
+* If `T` is a trait, then the size of `&T` is two words.
```
gnzlbg: Is this `true` ? I thought there was a mention somewhere that we might want to allow for bigger traits in the future (and this would prevent that).

reference/src/representation/pointers.md
```diff
@@ -0,0 +1,44 @@
...
+### Notes
+
+The representations of `&T` and `&mut T` are the same.
+
+The representation of `&T` when `T` is a trait is the same as that of:
+```rust
+#[repr(C)]
+struct DynObject {
+  data: &u8,
+  vtable: &usize,
```
gnzlbg: Why is this `vtable: &usize` ?

reference/src/representation/pointers.md
```diff
@@ -0,0 +1,44 @@
...
+The representation of `&[T]` is the same as that of:
+```rust
+#[repr(C)]
+struct Slice<T> {
+  ptr: &T,
+  len: usize,
+}
+```
+
+The validity requirements of `&T` include that all values are non-null, which
```
gnzlbg: I would just remove these notes, they are not part of the representation, and how pointers, references, and `Option<&T>` relate to each other is not something that will be forgotten during the validity discussions.

reference/src/representation/pointers.md
rkruppe: For multi-trait objects like `dyn Foo + Bar` (where neither is a marker trait), yeah. Arguably that would not fall under "`T` is **a** trait". By the same argument, the current wording would exclude e.g. `dyn Any+Send`. Would be good to explicitly note this limitation of the guarantee.

reference/src/representation/pointers.md
asajeffrey: To emphasize the fact that it's non-zero and word-aligned. Not part of the representation, but I believe we're going to be requiring this when we come to validity.

reference/src/representation/pointers.md
asajeffrey: Done.

reference/src/representation/pointers.md
asajeffrey: Done.

reference/src/representation/pointers.md
rkruppe: This seems like a very confusing way to make that point, and since it's not even really part of what this write-up is supposed to be about at this point, nor particularly useful, I'd really prefer not doing that. Just `*const ()` is good enough for now.

asajeffrey:
For slices, we could say something like the representation of `&[T]` is the same as that of:
```rust
#[repr(C)]
struct Slice {
  ptr: &[T;N],
  len: usize,
}
```
and then when we come to validity, say that `N` has got to be the value in the `len` field. This would cope with nasty cases, e.g. when `T` is uninhabited (and hence `&T` is uninhabited, but `&[T]` is inhabited).

reference/src/representation/pointers.md
asajeffrey: I made them both `&u8`.

asajeffrey:
I think I dealt with a ll the outstanding comments. Can we squash and merge?

reference/src/representation/pointers.md
```diff
@@ -0,0 +1,43 @@
...
+
+### Terminology
+
+Reference types are types of the form `&T` or `&mut T`.
+
+Raw pointer types are types of the form `*const T` or `*mut T`.
+
+### Representation
+
+The alignment of reference and raw pointer types is the word size.
```
rkruppe: I just realized this might not actually be true for custom DSTs (e.g., consider a custom DST with `u64` metadata on a 32 bit target). Since the rest of this text is already careful to be forward compatible with custom DSTs, it would be a shame if this part wasn't.

reference/src/representation/pointers.md
```diff
@@ -0,0 +1,43 @@
...
+
+The sizes of `&T`, `&mut T`, `*const T` and `*mut T` are the same,
+and are at least one word.
+
+* If `T` is a trait, then the size of `&T` is two words.
+* If `T` is a sized type then the size of `&T` is one word.
+* The size of `&[T]` is two words.
+
+Note that we do not make any guarantees about the sizes of
+multi-trait objects `&(dyn T + U)`or references to dynamically sized types.
```
rkruppe: nit: "... or references to *other* dynamically sized types"

reference/src/representation/pointers.md
asajeffrey: Hmm, worth thinking about. On 32-bit targets, does `u64` have 64-bit alignment? Are there architectures with alignment larger than the word size?

rkruppe: Oh, u64 is not a good example, because on most (all?) 32 bit targets it has alignment 4. But there are definitely tons of types with super-word alignment -- anyone can define them with `#[repr(align(N))]`, and vector types such as `__m128` usually require natural alignment (i.e., size = align).

reference/src/representation/pointers.md
asajeffrey: Fixed.

reference/src/representation/pointers.md
asajeffrey: So the question is do we want to tie our hands and say that metadata of custom DSTs is at most word-aligned? The trade-offs don't seem obvious to me, I'm trying to come up with an example and the best I can do is something like:
```rust
  let a: Box<T> = Box::new(T::new());
  let b: Box<(*T, usize, usize)> = Box::new(Box::Into_raw(a), 0, 0);
  let c: &&U = unsafe { mem::transmute(Box::into_raw(b)) };
```
The question is can this be insta-UB if `U`s metadata is super-word aligned?

This is the simplest example I can think of.

ubsan: `u64` has 8-byte alignment on Windows and MacOS 32-bit.

gnzlbg: My two cents is that we shouldn't say anything about the size and alignment of references here. 

We should just say that the layout of references and pointers is the same as the layout of the following `repr(Rust)` type:

```rust
struct Layout<Metadata> {
    ptr: *mut (), 
    metadata: Metadata
}
```

We can then say that in `*mut T`/`&mut T`, if `T` is a trait / slice / ... then `Metadata`'s layout is the same as the layout of some other particular type, and the layout of references and pointers follows from that.

reference/src/representation/pointers.md
```diff
@@ -0,0 +1,43 @@
...
+### Notes
+
+The representations of `&T` and `&mut T` are the same.
+
+The representation of `&T` when `T` is a trait is the same as that of:
+```rust
+#[repr(C)]
+struct DynObject {
+  data: &u8,
+  vtable: &u8,
```
gnzlbg: I still think that this notation using `&u8` is a bit weird =/

I've made a suggestion above, but maybe we should discuss this in today's meeting ?

reference/src/representation/pointers.md
rkruppe: When we have custom DSTs in the language, then *all* pointers are `... { ptr: *mut (), metadata: T::Metadata }`, as custom DSTs generalize slices, trait objects, and sized types (we can still spell this out for clarity but there's no need to have two things called `Metadata` or to introduce this `===` relation). But we don't have custom DSTs yet, so it feels very weird to define layout in terms of a concept the language doesn't have yet, even if we can technically avoid any reference to a specific non-existent `DynamicallySized` trait.

> Since I've used raw pointers, we have to say something about the size and alignment of raw pointers here, but we already said that their size is the same as usize, and for alignment we can just say that the raw pointer alignment is an implementation-defined multiple of their size.

Wait, what? If the alignment is not described yet, it should be described if we can at all get consensus for it. As far as I know, that alignment should be *exactly* the size. But even if that is controversial, it's the other way around, size has to be a multiple of the alignment (because element i of an array is at byte offset `i * alignof(T)`).

gnzlbg: >  But we don't have custom DSTs yet, so it feels very weird to define layout in terms of a concept the language doesn't have yet, even if we can technically avoid any reference to a specific non-existent DynamicallySized trait.

Even without the concept of custom DSTs, we do have the concept of generic structs, and their layout, and those can be used to specify the layout of slices and trait objects.

rkruppe: Sure, but unifying these two layouts into one generic struct, and the naming of the type parameter, gestures heavily at the concept of custom DSTs (would anyone make this suggestion if they didn't know of the generalization that custom DSTs bring?), without really buying us anything right now.

reference/src/representation/pointers.md
```diff
@@ -0,0 +1,43 @@
...
+Raw pointer types are types of the form `*const T` or `*mut T`.
+
+### Representation
+
+The alignment of reference and raw pointer types is the word size.
+
+The sizes of `&T`, `&mut T`, `*const T` and `*mut T` are the same,
+and are at least one word.
+
+* If `T` is a trait, then the size of `&T` is two words.
```
RalfJung: Should be using `dyn T`. Also `T` is a *type* anywhere else, maybe just use `&dyn Trait`?

reference/src/representation/pointers.md
```diff
@@ -0,0 +1,43 @@
...
+* If `T` is a sized type then the size of `&T` is one word.
+* The size of `&[T]` is two words.
+
+Note that we do not make any guarantees about the sizes of
+multi-trait objects `&(dyn T + U)`or references to other dynamically sized types.
```
RalfJung:
```suggestion
multi-trait objects `&(dyn T + U)` or references to other dynamically sized types.
```

reference/src/representation/pointers.md
RalfJung: Also, what about raw DST pointers?  We should clarify that then all pointers here become raw pointers. Or just the data pointer becomes a raw pointer? (That last question totally overlaps with validity though.)

reference/src/representation/pointers.md
asajeffrey: OK, I reworded this to say that reference types have at least word alignment, and have exactly word alignment in the case of references to sized types, `&[T]`, `&str` or `&dyn Trait`.

reference/src/representation/pointers.md
asajeffrey: There's already text saying that `*T` has the same representation as `&T`, which I think covers it.

reference/src/representation/pointers.md
asajeffrey: There are a few other places where `T` is a trait, e.g. `&(dyn T + U)`. I put in all the `dyn`s I could find.

asajeffrey:
OK, try again... are we okay with merging now?

reference/src/representation/pointers.md
RalfJung: Hm, okay this works for size+alignment and I guess the rest is up to validity invariants.

reference/src/representation/pointers.md
```diff
@@ -0,0 +1,53 @@
...
+
+### Representation
+
+The alignment of `&T`, `&mut T`, `*const T` and `*mut T` are the same,
+and are at least the word size.
+
+* If `T` is a trait, then the alignment of `&dyn T` is the word size.
+* If `T` is a sized type then the alignment of `&T` is the word size.
+* The alignment of `&[T]` is the word size.
+* The alignment of `&str` is the word size.
```
nikomatsakis: Is the intention here to leave room for future developments? If so, it is probably saying explicitly what we are leaving room for. Perhaps something like:

---

As described above, `&T` values are typically word aligned. However, this is not guaranteed to be true for all types `T`. For example, we expect to ultimately support some form of ["Custom Dynamically Sized Types"][cdst], and in that case a type `T` may define custom metadata to be attached to a reference `&T` which could potentially affect its alignment.

[cdst]: https://internals.rust-lang.org/t/custom-dst-discussion/4842

reference/src/representation/pointers.md
```diff
@@ -0,0 +1,53 @@
...
+  data: &u8,
+  vtable: &u8,
+}
+```
+
+The representation of `&[T]` is the same as that of:
+```rust
+#[repr(C)]
+struct Slice<T> {
+  ptr: &T,
```
nikomatsakis: Question: if the length is zero, then what is `ptr`? Perhaps `*const T` is a better choice -- certainly it will fit the "validity invariants" better (eg, I don't think there is any requirement that it be "dereferenceable" in the event that length is zero). However, [`from_raw_parts`](https://doc.rust-lang.org/std/slice/fn.from_raw_parts.html) does state:

> data must be non-null and aligned, even for zero-length slices. One reason for this is that enum layout optimizations may rely on references (including slices of any length) being aligned and non-null to distinguish them from other data. You can obtain a pointer that is usable as data for zero-length slices using `NonNull::dangling()`.

It seems worth linking to `slice::from_raw_parts` and reproducing some of that text (although arguably those are things best left for the next discussion?).

gnzlbg: While I don't like the use of references here much [0] the guarantee here is that "the _representation_ of `&[T]` is the same as of `Slice<T>`", - that is, the `Slice<T>` struct only applies to the representation / layout of `&[T]` and not to its validity. 

Ideally, we would use the same examples to show validity and representation, but we haven't settled the validity of these yet (it might well be that we need different "examples" for the `&[T]` and the `*[T]` cases). I think it is worth it to wait until the validity of these is settled before trying to "unify" all the examples.

---

[0] Personally, I don't like the use of `&T` here much, it has too many connotations, e.g., w.r.t. validity and safety, and we don't really care about them when just talking about the representation. It might be worth it to just say:

The representation of `&[T]` is the same as that of:  

```
struct Slice {
    ptr: *const (),
    len: usize,
}
```

There is no need to make `Slice<T>` generic since `T: Sized`.

reference/src/representation/pointers.md
asajeffrey: Is that note enough, or do we need to say more?

nikomatsakis: Well, I didn't find it =) maybe a bullet point like:

* Alignment in other cases may vary (e.g., for other dynamically sized types).

reference/src/representation/pointers.md
asajeffrey: Like I said above, we could also use
```rust
struct Slice<T> {
  ptr: &[T;N],
  len: usize,
}
```
and in practice, `N` and `len` are the same.

All these defns are equivalent, it's just a case of which is the more human-readable.

reference/src/representation/pointers.md
asajeffrey: Done. I added a matching comment to size, since the two defns line up.

reference/src/representation/pointers.md
asajeffrey: I added a comment that `*T` means `*const T` or `*mut T` when the mutability is unimportant, and used `*T` rather than `&T`. On  zulip, @gnzlbg was okay with this, have we finally closed the last issue?

asajeffrey:
OK, asking yet again, can we squash and merge?

RalfJung:
No objections from my side.

asajeffrey:
@avadacatavra I think you can squash and merge.

PR #54, opened 2018-12-10
Area proposal: validity invariants

RalfJung:
Which invariants derived from types are there that the compiler expects to be *always* maintained, and (equivalently) that unsafe code must *always* uphold (or else cause undefined behavior)?  This is what is called "validity invariant" in [my blog post](https://www.ralfj.de/blog/2018/08/22/two-kinds-of-invariants.html).

### Goals

* For every primitive type, determine which assumptions (if any) the compiler   makes about values *not* occurring at that type (serving as a lower bound for   what to declare invalid), and determine which popular patterns in unsafe code   might create "interesting" values of this type that safe code cannot create on   its own (serving as an upper bound for how much we want to declare invalid).   Both of these bounds are soft, but informative.
* Based on that, map out a design space of invariants that seem reasonable.
* Determine when exactly the validity invariant is assumed to hold.

### DIFF
--- a/active_discussion/validity.md
+++ b/active_discussion/validity.md
@@ -57,45 +57,46 @@ values.
 To start, we will create threads for each major category of types.
 
 * Integers and floating point types
-** Do we allow values that contain uninitialized bits?  If yes, what are the
-   rules for arithmetic and logical operations involving uninitialized bits,
-   e.g. in cases like `x * 0`?
+
+  * Do we allow values that contain uninitialized bits?  If yes, what are the
+    rules for arithmetic and logical operations involving uninitialized bits,
+    e.g. in cases like `x * 0`?
 
 * Raw pointers
-** Do we allow values that contain uninitialized bits?
-** Are there any requirements on the metadata?
+  * Do we allow values that contain uninitialized bits?
+  * Are there any requirements on the metadata?
 
 * References
-** Presumably, references must be non-NULL.
-** They probably also must be aligned, but is that required every time a
-   reference is taken?  Also see the [ongoing discussion in RFC 2582][RFC2582].
-** Can there ever be uninitialized bits in a reference?
-** Do they have to be dereferencable?  What exactly does that even mean?
-** Does `&[mut] T` have to point to data that is valid at `T`?  This interacts
-   with the question of whether `&*x` is allowed when `x` is a well-aligned
-   non-null dereferencable pointer that points to invalid data.
-** Out of scope: aliasing rules
+  * Presumably, references must be non-NULL.
+  * They probably also must be aligned, but is that required every time a
+    reference is taken?  Also see the [ongoing discussion in RFC 2582][RFC2582].
+  * Can there ever be uninitialized bits in a reference?
+  * Do they have to be dereferencable?  What exactly does that even mean?
+  * Does `&[mut] T` have to point to data that is valid at `T`?  This interacts
+    with the question of whether `&*x` is allowed when `x` is a well-aligned
+    non-null dereferencable pointer that points to invalid data.
+  * Out of scope: aliasing rules
 
 * Function pointers
-** Presumably, these must be non-NULL.  Anything else?  Can there ever be
-   uninitialized bits?
+  * Presumably, these must be non-NULL.  Anything else?  Can there ever be
+    uninitialized bits?
 
 * Unions
-** Do we make any restrictions here, or are unions just "bags of bits" that may
-   contain anything?  That would mean we can do no layout optimizations.
+  * Do we make any restrictions here, or are unions just "bags of bits" that may
+    contain anything?  That would mean we can do no layout optimizations.
 
 * Enums
-** Is there anything to say besides: The discriminant must be valid, and all
-   fields of the active variant must be valid at their respective types?
-** The padding between fields can be anything, including uninitialized.
+  * Is there anything to say besides: The discriminant must be valid, and all
+    fields of the active variant must be valid at their respective types?
+  * The padding between fields can be anything, including uninitialized.
 
 * Structs, tuples, arrays and all other aggregates (closures, ...)
-** Is there anything to say besides: All fields must be valid at their
-   respective types?
-** The padding between fields can be anything, including uninitialized.  It was
-** [recently determined][generators-maybe-uninit] that generators behave
-** different from other aggregates here.  Are we okay with that?  Should we push
-** for generator fields to reflect this in their types?
+  * Is there anything to say besides: All fields must be valid at their
+    respective types?
+  * The padding between fields can be anything, including uninitialized.  It was
+  * [recently determined][generators-maybe-uninit] that generators behave
+  * different from other aggregates here.  Are we okay with that?  Should we push
+  * for generator fields to reflect this in their types?
 
 [RFC2582]: https://github.com/rust-lang/rfcs/pull/2582
 [generators-maybe-uninit]: https://github.com/rust-lang/rust/pull/56100
--- a/active_discussion/validity.md
+++ b/active_discussion/validity.md
@@ -2,8 +2,8 @@
 
 This discussion is meant to focus on the question: Which invariants derived from
 types are there that the compiler expects to be *always* maintained, and
-(equivalently) that unsafe code must *always* uphold.  This is what is called
-"validity invariant" in
+(equivalently) that unsafe code must *always* uphold (or else cause undefined
+behavior)?  This is what is called "validity invariant" in
 [Ralf's blog post](https://www.ralfj.de/blog/2018/08/22/two-kinds-of-invariants.html),
 but we might also decide to change that name.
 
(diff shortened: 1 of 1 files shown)


active_discussion/validity.md
```diff
@@ -0,0 +1,120 @@
...
+
+* Enums
+  * Is there anything to say besides: The discriminant must be valid, and all
+    fields of the active variant must be valid at their respective types?
+  * The padding between fields can be anything, including uninitialized.
+
+* Structs, tuples, arrays and all other aggregates (closures, ...)
+  * Is there anything to say besides: All fields must be valid at their
+    respective types?
+  * The padding between fields can be anything, including uninitialized.  It was
```
rkruppe: Are generator fields (and more generally generator contents) exposed to user code at all? If not, this doesn't seem like something UCG have to say anything about, just pure rustc implementation details.

RalfJung: > Are generator fields (and more generally generator contents) exposed to user code at all? 

They are not, AFAIK.

But I think when we are in the business of type invariants, we should not forget about types like closures or generators.  Just because the user cannot write them directly, doesn't mean their invariants are not important to know -- and compiler devs also need guidance for how to approach validity invariants.

And, of course, unsafe code could start to play games with these fields.

active_discussion/validity.md
```diff
@@ -0,0 +1,120 @@
...
+    non-null dereferencable pointer that points to invalid data.
+  * Out of scope: aliasing rules
+
+* Function pointers
+  * Presumably, these must be non-NULL.  Anything else?  Can there ever be
+    uninitialized bits?
+
+* Booleans
+  * Is there anything to say besides: A `bool` must be `0x0` or `0x1`?  Do we
+    allow the remaining bits to be uninitialized?
```
gnzlbg: According to C, ` _Bool` has two values and `sizeof(_Bool) - 2` "trap" representations. We might need to reconcile Rust's "uninitialized" with C's "trap" representation terminology somewhere as part of the validity discussions. IIRC the behavior of C's "trap" representations is undefined, so we can probably define it however we want here, but then that's a requirement on the C platform to play ball.

active_discussion/validity.md
```diff
@@ -0,0 +1,120 @@
...
+
+* Unions
+  * Do we make any restrictions here, or are unions just "bags of bits" that may
+    contain anything?  That would mean we can do no layout optimizations.
+
+* Enums
+  * Is there anything to say besides: The discriminant must be valid, and all
+    fields of the active variant must be valid at their respective types?
+  * The padding between fields can be anything, including uninitialized.
```
gnzlbg: What is it there beyond uninitialized? Or what else could this padding be?

active_discussion/validity.md
rkruppe: It's certainly a related issue, but one that interacts with generator implementation details much more than with user-facing guarantees or semantics of Rust-the-language, so I'm not sure the group of people here is the best to discuss it. So I think it would be reasonable to punt on it to reduce the scope of the validity invariant discussion and get it done more quickly. But we could also open the issue for generators and maybe end up wrapping up the discussion without any decisions being made about generators.

> And, of course, unsafe code could start to play games with these fields.

Regardless of how generators end up being organized internally, I am quite certain they should be opaque, i.e., user code should not be allowed to poke at their contents in such a way.

gnzlbg: I always reasoned about closures as being equivalent to a `struct` that one could write by hand (and with `feature(fn_traits)` one actually can). The rules for building that struct, if they are ever specified, are part of the language, but the layout, validity, etc. of this struct would follow from the layout and validity invariants that we specify here for primitive types, `struct`s, etc. 

I would find it extremely weird if the same wouldn't be true for generators. That is, the exact type of the Generator would be unspecified, but this type would be a normal Rust type, e.g., a `struct` or an `union`, with normal Rust fields (`struct`, `union`s, primitive types, etc.).

active_discussion/validity.md
RalfJung: In my mental model, "trap representation" is pretty much equivalent with "indeterminate data" is equivalent with the "uninitialized bits".  I will add a remark along those lines.

Other than that, hold these thoughts until the actual discussion starts. :)

active_discussion/validity.md
RalfJung: I presume we don't *require* padding to be uninitialized.  They may also be initialized to some bit pattern.

IOW, using `U` to encode uninitialized bits, the following are both valid elements of `(u8, u16)`:
```
00000000UUUUUUUU0000000011111111
00000000111111110000000011111111
```

active_discussion/validity.md
```diff
@@ -0,0 +1,120 @@
...
+The validity invariant of a type is, basically, a set of bit patterns that is
+allowed to occur at that type.  ("Basically" because the invariant may also be
+allowed to depend on memory.)  To discuss this properly, we need to first agree
+on what "bit patterns" even are.  It is not enough to just consider sequences of
+0 and 1, because we also need to take uninitialized data into account.  For the
+purpose of this discussion, I think it is sufficient to consider every bit as
+being either 0, 1 or uninitialized.
+[That is not always sufficient](https://www.ralfj.de/blog/2018/07/24/pointers-and-bytes.html),
+but I think we can mostly ignore the extra complications introduced by pointer
+values.
```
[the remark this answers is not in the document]
RalfJung: What exactly do you mean by "this"?

Also, I think `const fn` is out-of-scope for this discussion area.

active_discussion/validity.md
RalfJung: > we could also open the issue for generators and maybe end up wrapping up the discussion without any decisions being made about generators.

That seems reasonable.  And I agree unsafe code shouldn't mess with implementation details of closures or generators, but that doesn't follow from them being in any way different than other `struct` -- it entirely follows from the fact that unsafe code must not make any assumption about that the struct (its fields etc) looks like.

"user code should not be allowed to poke at their contents in such a way" also applies to many, many other data types -- to pick an example, `HashMap` (or else we couldn't replace it by hashbrown as is being pursued right now).

active_discussion/validity.md
gnzlbg: What I tried to say is that maybe we should add a question: "Are Rust `uninitialized` bits equivalent to C's `trap` representations ? (example: `_Bool` ...)". I agree with you that the answer is probably yes, but that's something worth triple checking.

active_discussion/validity.md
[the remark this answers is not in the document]
gnzlbg: > What exactly do you mean by "this"?

I meant provenance, or in general, any other "data" attached to a bit pattern that affects its validity. 

For example, in C, a pointer to a valid allocation is valid and safe, a pointer to one past the end of a valid allocation is valid and unsafe (can't be safely dereferenced), and a pointer to two past the end of a valid allocation is invalid (out-of-bounds pointer arithmetic is undefined behavior). That is, pointers of all bit patterns can be valid, but the validity of a pointer does not have anything to do with its bit pattern.

I don't know whether it is a good idea to delay the discussion about validity of pointers and validity of integers representing memory addresses. These discussions interact with the validity of integers, references, etc.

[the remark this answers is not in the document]
RalfJung: LLVM has notions of dereferencable and in-bounds pointers which correspond to what you call safe and valid, respectively.

TBH I don't think we will make any headway on the topic of provenance.  There's not very much data as neither C nor LLVM have a precise model, and the informal models one can extrapolate from specifications and implementations are far too unconstrained.

I also don't think we need to make headway on provenance to talk about validity of pointers.  There are many interesting things to talk about even when pretending that pointers are integers.  Moreover, pointers that differ in their provenance but not in anything else make a difference when doing arithmetic and when accessing memory, but I don't think there is any reason to distinguish such pointers in terms of validity: I think we can safely declare that two pointers that map to the same integer are either both valid or both invalid.

active_discussion/validity.md
RalfJung: Given the lack of a formal model of C's trap representations, I think that's pretty much our choice. ;)

Rust `uninitialized` bits are like `poison` in LLVM, and can be soundly used to model `undef` in LLVM (but are not equivalent to `undef`).  That much I can say because `undef` and `poison` are much better specified than trap representations.

avadacatavra:
@RalfJung when you're happy with this, let's merge it

active_discussion/validity.md
```diff
@@ -0,0 +1,131 @@
+# Data type validity requirements
+
+This discussion is meant to focus on the question: Which invariants derived from
+types are there that the compiler expects to be *always* maintained, and
+(equivalently) that unsafe code must *always* uphold (or else cause undefined
+behavior)?  This is what is called "validity invariant" in
```
nikomatsakis: I'm curious whether there exist properties that unsafe code must always uphold but which it cannot **rely upon**. The idea would be: for now, we require that this is true, but we expect we may not require it forever. Technically, from the compiler's POV, we can tweak these predicates because that just makes less code UB -- but unsafe code could be relying on the fact that doing otherwise would be UB, and that could be a problem.

This seems like a more general thing -- it also e.g. arises in the reprsentation discussion -- but once we start getting into invariants, I expect we should think carefully about this distinction.

I would sort of like this to be noted somewhere as a kind of "cross-cutting concern" -- i.e., for each area, we are aiming to define these two predicates (what unsafe code must uphold, and what it can rely on) somewhat separately.

RalfJung: I agree in general.  We want some "wiggle room" in our definition of what unsafe code can do, certainly when it comes to the safety invariant.  I am less sure for the validity invariant.

My main problem with this is that I don't have a good understanding yet of how this works formally.  That's mostly because I did not spend enough time to think about this.

RalfJung:
> when you're happy with this, let's merge it

done

PR #56, opened 2018-12-14
add glossary.md and define some invariants

RalfJung:
People asked for definitions of what validity and safety invariants are, here you go.

### DIFF
--- a/reference/src/glossary.md
+++ b/reference/src/glossary.md
@@ -31,7 +31,7 @@ fn main() { unsafe {
 The *safety* invariant is an invariant that safe code may assume all data to uphold.
 This invariant is used to justify which operations safe code can perform.
 The safety invariant can be temporarily violated by unsafe code, but must always be upheld when interfacing with unknown safe code.
-It is not relevant when arguing whether some *program* has UB, but it is relevant when arguing whether some code safely encapsulates its unsafety -- IOW, it is relevant when arguning whether some *library* can be used by safe code to *cause* UB.
+It is not relevant when arguing whether some *program* has UB, but it is relevant when arguing whether some code safely encapsulates its unsafety -- IOW, it is relevant when arguing whether some *library* can be used by safe code to *cause* UB.
 
 In terms of code, some data computed by `TERM` (possibly constructed from some `arguments` that can be *assumed* to satisfy the safety invariant) is valid at type `T` if and only if the following library function can be safely exposed to arbitrary (safe) code as part of the public library interface:
 ```rust
--- a/reference/src/glossary.md
+++ b/reference/src/glossary.md
@@ -2,10 +2,10 @@
 
 #### Interior mutability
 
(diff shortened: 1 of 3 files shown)


reference/src/glossary.md
```diff
@@ -0,0 +1,55 @@
+## Glossary
+
+#### Interior mutability
+
+*Interior Mutability* means mutating memory where there also exists a live shared reference pointing to the same memory.
```
rkruppe: I'm the last person affected by this, but as far as I know "live" as used here is jargon too. Would it make sense to explain it too, link to an explanation, or paraphrase this informally?

reference/src/glossary.md
```diff
@@ -0,0 +1,55 @@
...
+This propagates recursively through references, but not through raw pointers.
+So, for example, if data pointed to by a `&T` or `& &mut T` is mutated, that's interior mutability.
```
rkruppe: "Data pointed to" is a bit ambiguous when talking about pointers-to-pointers, maybe talk more specifically about the data contained in the `T`?

reference/src/glossary.md
```diff
@@ -0,0 +1,55 @@
...
+```rust
+fn main() { unsafe {
+  let t: T = std::mem::transmute(TERM);
+} }
+```
+
+The *safety* invariant is an invariant that safe code may assume all data to uphold.
+This invariant is used to justify which operations safe code can perform.
+The safety invariant can be temporarily violated by unsafe code, but must always be upheld when interfacing with unknown safe code.
+It is not relevant when arguing whether some *program* has UB, but it is relevant when arguing whether some code safely encapsulates its unsafety -- IOW, it is relevant when arguning whether some *library* can be used by safe code to *cause* UB.
```
rkruppe:
```suggestion
It is not relevant when arguing whether some *program* has UB, but it is relevant when arguing whether some code safely encapsulates its unsafety -- IOW, it is relevant when arguing whether some *library* can be used by safe code to *cause* UB.
```

reference/src/glossary.md
RalfJung: TBH I used these terms specifically to avoid having to give a precise definition for things we haven't decided yet.^^

I will add a remark along those lines.

rkruppe: I didn't mean spelling out when a shared reference counts as being relevant for interior mutability, I just meant that the term "live" here suggests a concept of "liveness" as in program analysis -- a reference is "still around" or "may be used later" or something like that -- that some readers may not know.

reference/src/glossary.md
RalfJung: Is "immediately pointed to" better? I think we should have some terminology for this that's purely operational and does not refer to types.

reference/src/glossary.md
RalfJung: Oh I see. I'll add a remark along those lines.

reference/src/glossary.md
RalfJung: Done.

PR #57, opened 2018-12-19
add a chapter on enum representation

### LINKED ISSUE #10: Representation of enums
Discussion topic about how enums are represented in Rust. Some things to work out:

- What are the `#[repr]` options available for enums?
    - [RFC 2195][] defined the layout of `#[repr(C)]` enums with payloads.
    - [RFC 2363][] offers a proposal to permit specifying discriminations
- When do we **currently** perform `Option<T>`-like layout optimizations?
- When do we **guarantee** `Option<T>`-like layout optimizations?
    - For any Option-like enum? What about things like `Result<T, ()>`?
    - What are the conditions on T? (Obviously, must have some undefined values)
    - the [Rustonomicon](https://doc.rust-lang.org/nightly/nomicon/other-reprs.html) has some text on this topic
- Are there important enum variant optimizations we want freedom to do in the future that we might want to keep in mind?
- Size of empty enums and `!`: defined to be 0
- C-like enums: define, what does it say about representation?
    - document the effect of `#[repr(C)` and friends here

[RFC 2363]: https://github.com/rust-lang/rfcs/pull/2363
[RFC 2195]: https://rust-lang.github.io/rfcs/2195-really-tagged-unions.html

nikomatsakis:
This is a proposed summary chapter on enum representation. I made a best effort to represent what *seemed* to be the consensus from #10 but I would definitely appreciate review, as I confess I kind of wrote this in a hurry and could easily have missed some details or objections. 😛 

Here are the important details:

- Uninhabited enums (no variants) are guaranteed to share layout with `!`
- C-like enums are
- I copied the definition of the layout of `#[repr]` enums from [RFC #2195](https://rust-lang.github.io/rfcs/2195-really-tagged-unions.html)
- I coined the term "discriminant elision" to describe the "Option optimization"
- I adopted what seemed like a fairly restrictive definition of "option-like enums"; these are the enums for which we guarantee "discriminant elision". 
    - It *does* permit enums with more than one unit variant.
    - However, unit variants must have no fields, like `None`, and hence this definition excludes `Result<T, ()>`
    - I was thinking this would be better for now, so as to sidestep the problem of [defining zero-sized structs and so forth](https://github.com/rust-rfcs/unsafe-code-guidelines/issues/37), but I would be happy for us to revisit it.
- I left a few (non-normative, caveated) notes about niche values.

Here are a few questions:

- Is my definition of "option-like enums" acceptable? I think it is crucial to include `Option<&T>` as having a guaranteed optimization, but am somewhat flexible beyond that
- We never settled on a term for "niche values". I simply used "niche values" here but I was thinking after writing that perhaps "illegal" or "undefined values" would be better?
- In writing this summary, I realize I did not specify anything about variants with uninhabited types and so forth. Perhaps I should leave this as "future work"?

Fixes #10 

cc @eddyb in particular for technical review


reference/src/representation/enums.md
```diff
@@ -0,0 +1,376 @@
...
+**Disclaimer:** Some parts of this section were decided in RFCs, but
+others represent the consensus from issue [#10]. The text will attempt
+to clarify which parts are "guaranteed" (owing to the RFC decision)
+and which parts are still in a "preliminary" state.
+
+[#10]: https://github.com/rust-rfcs/unsafe-code-guidelines/issues/10
+
+## Background
+
+**C-like enums.** The simplest form of enum is simply a list of
```
rkruppe: These are called field-less enums nowadays (see e.g. [the reference](https://doc.rust-lang.org/reference/items/enumerations.html#custom-discriminant-values-for-field-less-enumerations)).

reference/src/representation/enums.md
```diff
@@ -0,0 +1,376 @@
...
+We cover each of the categories below. The layout rules for enums with
+explicit `#[repr]` annotations are specified in [RFC 2195][].
+
+[RFC 2195]: https://rust-lang.github.io/rfcs/2195-really-tagged-unions.html
+
+### Layout of an enum with no variants
+
+An enum with no variants can never be instantiated and is logically
+equivalent to the "never type" `!`. Such enums are guaranteed to have
+the same layout as `!` (zero size and alignment 1).
```
rkruppe: I was surprised to find out that `#[repr(align(N))] enum Foo { ... }` doesn't work. Might be worth mentioning, otherwise future readers might (like me) ask themselves whether they can't set alignment higher.

reference/src/representation/enums.md
```diff
@@ -0,0 +1,376 @@
...
+
+(Meta-note: The content in this section is not described by any RFC
+and is therefore "non-normative".)
+
+### Discriminant elision on Option-like enums
+
+**Definition.** An **option-like enum** is an enum which has:
+
+- one variant with a single field,
+- other variants with no fields ("unit" variants).
```
rkruppe: This definition is surprising to me. I don't recall any demand for that particular generalization and off-hand I can't think of a case where it would be really useful to have it guaranteed, and it requires a bunch of subtlety below that wouldn't otherwise be there.

ubsan: Yeah, why does this allow multiple other variants with no fields?

reference/src/representation/enums.md
```diff
@@ -0,0 +1,376 @@
...
+guaranteed to supply niche values corresponding to 256 - N (presuming
+that is a positive number). This is because a C-like enum must be
+represented using an integer and that integer must correspond to a
+valid variant: the precise size of C-like enums is not specified but
+it must be at least one byte, which means that there are at least 256
+possible bitstrings (only N of which are valid).
+
+Other enums -- or enums with a specified representation -- may supply
+niches if their representation invariant permits it, but that is not
+**guaranteed**.
```
asajeffrey: It seems odd to have a partial defn of "niche" here.

asajeffrey:
Comments added, mostly nits. The big one is that it would be nice to have the defn of "niche value" in this section rather than hidden away in another "representation invariant" section. This would make the "representation" section self-contained, and get rid of the need for "bitstring validity".

asajeffrey:
Comments added, mostly nits. The big one is that it would be nice to have the defn of "niche value" in this section rather than hidden away in another "representation invariant" section. This would make the "representation" section self-contained, and get rid of the need for "bitstring validity".

reference/src/representation/enums.md
nikomatsakis: It doesn't seem to require any subtlety in particular? But I'm fine to limit to one unit variant.

reference/src/representation/enums.md
```diff
@@ -0,0 +1,376 @@
...
+
+We cover each of the categories below. The layout rules for enums with
+explicit `#[repr]` annotations are specified in [RFC 2195][].
+
+[RFC 2195]: https://rust-lang.github.io/rfcs/2195-really-tagged-unions.html
+
+### Layout of an enum with no variants
+
+An enum with no variants can never be instantiated and is logically
+equivalent to the "never type" `!`. Such enums are guaranteed to have
```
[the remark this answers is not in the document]
nikomatsakis: Good question =) I don't know, have to investigate

reference/src/representation/enums.md
```diff
@@ -0,0 +1,376 @@
...
+#[repr(C)]
+struct TwoCasesVariantA(TwoCasesTag, u8, u16);
+
+#[repr(C)]
+struct TwoCasesVariantB(TwoCasesTag, u16);
+```
+
+Note that the `TwoCasesVariantA` and `TwoCasesVariantB` structs are
+`#[repr(C)]`; this is needed to ensure that the `TwoCasesTag` value
+appears at offset 0 in both cases, so that we can read it to determine
```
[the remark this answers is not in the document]
nikomatsakis: It's quoted directly from the RFC; it's also stronger than one would think is necessary. You only care about the first field being at offset 0, really.

reference/src/representation/enums.md
rkruppe: Perhaps not terribly subtle in the grand scheme of things, but the vague talk about "an adequate number of niche values" and the third example could be cut, and more drastically it would become feasible to not introduce the "niche" terminology at all at this stage (just enumerate the types we special-case in the compiler today).

reference/src/representation/enums.md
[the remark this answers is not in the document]
nikomatsakis: You cannot apply such annotations to empty enums (and I documented that)

reference/src/representation/enums.md
```diff
@@ -0,0 +1,376 @@
...
+where `union` represents a C union of all the possible variants. The
+type of the discriminant will be the integral type specified (`u8`,
+etc) -- if no type is specified, then the compiler will select one
+based on what a size a C-like enum would have with the same number of
+variants.
+
+This layout, while more compatible and arguably more obvious, is also
+less efficient than the non-C compatible layout in some cases in terms
+of total size.
+
```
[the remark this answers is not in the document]
nikomatsakis: added example

reference/src/representation/enums.md
```diff
@@ -0,0 +1,376 @@
...
+nullable `&u8` reference -- the `None` variant is then represented
+using the niche value zero. This is because a valid `&u8` value can
+never be zero, so if we see a zero value, we know that this must be
+`None` variant.
+
+In order for the optimization to apply, the payload type must define a
+number of niches greater than or equal to the number of unit variants.
+In the case of `Option<T>`, this means that any niche at all will
+suffice, as there is only one unit variant (`None`).
+
```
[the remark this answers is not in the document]
nikomatsakis: there is no stable way yet, but the compiler supports multiple niche values, and e.g. rustc takes advantage of this (all of our index types define values `0xFFFFFFF0` and above as niches, for example, so that they can be readily combined with enums)

reference/src/representation/enums.md
nikomatsakis: I removed that section regardless. but I'm not really sure what you meant by "partial" -- i.e., just covering enums? I was assuming each type would define for itself when it offers a niche. But we'll cover it later I guess.

nikomatsakis:
OK, pushed a few updates to resolve comments so far.

reference/src/representation/enums.md
```diff
@@ -0,0 +1,342 @@
...
+**Fieldless enums.** The simplest form of enum is one where none of
+the variants have any fields:
+
+```rust
+enum SomeEnum {
+  Variant1,
+  Variant2,
+  Variant3,
+```
```
asajeffrey: Missing `}`.

reference/src/representation/enums.md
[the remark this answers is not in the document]
asajeffrey: Which indexing types do you mean? `Option<usize>` takes 128 bits according to the playground.

reference/src/representation/enums.md
```diff
@@ -0,0 +1,342 @@
...
+  Variant44 = 44,
+  Variant45,
+}
+```
+
+As in C, discriminant values that are not specified are defined as
+either 0 (for the first variant) or as one more than the prior
+variant.
+
+**Data-carrying enums.** Enums whose enums have fields are called
```
RalfJung:
```suggestion
**Data-carrying enums.** Enums whose variants have fields are called
```

reference/src/representation/enums.md
```diff
@@ -0,0 +1,342 @@
...
+`!`. `#[repr]` annotations are not accepted on empty enums. Empty
+enums are guaranteed to have the same layout as `!` (zero size and
+alignment 1).
+
+### Layout of a fieldless enum
+
+If there is no `#[repr]` attached to a fieldless enum, it is
+guaranteed to be represented as an integer of sufficient size to store
+the discriminants for all possible variants. The size is selected by
+the compiler but must be at least a `u8`.
```
RalfJung: This is not correct, since enums with a single field are a ZST (and they are fieldless according to your definition).

reference/src/representation/enums.md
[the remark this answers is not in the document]
RalfJung: Types like `NodeId`, `CrateId` etc. are defined using [this macro](https://github.com/rust-lang/rust/blob/4755e2f3b635ad7391ba88b18672b05cbae5ba4c/src/librustc_data_structures/indexed_vec.rs#L67) which uses the `rustc_layout_scalar_valid_range_end` attribute to create a niche.

reference/src/representation/enums.md
```diff
@@ -0,0 +1,342 @@
...
+    A(u8, u16),
+    B(u16),
+}
+```
+
+This will be laid out equivalently to the following more 
+complex Rust types:
+
+```
+union TwoCasesRepr {
```
RalfJung: No `repr(C)` here? AFAIK we don't otherwise guarantee the offset to be 0.

reference/src/representation/enums.md
```diff
@@ -0,0 +1,342 @@
...
+field which it contains; in the case of `Option<T>`, the payload has
+type `T`.
+
+**Definition.** In some cases, the payload type may contain illegal
+values, which are called **niches**. For example, a value of type `&T`
+may never be NULL, and hence defines a niche consisting of the
+bitstring `0`.  Similarly, the standard library types [`NonZeroU8`]
+and friends may never be zero, and hence also define the value of `0`
+as a niche. (Types that define niche values will say so as part of the
+description of their representation invariant, which -- as of this
```
RalfJung: "representation invariant"? I don't think I've seen that term before.

Are you suggestion we rename "validity invariant" to "representation invariant"? (Not that I am necessarily opposed, just wondering where this term is coming from.)

reference/src/representation/enums.md
nikomatsakis: You're right, I meant to account for that but forgot, thanks

reference/src/representation/enums.md
```diff
@@ -0,0 +1,342 @@
...
+  Variant22 = 22,
+  
+  // Default discriminant is one more than the previous,
+  // so 23 will be assigned.
+  Variant23
+}
+```
+
+**Unresolved question:** What about platforms where `-fshort-enums`
+are the default? Do we know/care about that?
```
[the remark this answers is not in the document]
nikomatsakis: updated to match what you suggested here, I believe

reference/src/representation/enums.md
nikomatsakis: That text was copied from the RFC. I suppose it may be incorrect. Or maybe I missed a line.

reference/src/representation/enums.md
nikomatsakis: I meant validity invariant

nikomatsakis:
Addressed outstanding feedback I believe. =)

nikomatsakis:
I'm marking this with FINAL COMMENT PERIOD, it's been idle for quite a while.

reference/src/representation/enums.md
```diff
@@ -0,0 +1,348 @@
...
+enums are guaranteed to have the same layout as `!` (zero size and
+alignment 1).
+
+### Layout of a fieldless enum
+
+If there is no `#[repr]` attached to a fieldless enum, the compiler
+will represent it using an integer of sufficient size to store the
+discriminants for all possible variants -- note that if there is only
+one variant, then 0 bits are required, so it is possible that the enum
+may have zero size. In the absence of a `#[repr]` annotation, the
```
gnzlbg: I think we should open an issue about whether we want to guarantee that fieldless enums with one variant have zero size or not (and reference that issue here).

reference/src/representation/enums.md
```diff
@@ -0,0 +1,348 @@
...
+
+If no explicit `#[repr]` attribute is used, then the layout of a
+data-carrying enum is typically **not specified**. However, in certain
+select cases, there are **guaranteed layout optimizations** that may
+apply, as described below.
+
+#### Discriminant elision on Option-like enums
+
+(Meta-note: The content in this section is not described by any RFC
+and is therefore "non-normative".)
```
gnzlbg: We should open an issue about this. We might want to fill in an RFC to guarantee this as part of the UCG (or related work).

reference/src/representation/enums.md
```diff
@@ -0,0 +1,348 @@
...
+```
+
+As in C, discriminant values that are not specified are defined as
+either 0 (for the first variant) or as one more than the prior
+variant.
+
+**Data-carrying enums.** Enums whose variants have fields are called
+"data-carrying" enums. Note that for the purposes of this definition,
+it is not relevant whether those fields are zero-sized. Therefore this
+enum is considered "data-carrying":
```
gnzlbg: We probably also want to open an issue about data-carrying enums with zero-sized data, e.g., in case we want to guarantee that their layout is equivalent to that of a fieldless enum.

gnzlbg:
Still LGTM. 

My only nits are that Rust code in the wild already relies on many enum optimizations that are not guaranteed, and that makes the unsafe code guidelines less useful and more complicated than what they could be.

For example, the answer to the question: "Can I use `Option<&T>` on C FFI where pointers are expected?" is not "Yes.", but rather, "That relies on an optimization that is not guaranteed to happen, but in practice, yes, you can rely on that.". The same applies to data-carrying enums with ZST variants, fieldless enums with one variant, etc. 

We don't have to provide better answers to these questions right now, but we should strive to do so in 2019, e.g., via an RFC about guaranteed enum optimizations or something like that.

RalfJung:
> That relies on an optimization that is not guaranteed to happen

Doesn't this guarantee the optimization?

> **Option-like enums where the payload defines at least one niche value
are guaranteed to be represented using the same memory layout as their
payload.**

gnzlbg:
> Doesn't this guarantee the optimization?

If it were normative, it would, but a couple of lines above the document states that everything that follows isn't normative :/ 

We discussed this briefly in the meeting, and @nikomatsakis wanted to open an issue to track all non-normative things that we'd like to RFC at some point and why.

RalfJung:
> but a couple of lines above the document states that everything that follows isn't normative :/

I agree it should be made normative through an RFC, but you made it sound like a change in this document could make you happy ("my only nits are ...").  I don't understand how that can be the case.  Nothing we write here can change the fact that this is not normative.

gnzlbg:
Ah no, sorry! I didn't meant to say that, only that we should fill some issues about this and once we do that hyperlink them here, but that can be done as a follow up PR, once this is merged and those issues are filled.

PR #58, opened 2018-12-20
define representation, layout

RalfJung:
This describes how we currently use these terms.

However, it seems wasteful to use two words for the same thing.  And @ubsan has some other ideas for what "representation" can mean: When relating mathematical values to objects in memory, a "representation" defines how mathematical objects map to bitstrings.


asajeffrey:
Oh, I expected that to suggest a change, not add a commit to the PR.

RalfJung:
@asajeffrey that's fine though, collaborative editing :)

PR #59, opened 2018-12-21
extend def.n of interior mutability

RalfJung:
We usually also disallow mutating through pointers obtained from shared references. That's a separate condition from there being live shared references.

See https://internals.rust-lang.org/t/writing-through-a-pointer-derived-from-a-shared-reference-after-that-reference-is-dead/9078 for how this might matter.

### DIFF
--- a/reference/src/glossary.md
+++ b/reference/src/glossary.md
@@ -2,13 +2,14 @@
 
 #### Interior mutability
 
-*Interior Mutability* means mutating memory where there also exists a live shared reference immediately (i.e., non-transitively) pointing to the same memory.
-This propagates recursively through references, but not through raw pointers.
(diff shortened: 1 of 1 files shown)


reference/src/glossary.md
```diff
@@ -2,14 +2,15 @@
 
 #### Interior mutability
 
-*Interior Mutability* means mutating memory where there also exists a live shared reference immediately (i.e., non-transitively) pointing to the same memory.
-This propagates recursively through references, but not through raw pointers.
+*Interior Mutability* means mutating memory where there also exists a live shared reference pointing to the same memory; or mutating memory through a pointer derived from a shared reference.
```
nikomatsakis: grammatical nit: I believe this is a definition of "interior **mutation**". I guess interior **mutability** is the ability to do interior mutation without it being UB.

RalfJung: Better like this?

PR #61, opened 2018-12-26
add one of eddyb's optimizations

RalfJung:
Cc @eddyb

### DIFF
--- a/reference/src/optimizations/return_value_optimization.md
+++ b/reference/src/optimizations/return_value_optimization.md
@@ -1,0 +1,20 @@
+We should turn
+
+```rust
+// y unused
+let mut x = f();
+g(&mut x);
+y = x;
+// x unused
+```
+
+into
+
+```rust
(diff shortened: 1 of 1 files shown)


reference/src/optimizations/return_value_optimization.md
```diff
@@ -0,0 +1,20 @@
...
+into
+
+```rust
+y = f();
+g(&mut y);
+```
+
+to avoid a copy.
+
+The potential issue here is `g` storing the pointer it got as an argument elsewhere.
```
gnzlbg: If `x` is unused after the call to `g()`, can that pointer be used for anything ?

RalfJung: `g` could have stored it in a global variable and enable other code to use it, for example.

nikomatsakis:
We decided to merge this in the meeting today.

PR #63, opened 2018-12-30
C11 fixes the valid reprs of _Bool

gnzlbg:
* [C11 6.2.5p2](http://port70.net/~nsz/c/c11/n1570.html#6.2.5p2): "An object declared as type `_Bool` is large enough to store the values `0` and `1`".

* [C11 6.2.6.2p5](http://port70.net/~nsz/c/c11/n1570.html#6.2.6.2p5): "For any integer type, the object representation where all the bits are zero shall be a representation of the value zero in that type.". This fixes the `false` bit-pattern to `0x0`.

* [C11 6.3.1.2p1](http://port70.net/~nsz/c/c11/n1570.html#6.3.1.2p1): "When any scalar value is converted to `_Bool`, the result is `0` if the value compares equal to `0`; otherwise, the result is `1`". Converting any C integer to `_Bool` creates an integer that contains either `0` or `1` - this fixes the `false` bit-pattern to `0x0` and the `true` bit-pattern to `0x1`.

* [C11 7.18p3](http://port70.net/~nsz/c/c11/n1570.html#7.18p3): "`true` which expands to the integer constant `1`, `false` which expands to the integer constant `0`" - this is not so relevant since these expand to "integer constant"s (and not `_Bool`s), so an integer to `_Bool` conversion needs to kick in to initialize a `_Bool`.

* [C11 Footnote 122](http://port70.net/~nsz/c/c11/n1570.html#note122): "While the number of bits in a `_Bool` object is at least `CHAR_BIT`, the width (number of sign and value bits) of a `_Bool` may be just `1` bit".

---

All in all, a `_Bool` can can contain two valid values, 0 and 1, all other representations appear to be trap representations but I cannot find any clear statement in the standard stating precisely this.

--

cc @mjbshaw

### DIFF
--- a/reference/src/representation/integers-floatingpoint.md
+++ b/reference/src/representation/integers-floatingpoint.md
@@ -49,9 +49,9 @@ The C integer types specify a minimum size, but not the exact size. For this rea
 ## Controversies
 There has been some debate about what to pick as the "official" behavior for bool:
 * Rust does what C does (this is what the lang team decided)
-    * and in all cases you care about, that is 1 byte that is 0 or 1
+    * and in all cases you care about, that is 1 byte
 or
-* Rust makes it 1 byte with values 0 or 1
+* Rust makes it 1 byte
     * and in all cases you care about, this is what C does
     
 Related discussions: [document the size of bool](https://github.com/rust-lang/rust/pull/46156), [bool== _Bool?](https://github.com/rust-rfcs/unsafe-code-guidelines/issues/53#issuecomment-447050232), [bool ABI](https://github.com/rust-lang/rust/pull/46176#issuecomment-359593446)


gnzlbg:
> Is _Bool an "integer type" in this sense?

Yes, `_Bool` is an unsigned integer type, [C11 6.2.5p6](http://port70.net/~nsz/c/c11/n1570.html#6.2.5p6): "The type `_Bool` and the unsigned integer types that correspond to the standard signed integer types are the standard unsigned integer types".

---

> This would be a quite important thing to settle though. If there are other ways to construct a _Bool object that don't perform a "conversion from a scalar type" (e.g., using memset) then it may be misleading to say true is 0x1 -- maybe on some implementation it's just the conventional/canonical representation but any non-zero value is considered valid and equivalent to true.

Yes, `true` is an integer constant expression, so that technically allows the compiler to transform a comparison of `_Bool` with `true` into pretty much anything. For example, `_Bool == true` could be transformed into `_Bool != false` (`!= 0`), which means that a `_Bool` containing any value that isn't `0` is actually `true`. This leaves me wondering about how many trap representations `_Bool` has, if any. Also, the "padding bits" parts of the integer representations leaves me wondering whether the canonical representation for `true` could be something different than `0x1` on some platforms.

gnzlbg:
I am not sure what the implications of that are, particularly for the validity of `bool`. Since `bool == _Bool`,`bool` is C FFI safe, and it is ok to construct a `_Bool` by transmuting a `3_u8`, then that kind of implies that transmuting a `3_u8` into a `bool` has to create a valid `bool`.

mjbshaw:
> it is ok to construct a `_Bool` by transmuting a `3_u8`

It's implementation-defined whether or not that's okay in C. It might be UB. The only thing that's really guaranteed in C is that all-bits-zero == 0. There can be other representations that also equal 0 (or 1) due to the padding bits.

gnzlbg:
> It's implementation-defined whether or not that's okay in C. 

The issue is that **iff** the C implementation defines that it is ok to do that for `_Bool` (like GCC does), then Rust needs to define that it is ok to do so for `bool` as well.

mjbshaw:
> The issue is that **iff** the C implementation defines that it is ok to do that for `_Bool` (like GCC does)

GCC does not say it's okay to store `0x03` in a `_Bool`. It's UB for GCC (try compiling with `-fsanitize=undefined`). It's UB for all implementations Rust cares about, as they all require `_Bool` to be the bit patterns 0x00 or 0x01, and nothing else is permitted.

<sub>(apologies for bringing up a bit of a red herring in my comment above).</sub>

gnzlbg:
@mjbshaw Do you have a source for that ? Note that just because `-fsanitize=bool` complains about this does not mean that it is undefined behavior (e.g. `-fsanitize` can also complain about unsigned integer overflow, which is defined behavior).

The source of `-fsanitize=bool` in compiler-rt (https://github.com/llvm-mirror/compiler-rt/blob/master/lib/ubsan/ubsan_handlers.cc#L429) does not have many comments. AFAICT the `-fsanitize=bool` check is correct for C++'s `bool`, which can only contain `true` and `false` values like the check tests, but C's `_Bool` does not have that restriction because of padding bits. AFAIK in LLVM both `bool` and `_Bool` are always `trunc`ated to an `i1`, discarding all padding bits, so their value does not really matter.

mjbshaw:
@gnzlbg The simplest source/proof I can offer is that clang/gcc/icc/etc. will optimize based on the assumption that `_Bool` is 0x00 or 0x01 (and nothing else). For example, [compile the simple program](https://gcc.godbolt.org/z/IqjQ--):

```c
unsigned int foo(unsigned int x, _Bool b) {
    return x - (unsigned int)b;
}
```

If `b` has any value other than 0x00 or 0x01, the results will be incorrect.

gnzlbg:
Then that has to mean that all other representations beyond `0x0` and `0x1` are trap representations on these implementations.

gnzlbg:
Closing because while the representation of `false` (`0x0`) is guaranteed by the C standard, the representation of `true` is not. When specifying the validity of `bool` we can mention the `false` / `0x0` part.

gnzlbg:
FWIW, the important bit for `bool`'s validity is that the actual number of valid representations of `bool` is implementation-defined, and not just `true` and `false` (it's only `true` and `false` on all platforms that Rust currently supports, it would be great if those platforms would document that).

RalfJung:
> Then that has to mean that all other representations beyond 0x0 and 0x1 are trap representations on these implementations.

Note that this means they have to be trap representations for all compilers on those platforms, or else linking object files created by different compilers could cause unexpected UB.

PR #67, opened 2019-01-04
Use layout consistently instead of mixing it with representation

### LINKED ISSUE #65: Use layout consistently in the UCG documents
See https://github.com/rust-rfcs/unsafe-code-guidelines/pull/58#issue-240240656 . 

In the last meeting we discussed that we should use the term "layout" consistently and that we should not use the term "representation" to refer to "layout". This reserves the term "representation" for other purposes, like relating how mathematical objects are mapped to bitstring (cc @ubsan ). 

The name of the `repr` attribute is unfortunate, but we can just document it as an attribute that controls "layout" (omitting the term "representation").

gnzlbg:
Closes #65 .


reference/src/glossary.md
```diff
@@ -49,16 +49,18 @@ Moreover, such unsafe code must not return a non-UTF-8 string to the "outside" o
...
 For some more information, see [this blog post](https://www.ralfj.de/blog/2018/08/22/two-kinds-of-invariants.html).
 
-#### Layout / Representation
+#### Layout
 
 The *layout* of a type defines its size and alignment as well as the offsets of its subobjects (e.g. fields of structs/unions/enum/... or elements of arrays).
 Moreover, the layout of a type records its *function call ABI* (or just *ABI* for short): how the type is passed *by value* across a function boundary.
 
-We often use *representation* as a synonym for "layout".
+note: Originally, *layout* and *representation* were treated as synonyms, and Rust language features like the `#[repr]` attribute reflect this. In this
```
```suggestion
Note: Originally, *layout* and *representation* were treated as synonyms, and Rust language features like the `#[repr]` attribute reflect this. In this
```

Also, I've been using one-line-per-sentence above, please stick to that -- or use 80 columns or so, but this newline here seems somewhat odd.^^

RalfJung:
Yeah, this seems reasonable.

PR #83, opened 2019-02-01
Stacked Borrows: add code for the two core operations

RalfJung:
Namely, accessing memory and dereferencing a pointer.

As requested by @ubsan

### DIFF
--- a/wip/stacked-borrows.md
+++ b/wip/stacked-borrows.md
@@ -106,7 +106,7 @@ pub enum RetagKind {
 
 * A `Default` retag happens on the return value of every function that gets called (i.e., this is the first statement in the basic block that the call will return to).
 
-* The automatically generated drop shims perform a `Raw` retag of their argument because they use it as a raw pointer.
+* The automatically generated drop shims (generated as the body of `ptr::real_drop_in_place`) perform a `Raw` retag of their argument because they use it as a raw pointer.
 
 ## Operational semantics
 
--- a/wip/stacked-borrows.md
+++ b/wip/stacked-borrows.md
@@ -2,7 +2,9 @@
 
 **Note:** This document is not normative nor endorsed by the UCG WG.  It is maintained by @RalfJung to reflect what is currently implemented in [Miri].
 
-For more background on the how and why of Stacked Borrows, see this [blog post on the first implementation][stacked-borrows-1] and this [follow-up][stacked-borrows-2].
+This is not a guide!
+It is more of a reference.
+For a more guide-like introduction of Stacked Borrows, see this [blog post on the first implementation][stacked-borrows-1] and this [follow-up][stacked-borrows-2].
 Also see some [notes from a discussion at the 2019 Rust All-Hands][all-hands].
 
 [Miri]: https://github.com/solson/miri/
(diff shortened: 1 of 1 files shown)


gnzlbg:
You might want to include in a different PR a glossary at the beginning of this file with a brief description of what each technical term introduced here means. I think that would improve the document.

RalfJung:
Could you give a list of terms that you think need explanation? That would help a lot.

gnzlbg:
Sure, I was thinking, something like this

* `Timestamp`: a monotonically increasing counter. 

* `place`:  a memory location (maybe link to the general glossary here).

* `Borrow`: a per-pointer tag `tag` indicating whether the memory referred by the point is uniquely referred by this pointer or aliased by other pointers.

* `CallId`: Per-call frame timestamp.

*  `FnBarrier`: <it is unclear to me what this does, looking at the document, it is unclear to me when these are pushed or why, only when they are popped>

* `Stack`: a per-place `stack` of borrows where the borrow with the largest `Timestamp` is at the top. Each borrow is either unique, was created from a raw pointer, or <barrier> whatever that does. The `Stack` might be frozen since a particular `Timestamp`.

etc.

I think it might be a good idea to give a brief definition of each of the most commonly use terms at the beginning. So that when the reader goes through the text it already has an intuition for what they are. Particularly since some types are used before they are introduced (e.g. `Raw`, `CallId`, etc.).

I had some problems reading the text sequentially, finding myself trying to jump back and forth a couple of times trying to find an explanation for what some types are supposed to represent.

RalfJung:
Thanks!

Notice that this document is not intended as a guide. I expect that people first read the blog posts, and then come here as a reference for the details on one operation.  But I agree adding a brief description (doc-comment style) to each type makes sense.

RalfJung:
I have added doc-comments to the types and enum variants. I hope that helps.

wip/stacked-borrows.md
```diff
@@ -15,31 +19,50 @@ Every pointer value has a *tag* (in addition to the location in memory that the
...
     borrows: Vec<BorStackItem>, // used as a stack; never empty
     frozen_since: Option<Timestamp>, // virtual frozen "item" on top of the stack
 }
 
-// Extra per-call-frame state
+/// Extra per-call-frame state: an ID uniquely identifying a stack frame.
 type CallId = nat;
 
-// Extra global state
+/// Extra global state: the next unused `Timestamp`, as well as the next unused `CallId`.
```
gnzlbg: What does "unused" mean here?

wip/stacked-borrows.md
```diff
@@ -120,14 +143,15 @@ impl Tracking {
...
+In the following, we pretend there exists a function `barrier_is_active(id)` that can check this.
 
 **Note**: Miri uses a slightly more complex system with a `HashSet<CallId>` tracking the set of active `CallId`; that is just an optimization to avoid having to scan the call stack all the time.
 
 ### Allocating memory
 
 When allocating memory, we have to initialize the `Stack` associated with the new locations, and we have to choose a `Borrow` (a tag) for the initial pointer to this memory.
 
-For most memory, the stack of each freshly allocated memory location is `Stack { borrows: vec![Shr], frozen_since: None }`, and the initial pointer to that memory has tag `Shr(None)`.
+For most memory, the stack of each freshly allocated memory location is `Stack { borrows: vec![Raw], frozen_since: None }`, and the initial pointer to that memory has tag `Alias(None)`.
```
gnzlbg: Do we also carry the alignment of that memory ? The section below on dereferencing raw pointers states that `size_of_val` is checked, but it says nothing about checking that the pointer is at an alignment boundary of `T`.

gnzlbg: Or is that implied as part of "the pointer must be dereferenceable" ?

wip/stacked-borrows.md
```diff
@@ -176,27 +325,52 @@ we determine the extent of memory that this place covers using `size_of_val` and
    Remember the position of the item matching the tag in the stack.
```
gnzlbg: typo: "we determine the extent of memory" should start with a capital letter: "We determine ..."

wip/stacked-borrows.md
RalfJung: "not used"? I don't get the question.

wip/stacked-borrows.md
RalfJung: Stacked Borrows does not care about alignment at all, and that's why it is not mentioned.

wip/stacked-borrows.md
RalfJung: Thanks! Fixed.

wip/stacked-borrows.md
gnzlbg: Miri complains both on the creation of an unaligned reference and on the pointer dereference below, so I wonder how / where this is tracked ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=4db5abaec7e73cacba66f65c683e9cac)):

```
use core::{arch::x86_64::__m128, mem::transmute};
#[derive(Copy, Clone)]
#[repr(align(8))]
struct U8(u8);
fn main() {
    unsafe {
        type U = U8;
        let a: [U8; 16] = [U8(0); 16];
        let b: *const __m128 = &a as &[U] as *const [U] as *const U as *const __m128;
        let _c: &__m128 = transmute(b);
        let _d = *b;
    }
}
```

RalfJung: The former is the validity invariant, and the latter is the normal rule for a memory access.

This question would be answered by a proper definition of MIR semantics (and, in particular, places), which I will write... eventually.^^  But it is not part of Stacked Borrows and IMO should not be answered here.

gnzlbg: > But it is not part of Stacked Borrows and IMO should not be answered here.

Fair enough, and thanks!

PR #92, opened 2019-02-22
move function-pointers to layout/

RalfJung:


RalfJung:
Is there any chance you could merge things in a way that actually preserves my commits?  I proposed 3e0c6d48e81ceb25efc37089ad3782f70b430e72 to be merged, but the commit that got actually merged is 19f422384c5f8316235ac48917a59a6becbda99f. This means git doesn't actually know my branch got merged, which makes it much harder for me to track which branches I can delete.

gnzlbg:
I used the repo's default which appears to be "Rebase and Merge". Maybe we should change them to something else?

RalfJung:
I'd be strongly in favor of that. (For me it offers "Merge" per default, but maybe I changed it once and it remembers that?)

gnzlbg:
So I think you are right. In the repo settings one can only configure what's allowed, but for each user it might just remember what they used last time (hopefully on a per repo basis). I tried looking whether I could change my defaults for this repo but that does not seem possible. Sorry about that.

PR #94, opened 2019-03-06
Layout of arrays

### LINKED ISSUE #91: Layout of arrays
We are missing the layout of arrays (I think we just forgot about this). Tangentially related to the layout of homogeneous structs: https://github.com/rust-rfcs/unsafe-code-guidelines/issues/36

gnzlbg:
[Rendered](https://github.com/gnzlbg/unsafe-code-guidelines/blob/repr_array/reference/src/layout/arrays.md).

Closes #91 .


reference/src/layout/arrays.md
```diff
@@ -0,0 +1,33 @@
...
+```rust,ignore
+struct A(u16, u8); // size_of == 3, align_of == 4
+type B = [A; 4];  // => stride == 4 > 3
+```
+
+The size and alignment of `Vector` element types match, such that `Vector` types
+and arrays are layout compatible.
+
+`repr(C)` arrays have the same layout as C arrays and are passed by pointer in C
+FFI according to the C ABI.
```
RalfJung: This is just a consequence of the prior statements, not a new rule, correct? That should be made more explicit.

reference/src/layout/arrays.md
```diff
@@ -0,0 +1,33 @@
...
+
+Having a stride larger than the element size allows:
+
+```rust,ignore
+struct A(u16, u8); // size_of == 3, align_of == 4
+type B = [A; 4];  // => stride == 4 > 3
+```
+
+The size and alignment of `Vector` element types match, such that `Vector` types
+and arrays are layout compatible.
```
RalfJung: Do you mean `Vec`? But the element type of a `Vec<T>` is `T`, there's no slice here.

I am not sure if `Vec` is worth mentioning here; they must be layout compatible because `Vec` allows slice indexing -- but that's a concern for whoever implements such a library.

reference/src/layout/arrays.md
```diff
@@ -0,0 +1,33 @@
...
+
+Array types, `[T; N]`, store `N` values of type `T` contiguously with a constant
+_stride_ (the distance between each two consecutive values).
+
+The offset of the first array element is `0`.
+
+The stride of the array is computed as the size of the element type rounded up
+to the next multiple of the alignment of the element type. That is, the _stride_
+of an array can be larger than the element size iff the alignment requirements
+of the element are larger than its size.
```
RalfJung: I think it is worth mentioning that in current Rust, we always have `size >= align` and hence we always have `stride == size` -- but we might want to keep the door open to change that in the future.

reference/src/layout/arrays.md
```diff
@@ -0,0 +1,33 @@
...
+
+```rust,ignore
+struct A(u16, u8); // size_of == 3, align_of == 4
+type B = [A; 4];  // => stride == 4 > 3
+```
+
+The size and alignment of `Vector` element types match, such that `Vector` types
+and arrays are layout compatible.
+
+`repr(C)` arrays have the same layout as C arrays and are passed by pointer in C
```
rkruppe: There's no such thing as a "`repr(C)` array". Do you mean an array those *element type* is compatible with some C type?

reference/src/layout/arrays.md
```diff
@@ -0,0 +1,33 @@
...
+
+The stride of the array is computed as the size of the element type rounded up
+to the next multiple of the alignment of the element type. That is, the _stride_
+of an array can be larger than the element size iff the alignment requirements
+of the element are larger than its size.
+
+Having a stride larger than the element size allows:
+
+```rust,ignore
+struct A(u16, u8); // size_of == 3, align_of == 4
```
rkruppe: This is wrong, `size_of` is 4 here. In some sense the size *could be* 3, but it isn't. If we ever introduce a stride concept so that we can make that disctinction, backwards compatibility might force us to create a new term for the quantity that is 3 then (code might manually compute the size of an array as `size_of::<ElemTy>() * LENGTH`).

reference/src/layout/arrays.md
gnzlbg: I think I meant just "arrays", as in "arrays are passed by pointer in C FFI" (nothing else).  cc @RalfJung 

What I was referring to is that in: 

```rust
extern "C" { fn foo(x: [i32; 2]); }
unsafe { foo([0; 2]) }; 
```

the array is not memcpy'ed into `foo`, but passed by pointer. 

But maybe it is not worth to mention this here; that's ABI / calling convention and not layout.

reference/src/layout/arrays.md
gnzlbg: I meant `Vector` (as in `layout/vector.md` - the other file modified in this PR).

reference/src/layout/arrays.md
gnzlbg: Let's continue this discussion below (i've cc'ed you).

reference/src/layout/arrays.md
rkruppe: No that's not true. The array is passed as an aggregate, which depending on ABI details might end up on the stack, but it usually won't be ABI compatible with a C function that looks like it takes an array argument. [See LLVM IR here, for example.](https://play.rust-lang.org/?version=stable&mode=release&edition=2018&gist=8d07d1734f3b2ff396dbe859aca5887c)

reference/src/layout/arrays.md
RalfJung: The intro to this code says "Having a stride larger than the element size allows" -- so indeed size is not 3 today, but it could be.

reference/src/layout/arrays.md
RalfJung: Yeah... I don't think fixed-size arrays are FFI-safe. Does the lint not fire for them?

gnzlbg: Indeed. Arrays are Aggregates at least in the SysV ABI, so that makes sense.

reference/src/layout/arrays.md
RalfJung: Oh. The code quotes threw me off -- it's not like we have a `Vector` syntax for that in Rust.

reference/src/layout/arrays.md
gnzlbg: > This is wrong, size_of is 4 here.

Do we guarantee that in the layout of structs ?

reference/src/layout/arrays.md
gnzlbg: Please check if the last commit addresses this.

reference/src/layout/arrays.md
gnzlbg: I guess I can hyperlink them to make that more clear.

gnzlbg: Done.

RalfJung: Please also remove the `quotes` and mention "SIMD" -- I suppose for most Rust programmers, the first association with "vector" is `Vec`, not SIMD types.

reference/src/layout/arrays.md
gnzlbg: > Does the lint not fire for them?

No, it does not. I've removed this.

The only open question is whether this should mention that arrays are not FFI safe.

reference/src/layout/arrays.md
rkruppe: We don't guarantee the size of this struct, but the fact is, `size_of::<A>()` is not 3 on any existing Rust implementation. Saying it *is* 3 is simply wrong. If you really want to make the point that its size *could* be 3, then say *that*. As is, this section is plainly misleading.

RalfJung:
I think this can be ordered better, to be more understandable:

* First the definition of stride.
* Then a section with some consequences of that definition: When `size` is a multiple of `align`, then `stride == size`. This is mandated for the element types of a SIMD vector. In fact this holds true for all current Rust types.
* Then the "open question" section.

Giving the *hypothetical* example where `A` has size 3 so early is really confusing.

reference/src/layout/arrays.md
gnzlbg: Done, I hope it is clearer now. Let me know if you have a better idea about how to write this.

rkruppe:
I don't think this chapter should at all try to accomodate hypothetical future language extensions that allow array stride to differ from the size in other contexts. Besides all the reasons that it's unlikely to ever happen (backcompat hazards, massive churn, questionable gains [1]), writing for this hypothetical future complicates the wording of the layout documentation and it seems difficult to adequately describe that future without being misleading about or distracting from the status quo.

It's also far from obvious to me what terminology would be used if it ever comes to that, so however we described it now might turn out to be "wrong" in the future. For example, it seems equally plausible that the current `size_of` would remain unchanged and so correspond to that's called *stride* here, and a new term would be introduced for "size without trailing padding".

I am also concerned that if this text distinguishes `size_of` specifically (as opposed to another concept of "size") from the stride, then unsafe code authors will be in a pickle: how are they to reason about the size of an array? To be compatible with the claimed future extension, they'd need the *stride*, but there's no way to access that today, they have only `size_of`.

[1]: the gains are questionable because small types benefit the most from nixing padding, but those same types also benefit from being padded to word alignment so they can be accessed with fewer loads/stores.

reference/src/layout/arrays.md
rkruppe: Well, Rust arrays are definitely layout compatible with C arrays, and if they're contained in a struct the struct will also be ABI compatible. The root of the issue is that a C signature like `void foo(int x[4]);` does not actually mean that an array is passed, it decays to a pointer. This is definitely worth noting (and IMO the ctypes lint should fire for array arguments) but it's actually "just" a subtlety of C surface syntax, not an actual difference in ABI rules.

reference/src/layout/arrays.md
```diff
@@ -0,0 +1,61 @@
...
+== size`, and the elements are laid out contiguously in memory, e.g., `[u8; 4]`.
+In this case, the size of the array can be computed as `size_of::<T>() * N`[^1].
+
+> **Note:** In the current Rust implementation, _size_ is always a multiple of
+> the element's _alignment_, and therefore `stride == size` always holds. This
+> is, however, not guaranteed by the [layout of structs and tuples].
+
+[^1]: The alignment of the array is, however, unspecified. For example, the
+[SysV AMD64 ABI] requires array arguments to be at least 16 byte aligned to
+allow the use of SSE instructions.
```
rkruppe: I don't think this is quite right for this context. I believe you are referring to

> An array uses the same alignment as its elements, except that a local or globalarray variable of length at least 16 bytes or a C99 variable-length array variablealways has alignment of at least 16 bytes.

That is quite different from the ABI alignment that `std::mem::align_of[_val]` and all these layout documents describe, which applies regardless of storage location and is also a restriction on unsafe code that allocates memory/constructs objects. But since globals and locals are entirely allocated by the compiler, overaligning them need not impact the alignment.

reference/src/layout/arrays.md
gnzlbg: I did not say that it is 3, I said that for the rules to allow it to be 3 we would need to support a `stride > size`. Anyways, the way I wrote that was super-confusing, I hope it is clearer now.

gnzlbg:
> they'd need the stride, but there's no way to access that today, 

Can't one construct a two element array and measure the pointer offset in bytes? That would be a super annoying thing to do, but is there a case where this is not possible?

reference/src/layout/arrays.md
gnzlbg: I see what you mean. So what do we know about the array alignment ? Is it the same as the alignment of its element? Or can it be larger ?

For example, supposing I wanted to interface with C via an array of `repr(C)` types

```rust
extern {
    fn foo(x: *mut [ReprC; 8]) -> *mut [ReprC; 8];
}
```

is it enough if C returns me an array where the alignment of the array matches the alignment of `ReprC`, or can Rust decide that this array has a larger alignment requirement making this illegal ?

gnzlbg: If Rust arrays of ReprC types are layout compatible with C arrays, then the alignment requirements of these arrays would need to match those of C. But I'm not certain that this is (or should be) the case of arrays of `repr(Rust)` types.

rkruppe:
Hm, you're right, that should work (even in constant expressions) eventually. But I think it'll be so annoying and niche that nobody will bother and so the backwards compatibility problem of introducing stride != size would increase further.

gnzlbg:
> Hm, you're right, that should work (even in constant expressions) eventually. But I think it'll be so annoying and niche that nobody will bother and so the backwards compatibility problem of introducing stride != size would increase further.

Sure, like, if somebody wanted to implement or propose this in a more proper form, I'll expect them to add an `std::mem::stride::<[T; N]>()` intrinsic or similar.

I think here we want to document what we currently guarantee without forbidding future work on that area, but at the same time documenting how things currently work.

reference/src/layout/arrays.md
rkruppe: Rust unilaterally imposing larger ABI alignment requirements on arrays than the corresponding C array type would break FFI as you say, so clearly we can't do that. For Rust types that aren't suitable for FFI (e.g., unspecified layout), there's no such external constraint, but I wasn't talking about those anyway. I'm just saying that the SysV aspect you refer to here has no relation as all to the `align_of` of (C-compatible) arrays. As such, it's potentially interesting trivia but not relevant to the layout document.

gnzlbg: So I've removed that part, and I've written down the guarantees about array alignment down.

rkruppe:
I don't have a problem with leaving open the possibility of such future changes in some way, I think the current write-up is acceptable if we really want to refuse closing any doors. However, at the same time, I'm sad to put this extra mental burden on unsafe code authors when any such proposal will most likely either die from multiple contradictory backwards compatibility constraint or have to come up with a whole new and weird taxonomy of layout anyway.

reference/src/layout/arrays.md
gnzlbg: So I worded the issues with C FFI in the last commit. I hope this is now ok.

reference/src/layout/arrays.md
```diff
@@ -0,0 +1,76 @@
...
+
+However, a future Rust implementation could implement a layout optimization that
+reduces the size of `A` to `3`. For the elements of `B` to be properly aligned,
+`B` would need to choose a `stride == 4`, resulting in a `stride > size`.
+
+Guaranteeing `stride >= size` is forward-compatible with such
+layout-optimization proposals:
+  
+  * [rust-lang/rfcs/1397: Spearate size and stride for types](https://github.com/rust-lang/rfcs/issues/1397)
+  * [rust-lang/rust/17027: Collapse trailing padding](https://github.com/rust-lang/rust/issues/17027)
```
RalfJung: Shouldn't slices be mentioned somewhere?

reference/src/layout/arrays.md
```diff
@@ -0,0 +1,76 @@
...
+> **Note:** In the current Rust implementation, _size_ is always a multiple of
+> the element's _alignment_, and therefore `stride == size` always holds. This
+> is, however, not guaranteed by the [layout of structs and tuples].
+
+[^1]: When `stride > size` the pointer needs to be advanced by the array
+    _stride_ instead of by the element _size_.
+
+[layout of structs and tuples]: ./structs-and-tuples.md
+
+The [layout of Vector types][Vector] [^2] requires the _size_ and _alignment_ of
```
RalfJung: I'd still prefer if we spoke about "SIMD vectors". But it seems I am alone with that position?

rkruppe: No, I agree that the extra clarity is worth it here. The whole text should be consistent about this, though.

reference/src/layout/arrays.md
```diff
@@ -0,0 +1,76 @@
...
+> foreign function declarations, e.g., `extern { fn foo(x: [T; N]) -> [U; M];
+> }`. Pointers to arrays are fine: `extern { fn foo(x: *const [T; N]) -> *const
+> [U; M]; }`.
+
+The _stride_ of the array is the distance between a pair of consecutive values
+within the array, and it is constant for all element pairs. It is computed as
+the _size_ of the element type rounded up to the next multiple of the
+_alignment_ of the element type.
+
+When the element _size_ is a multiple of the element's _alignment_, then `stride
```
RalfJung: What about putting this and the next paragraph into a subsection like "Special case: stride == size"?

reference/src/layout/arrays.md
gnzlbg: I don't know. What should we say about them?

reference/src/layout/arrays.md
gnzlbg: I can rename the `Vector` layout to `Packed SIMD Vector` layout to make this clearer (and differentiate these from Cray Vectors).

reference/src/layout/arrays.md
gnzlbg: Sounds good !

reference/src/layout/arrays.md
RalfJung: Well it seems to me that the layout of arrays and the layout of a slice are closely related -- they must have the same stride etc. Otherwise unsizing would be wrong.

reference/src/layout/arrays.md
RalfJung: "cray vectors"? Now I am even more confused.^^ I also don't feel "packed" adds anything here because anyway there'd be no padding between the vector elements. So for an outsider it just seems redundant. Seems like you are trying to make a distinction with some obscure other kind of vectors that is long obsolete?

Remember you are not writing this text for people that are experts in vector types like you are. I wrote some manually vectorized code once and even I get confused by your terminology.

reference/src/layout/arrays.md
gnzlbg: Does that belong here or should there be a "Layout of slices" document ?

I can rename this document to "Arrays and Slices", and add a sub-section about the layout of slices, where the text could say:

"The layout of a slice `[T]` with `N` elements is the same as that of a `[T; N]` array. "

But I am not sure if that's enough.

RalfJung: It seems like a good start, at the least.

gnzlbg: So I've added a sub-section for it.

reference/src/layout/arrays.md
gnzlbg: I hope I clarified all uses of "vector" with "packed SIMD vector" in the last commit. There wasn't really a need to mention other vector types anywhere AFAICT. Better to worry about that when / if we add new vector types.

RalfJung:
Looks good, thanks!

reference/src/layout/arrays-and-slices.md
```diff
@@ -0,0 +1,86 @@
...
+element type. If the element type is `repr(C)` the layout of the array is
+guaranteed to be the same as the layout of a C array with the same element type.
+
+> **Note**: the type of array arguments in C function signatures, e.g., `void
+> foo(T x[N])`, decays to a pointer. That is, these functions do not take arrays
+> as an arguments, they take a pointer to the first element of the array
+> instead. Array types are therefore _improper C types_ (not C FFI safe) in Rust
+> foreign function declarations, e.g., `extern { fn foo(x: [T; N]) -> [U; M];
+> }`. Pointers to arrays are fine: `extern { fn foo(x: *const [T; N]) -> *const
+> [U; M]; }`.
```
rkruppe: Maybe worth noting that structures and unions containing arrays are also fine.

reference/src/layout/arrays-and-slices.md
```diff
@@ -0,0 +1,86 @@
...
+
+The _stride_ of the array is the distance between a pair of consecutive values
```
rkruppe: This repeats the first sentence, which reads a bit strange to me since they're not that far apart. I'm not sure if omitting this and just stating how the stride is computed here is any better?

gnzlbg: I've changed it a bit. Is it better now?

gnzlbg:
@ubsan read this as well and found it ready. We can always amend this in subsequent PRs.

PR #98, opened 2019-03-14
Be more explicit about the layout guarantees of integer and floating-point types

gnzlbg:
I was going through issue #9 and there it was mentioned that we wanted to guarantee that the fixed-width integer (and floating-point) types can interoperate with C's fixed-width integer and floating point types. Going through the text this felt implied, in particular, because C's fixed-width integer types guarantee some things that were not mentioned in the text (e.g. 2's complement for signed integers).

This PR attempts to call this out more explicitly. In particular in #9 there was a question about whether one should use, e.g., `libc::uint8_t` in C FFI or whether using `u8` is ok. This PR explicitly mentions that using `u8` is ok.


reference/src/layout/integers-floatingpoint.md
```diff
@@ -37,6 +37,25 @@ Note: on all platforms that Rust's currently supports, the size and alignment of
...
 
+## Fixed-width integer types
+
+Rust's fixed-width integer types `{i,u}{8,16,32,64}` have the same layout as the
+C fixed-width integer types from the `<stdint.h>` header
+`{u,}int{8,16,32,64}_t`. That is:
+
+* these types have no padding bits,
+* their size exactly matches their bit-width,
+* negative values of signed integer types are represented using `2`'s complement.
```
RalfJung: Nit: `2` here should not be in code font, it's not quoting code.

reference/src/layout/integers-floatingpoint.md
```diff
@@ -37,6 +37,25 @@ Note: on all platforms that Rust's currently supports, the size and alignment of
 
 For full ABI compatibility details, see [Gankro’s post](https://gankro.github.io/blah/rust-layouts-and-abis/#the-layoutsabis-of-builtins).
 
+## Fixed-width integer types
+
+Rust's fixed-width integer types `{i,u}{8,16,32,64}` have the same layout as the
+C fixed-width integer types from the `<stdint.h>` header
+`{u,}int{8,16,32,64}_t`. That is:
```
RalfJung: This section overlaps in scope with "Relationship to C integer hierarchy". Shouldn't the two be merged?

reference/src/layout/integers-floatingpoint.md
```diff
@@ -37,6 +37,25 @@ Note: on all platforms that Rust's currently supports, the size and alignment of
...
+* negative values of signed integer types are represented using `2`'s complement.
+
+Therefore these integer types are safe to use directly in C FFI where the
+fixed-width integer types are expected.
+
+## Fixed-width floating point types
+
+Rust's `f32` and `f64` types have the same layout as C's `float` and `double`
+types, respectively. Therefore these floating-point types are safe to use
+directly in C FFI where the appropriate C types are expected.
```
rkruppe: I don't think we can state that in such generality, at least not unilaterally. I believe C allows implementing `float` and `double` as the same format, so that `double` *also* corresponds to `f32` (this is potentially useful if you target a platform with f32 hardware but no f64 support). Annex F would not apply to such an implementation, but as long as the implementation doesn't define `__STDC_IEC_559__` as 1 that's conforming.

reference/src/layout/integers-floatingpoint.md
gnzlbg: That section states some of the properties of C native integer types (e.g. `int`) and states that Rust types are not guaranteed to be layout compatible with those.

I could move it to a note / sub-section of this one (the fixed-integer one). I think it should not only mention what C guarantees about its native integer types (e.g. `int`), but also what it does not guarantee (e.g. padding bits, representation of negative numbers, etc.), to explain why Rust fixed-width integer types aren't guaranteed to always match to any of those. 

It should mention that in practice one can, however, use "implementation-defined" information to properly interface with these integer types by using the appropriate Rust types (and this is in fact what `libc` does, so we want to explicitly mention that this is ok). 

We should also note that in practice, all platforms that Rust currently support have integers with 2s complement representation, without padding bits, etc.

reference/src/layout/integers-floatingpoint.md
gnzlbg: @rkruppe AFAICT, if `__STDC_IEC_559__` is not defined, there are no guarantees whatsoever about what `float` and `double` could be (except that `double` is at least as large as `float`).

Maybe I should rewrite this to say that `f32` and `f64` only matches `float` and `double` if `<math.h>` defines `__STDC_IEC_559__`, and that otherwise they don't necessarily match. 

Do all platforms that Rust currently support define `__STDC_IEC_559__`? We could add some info about this as a note.

rkruppe: I am a bit hesitant to recommand tying outselves to `__STDC_IEC_559__`, because although there is a long line of official sentiments that Rust's floating point types are always IEEE 754, it is definitely possible for a C implementation to provide Rust-compatible floating point types without implementing the rest of Annex F (e.g., floating point environment access). Indeed Clang is currently in that situation (though in practice the `math.h` you get defines `__STDC_IEC_559__` and so it's technically not conforming).

reference/src/layout/scalars.md
```diff
@@ -0,0 +1,116 @@
...
+  _implementation-defined_, but can at most be `usize::max_value()` since
+  `ptr.add(count: usize)`.
+
+> **Note**: in the current Rust implementation:
+> 
+> * the maximum size of Rust values is limited to `isize::max_value()`. The LLVM
+> `getelementptr` instruction uses signed-integer field offsets. Rust calls
+> `getelementptr` with the `inbounds` flag which assumes that field offsets do
+> not overflow,
+> * the maximum number of elements in an array is `usize::max_value()`,
```
RalfJung: It might be worth mentioning that only ZST arrays can actually be that large; non-ZST arrays would otherwise violate the first clause.

reference/src/layout/scalars.md
```diff
@@ -0,0 +1,116 @@
...
+They have the same layout as the [pointer types] for which the pointee is
+`Sized`, and are layout compatible with C's `uintptr_t` and `intptr_t` types.
+
+> **Note**: Rust's `usize` and C's `unsigned` types are **not** equivalent. C's
+> `unsigned` is at least as large as a short, allowed to have padding bits, etc.
+> but it is not necessarily pointer-sized.
+
+The layout of `usize` determine the following:
+
+- the maximum size of Rust values is _implementation-defined_, but can at most
```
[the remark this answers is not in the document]
rkruppe: A relatively easy way to avoid that issue while also being less jargon-intense would be to talk about the sizes of allocations. Just risks people thinking about *heap* allocation to the exclusion of all other kinds.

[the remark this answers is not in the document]
gnzlbg: @RalfJung once #40 is merged we should probably start linking these terms to their definitions. 

@rkruppe with #40 we might be talking about the size of "places", which would remove the ambiguity with heap allocations, but it is still jargon heavy.

PR #100, opened 2019-03-14
Note the layout of pointers to T:Sized matches that of C pointers

### LINKED ISSUE #99: Are raw pointers to sized types usable in C FFI ?
AFAICT we don't mention this explicitly anywhere.

https://github.com/rust-lang/unsafe-code-guidelines/blob/master/reference/src/layout/pointers.md#representation mentions that they are one word in size, but maybe we should spell out that they are layout compatible with C pointers (including alignment) to make it clear that's ok to use raw pointers in C FFI.

gnzlbg:
Closes #99 .

### DIFF
--- a/reference/src/layout/pointers.md
+++ b/reference/src/layout/pointers.md
@@ -32,8 +32,12 @@ and are at least one word.
 The layouts of `&T`, `&mut T` and `*T` are the same.
 
 If `T` is sized, references and pointers to `T` have a size and alignment of one
-word and have therefore the same layout as C pointers - they can be used in C FFI
-where pointers are expected.
+word and have therefore the same layout as C pointers.
+
+> **warning**: while the layout of references and pointers is compatible with
+> the layout of C pointers, references come with a _validity_ invariant that
+> does not allow them to be used when they could be `NULL`, unaligned, dangling,
+> or, in the case of `&mut T`, aliasing.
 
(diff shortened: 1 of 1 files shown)


reference/src/layout/pointers.md
```diff
@@ -31,6 +31,10 @@ and are at least one word.
 
 The layouts of `&T`, `&mut T` and `*T` are the same.
 
+If `T` is sized, references and pointers to `T` have a size and alignment of one
+word and have therefore the same layout as C pointers - they can be used in C FFI
+where pointers are expected.
```
RalfJung: It might be worth mentioning that references come with a validity invariant that makes them *not* usable in FFI contexts whenever there might be NULL, unaligned or dangling pointers.

gnzlbg: I think i'll just leave the last part about usable in C FFI out. Stating that they have the same layout as C pointers is enough.

RalfJung: Still feels like a warning would be prudent. IIRC we plan to eventually merge the validity and layout sections for each of these types, then such a warning would be even more natural.

gnzlbg: Done.

RalfJung:
Nobody complained for 4 weeks, merging :)

PR #104, opened 2019-03-19
Publish the book

gnzlbg:


RalfJung:
On which URL should this appear, when it's working?

gnzlbg:
Hopefully on `https://rust-lang.github.io/unsafe-code-guidelines`, but once its working we could tune that.

PR #110, opened 2019-04-11
Add rationale for bool layout

gnzlbg:
This PR adds the rationale for `bool` layout guarantees. 

cc @nikomatsakis - this is how the "Rationale" details could look like.

### DIFF
--- a/reference/src/layout/scalars.md
+++ b/reference/src/layout/scalars.md
@@ -19,6 +19,15 @@ taking on the values 1 (`true`) or 0 (`false`).
 
 [Rust Layout and ABIs]: https://gankro.github.io/blah/rust-layouts-and-abis/#the-layoutsabis-of-builtins
 
+<details><summary><b>Rationale</b></summary> This allows directly using `bool` in the
+  FFI to interface with C's `_Bool` without having to use a `c_bool` type alias.
+  This might require Rust programs to define and use a `struct Bool(u8);` on
+  platforms where `_Bool` is larger than 1 byte to save space. [T-Lang
+  decided](https://github.com/rust-lang/rust/pull/46176#issuecomment-359593446)
+  that simplifying C FFI on all platforms is a better trade-off than
+  having small `bool`s on uncommon platforms.
+</details>
+
 ## `char`
 
 Rust char is 32-bit wide and represents an [unicode scalar value]. The alignment


RalfJung:
:+1: for the general format.

gnzlbg:
FY this is how it'd look like:

**Collapsed:**

collapsed

---

**Expanded:**

expanded

cc @Centril - I don't know if something like this is what T-lang had in mind about adding inline rationale, but this would be one way to do it.

reference/src/layout/scalars.md
```diff
@@ -19,6 +19,15 @@ taking on the values 1 (`true`) or 0 (`false`).
 
 [Rust Layout and ABIs]: https://gankro.github.io/blah/rust-layouts-and-abis/#the-layoutsabis-of-builtins
 
+<details><summary><b>Rationale</b></summary> This allows directly using `bool` in the
```
Centril: I would suggest moving "This..." to a new after `</summary>` which makes things clearer when reading the raw text.

reference/src/layout/scalars.md
```diff
@@ -19,6 +19,15 @@ taking on the values 1 (`true`) or 0 (`false`).
...
+  FFI to interface with C's `_Bool` without having to use a `c_bool` type alias.
+  This might require Rust programs to define and use a `struct Bool(u8);` on
+  platforms where `_Bool` is larger than 1 byte to save space. [T-Lang
+  decided](https://github.com/rust-lang/rust/pull/46176#issuecomment-359593446)
```
Centril: "View file" produces weird results for this link.

Centril:
> cc @Centril - I don't know if something like this is what T-lang had in mind about adding inline rationale, but this would be one way to do it.

This inline rationale looks good for `bool`.

For newer things that haven't been decided, it seems fine to use this form of inline rationale as well.
However, you may want to include a *summary* of the rationale and link to a larger design document with an expanded rationale, drawbacks, alternatives, and so on.

gnzlbg:
> However, you may want to include a summary of the rationale and link to a larger design document with an expanded rationale, drawbacks, alternatives, and so on.

When we need that we could add an Appendix about rationale in the book, and if that gets out of control, we could even split that into a second document with rationale (e.g. a second book in this repo). Ideally, when RFCs, FCPs, summary comments, etc. are written about decisions that would affect the UCGs, those documents will be written, and updated (e.g. the RFC being updated after the tracking issue is closed), such that they can just be linked from here. 

I don't think it is reasonable to expect that we will ever have the throughput to write RFCs with extended rationale and alternatives for everything in the language that interacts with unsafe code, mir model, memory model, etc. in the reasonable future, when we barely have throughput to merely document how integers work.

gnzlbg:
I agree that the inline rationale should be brief and to the point, and link to sources that expand more on the decision, and I think that github issues / PRs with 100s of comments do a poor job at that, but summary comments, RFC documents, etc. are lately getting much better and could do this job, and when this isn't the case, finding out and summarizing the rationale is a lot of work, that if somebody puts in, we can include in its own separate document here (although it might be better in the reference, as an update to the original RFC, etc.).

gnzlbg:
So I'm closing this. It appears that for now, this style (modulo nitpicks) for inline rationale is better than nothing. If anyone wants to take a slab at starting to add inline rationale for the different sections, go for it. If things start getting out of hand, we can explore some of the other approaches here like an Appendix for rationale or a parallel book.

RalfJung:
> It appears that for now, this style (modulo nitpicks) for inline rationale is better than nothing.

But then why would you close this instead of merging it?

gnzlbg:
The text itself was just a strawman intended to see how this would render.

I don't know how we would go about merging any rationale. For issues like this one, which have been FCPed by T-Lang, but for which there isn't really a comprehensive summary comment anywhere, and the discussion involves dozens of comments across many different channels, T-Lang should at least FCP/review the rationale - otherwise it feels a bit like putting words in T-Lang's mouth.

RalfJung:
> For issues like this one, which have been FCPed by T-Lang, but for which there isn't really a comprehensive summary comment anywhere, and the discussion involves dozens of comments across many different channels, T-Lang should at least FCP/review the rationale - otherwise it feels a bit like putting words in T-Lang's mouth.

I don't see that. Everything in this repo is the UCG speaking and needs an RFC to be normative. We already established that for the actual text. Why should it be any different for the rationale?

gnzlbg:
The RFC for the UCGs Cover both new stuff with new rationale, and old Stuff
with old rationale that has already been RFCed.

 For this older stuff that has already been RFCed, it would be helpful if
T-Lang, which should be able to recall the rationale, would already give us
a thumbs up on it, so that during the RFC of the UCGs we can focus on the
new stuff.

PR #111, opened 2019-04-14
The Basic Embedded Abstract Machine (BEAM) and `unsafe` embedded code

japaric:
Hello UCG WG. I'm @japaric from the [Embedded WG](https://github.com/rust-embedded/wg).

In embedded Rust, and specially in Rust code that runs on (single-core)
microcontrollers, we use `unsafe` code in (rather) unusual ways (spoilers: we
use `static mut` variables, a lot), which we believe to be safe and sound. I'd
like to discuss the soundness of these patterns (i.e. whether the compiler /
LLVM will mis-optimize them or not) with the UCG WG so I'll create (roughly) one
discussion PR per pattern and list them at the end of this comment. But before
you click on those PRs please read the rest of this comment.

Concurrency on microcontrollers looks quite different from the OS-thread
concurrency model one uses in non-`no_std` programs and even among
microcontrollers details around how one does concurrency vary a bit between
architectures. So to make sure we are all on the same page when discussing
`unsafe` embedded Rust code I'm specifying in this PR an *abstract
machine* that's representative of the microcontrollers one can run Rust
code on. I'll call this abstract machine: the Basic Embedded Abstract Machine
(BEAM).

[Rendered specification](https://github.com/japaric/unsafe-code-guidelines/blob/beam/wip/beam-spec.md)

Assuming that I do a good job, this specification should be all you need to know
about microcontrollers to discuss the soundness of the `unsafe` patterns listed
below. All the PRs have code examples that target the BEAM.

- #112 Runtime initialization of static variables
- #113 Unsynchronized access to a shared `static mut` variable
- #114 Critical sections
- #115 `&'static mut` references to static variables and LLVM `noalias`
- #116 The meaning of Send and Sync in the context of interrupt handlers

### DIFF
--- a/wip/beam-spec.md
+++ b/wip/beam-spec.md
@@ -1,0 +1,437 @@
+# The Basic Embedded Abstract Machine (BEAM)
+
+## Summary
+
+This document contains the specification of an *abstract machine*. The goal of
+this specification is facilitating discussions about the soundness of `unsafe`
+concurrent code that runs on microcontrollers.
+
+The BEAM is meant to be representative of the microcontrollers one can run Rust
+code on, as such it has concurrency capabilities commonly found in this kind of
+hardware but exposes these capabilities in a way that's not tied to any
+particular architecture.
+
+## Program execution
+
+- The BEAM has a single core / processing-unit and can only run a single process
+  on this core.
+
+- When the BEAM executes a program it uses the function called `main` as the
+  entry point. Returning from this function is UNSPECIFIED behavior.
+
+- The signature of `main` must be `[unsafe] fn() [-> !]`. That is the program
+  takes no arguments.
+
+## Interrupt handling
+
+(Interrupt handling is a preemption mechanism that in its most basic form works
+as follows: when an interrupt *signal* arrives the processor suspends the
+execution of the current function and jumps to a special subroutine known as the
+interrupt *handler*. When this subroutine returns the processors resumes the
+execution of the function that was preempted by the interrupt handler.)
+
+- BEAM's interrupt handling capabilities can be summarized as: prioritization of
+  interrupts, individual interrupt masking and global interrupt masking.
+
+- The BEAM has 8 *interrupts* whose *handlers* are named: `INTERRUPT0`,
+  `INTERRUPT1`, all the way to `INTERRUPT7`.
+
+- The signature of all interrupt handlers must be `[unsafe] fn()`. That is these
+  handlers don't take arguments nor can they return values.
+
+- If not defined in the program, interrupt handlers default to:
+
+``` rust
+#[no_mangle]
+unsafe fn INTERRUPT_() {
+    loop {}
+}
+```
+
+- When the BEAM starts, all interrupts are *disabled*, *unmasked* and are
+  configured with an initial *priority* of `1`; and the *running priority* of
+  the system is set to `0`.
+
+- Interrupts can be assigned any *priority* in the range `1..=255`.
+
+- At any point in time an interrupt signal may arrive and change the state of an
+  interrupt to *pending*. That is interrupt signals occur asynchronously to the
+  execution of the program.
+
+- If interrupts are *enabled*, setting the state of an *unmasked* interrupt to
+  *pending* results in its handler preempting the current function *if and only
+  if* the *priority* of the interrupt is *higher* than the current *running
+  priority*, otherwise the interrupt will be kept in the *pending* state until
+  the *running priority* is low enough to allow preemption.
+
+- The *pending* state of an interrupt gets cleared when its handler starts
+  executing.
+
+- Once an interrupt handler starts executing the *running priority* is
+  automatically set to the *priority* of the interrupt being serviced. When the
+  interrupt handler returns the running priority is restored to its previous
+  (before preemption) value.
+
+- While interrupts are *disabled* preemption can not occur. Interrupts in the
+  pending state will remain in that state until interrupts are re-enabled.
+
+- Setting the state of a *masked* interrupt to *pending* will never result in
+  preemption, no matter the priority of the interrupt. Such interrupt will only
+  be serviced after it has been unmasked.
+
+- The behavior of setting the *running priority* within the context of an
+  interrupt handler to a value *lower* than the *priority* of that interrupt
+  results in UNSPECIFIED behavior.
+
+- Changing the *priority* of an interrupt from within its handler results in
+  UNSPECIFIED behavior.
+
+- Interrupt handling can be summarized as follows:
+
+``` rust
+// pseudo-code
+if interrupts_are_enabled {
+    if interrupt_N_is_pending {
+        if !interrupt_N_is_masked {
+            if interrupt_N_priority > running_priority {
+                atomic {
+                    interrupt_N_is_pending = false;
+                    running_priority = interrupt_N_priority;
+                    start_INTERRUPTN();
+                }
+            }
+        }
+    }
+}
+```
+
+## Instructions
+
+The BEAM has a few special instructions that can be invoked using inline
+assembly `asm!`.
+
+- `asm!("ENABLE_INTERRUPTS")`. This instruction *enables* all interrupts.
+
+- `asm!("DISABLE_INTERRUPTS")`. This instruction *disables* all interrupts.
+
+- `asm!("SET_RUNNING_PRIORITY $0")`. This instruction changes the *running
+  priority* to the value `$0`.
+
+## Registers
+
+The BEAM has a special region of memory; write operations performed on this
+memory region can affect how interrupt handling works. Each significant piece of
+memory in this memory region is referred to as a *register*.
+
+- All registers are byte-sized
+
+- A program can only access the memory mapped registers by name using (part of)
+  the following `extern` block.
+
+``` rust
+extern "C" {
+    // write-only register
+    static MASK_INTERRUPT: UnsafeCell<u8>;
+
+    // write-only register
+    static UNMASK_INTERRUPT: UnsafeCell<u8>;
+
+    // read-write register
+    static PEND_INTERRUPT: UnsafeCell<u8>;
+
+    // read-only register
+    static RUNNING_PRIORITY: UnsafeCell<u8>;
+
+    // read-write register
+    static INTERRUPT_PRIORITY: [UnsafeCell<u8>; 8];
+}
+```
+
+- Performing *non*-volatile operations on these registers results in UNSPECIFIED
+  behavior.
+
+- Performing memory operations on these registers other than byte-sized reads /
+  writes results in UNSPECIFIED behavior.
+
+- The `MASK_INTERRUPT` register. Writing a value of `1` to bit `N` causes
+  interrupt `N` to be *masked*. Writing a value of `0` to any bit has no effect.
+  Reading this register returns an UNSPECIFIED value.
+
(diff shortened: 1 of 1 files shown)


gnzlbg:
At this rate I think we are probably years away from being at a point where starting to discuss these issues would make sense.

I don’t think it makes sense do delay a layout / validity MVP with any of this.

comex:
Hmm, the name collides with the [BEAM virtual machine](https://en.wikipedia.org/wiki/BEAM_(Erlang_virtual_machine)) that powers Erlang.

RalfJung:
Thanks @japaric for posting this! When you asked about some discussion I expected a few issues with questions, not several books worth of content. oO

> I don’t think it makes sense to delay a layout / validity MVP with any of this.

I don't think @japaric proposed that. :)

But I agree that it seems hard to make hard statements about using Rust in a particular environment before we have set in stone anything about the semantics of the language.^^

However, just like I am experimenting with a Stacked Borrows spec here without that expressing even the consent of the UCG yet, and focusing only on a narrow aspect of Rust program execution while ignoring the (not-yet-specified) rest, it seems fine to me to do the same with BEAM. I am just afraid you overwhelmed us thoroughly, @japaric -- we barely have the resources to keep our discussions going and have a hard time finding people to do write-ups, so it will be hard to muster the resources to properly review 6 PRs.  Is there a way you could stretch this over time so that we don't feel like we are drowning in content?^^

wip/beam-spec.md
```diff
@@ -0,0 +1,437 @@
...
+- The signature of `main` must be `[unsafe] fn() [-> !]`. That is the program
+  takes no arguments.
+
+## Interrupt handling
+
+(Interrupt handling is a preemption mechanism that in its most basic form works
+as follows: when an interrupt *signal* arrives the processor suspends the
+execution of the current function and jumps to a special subroutine known as the
+interrupt *handler*. When this subroutine returns the processors resumes the
+execution of the function that was preempted by the interrupt handler.)
```
RalfJung: I think this should not be just a parenthetical. This is important background for people not familiar with the area.

wip/beam-spec.md
```diff
@@ -0,0 +1,437 @@
...
+
+- BEAM's interrupt handling capabilities can be summarized as: prioritization of
+  interrupts, individual interrupt masking and global interrupt masking.
```
RalfJung: Could you explain what these capabilities *mean*?

wip/beam-spec.md
```diff
@@ -0,0 +1,437 @@
...
+``` rust
+#[no_mangle]
+unsafe fn INTERRUPT_() {
+    loop {}
+}
+```
+
+- When the BEAM starts, all interrupts are *disabled*, *unmasked* and are
+  configured with an initial *priority* of `1`; and the *running priority* of
+  the system is set to `0`.
```
RalfJung: What is the difference between "disabled" and "masked"? I'd have thought these are the same thing...

It seems like you have some kind of per-interrupt state machine in mind here, with states like disabled, pending, masked, ...? If yes, then please write down that state machine explicitly.

wip/beam-spec.md
RalfJung: Oh, I now see you do that later. The way this gets mentioned in an enumeration here made me think you expect the reader to understand what all of this means -- you go on with the next enumeration item, so the previous one is done.

wip/beam-spec.md
```diff
@@ -0,0 +1,437 @@
...
+            if interrupt_N_priority > running_priority {
+                atomic {
+                    interrupt_N_is_pending = false;
+                    running_priority = interrupt_N_priority;
+                    start_INTERRUPTN();
+                }
+            }
+        }
+    }
+}
```
RalfJung: This would have been really useful further up when I tried to figure out what all that English text means.

wip/beam-spec.md
```diff
@@ -0,0 +1,437 @@
...
+  `i`-th register changes the static priority of interrupt `i` to `X`. Reading
+  the `i`-th register returns the static priority of interrupt `i`.
+
+## Examples
+
+This section contains examples (real Rust code) that target the BEAM. They are
+meant to improve your understanding of the above specification but they are not
+part of the specification itself.
+
+In all these examples the BEAM goes from state A, to B, to C, etc.
```
RalfJung: This sentence makes no sense unless one has already read an example. Maybe announce in advance there will be comments detailing the states the machine can be in?

RalfJung:
@japaric how do we proceed with this? We'd like to get our PR list cleaned up a bit here. ;)

My proposal is to merge this PR here, ideally after applying some of the feedback we gave above. The other PRs are, as you stated, more discussions than resolutions. so I think they should be turned into issues referenced the WIP document here. We can have a label collecting those issues. This is similar to what we do for Stacked Borrows, which also has [a WIP document](https://github.com/rust-lang/unsafe-code-guidelines/blob/master/wip/stacked-borrows.md) and [several open issues/discussions](https://github.com/rust-lang/unsafe-code-guidelines/issues?q=is%3Aissue+is%3Aopen+label%3Atopic-stacked-borrows).

wip/beam-spec.md
gnzlbg: This explains the "signal" and "interrupt handler" concepts. I think it might be worth it to also explain here "masking", since that term is used below a couple of times before being introduced.

wip/beam-spec.md
```diff
@@ -0,0 +1,437 @@
...
+                    interrupt_N_is_pending = false;
+                    running_priority = interrupt_N_priority;
+                    start_INTERRUPTN();
+                }
+            }
+        }
+    }
+}
+```
+
```
gnzlbg: From the text, it is obvious that preemption can occur in the middle of a function, but I was wondering, at which precise granularity can preemption occur ? Or maybe in different words, what are the "atomic" operations of the abstract machine, during which preemption does not occur ? 

For example. consider a single instruction like `add` in an architecture in which it returns a result in one register, and sets a flag in another register indicating whether the add overflowed. Now consider two arguments for which the `add` overflows. Could the execution of this instruction be preempted after setting the flag register but before writing the result value? That would allow an interrupt handler to remove the overflow flag, and would prevent the interrupted code from knowing that an overflow happened. 

This should not be able to happen, that is, such an `add` operation that modifies multiple registers would be non-preemptable in the middle of its execution. Or in other words, preemption can only happen across these non-preemptable operations.

I don't really know what's the best way to declare / specify at which granularity can pre-emption happen, but this should be specified somewhere.

gnzlbg: For example, in a different less-useful abstract machine, preemption could only be allowed across, for example, function calls (e.g. in BEAM, this would be like setting the priority to 255 when entering every function, and setting it to 0 right before calling a function, such that the only window in which this can happen, is right before entering the next function).

wip/beam-spec.md
```diff
@@ -0,0 +1,437 @@
...
+    // State C (RUNNING_PRIORITY = 2, PEND_INTERRUPT = 0b0000_0010)
+    //                                                          ^
+}
+
+// INTERRUPT_PRIORITY[1] = 1
+#[no_mangle]
+unsafe fn INTERRUPT1() {
+    // State D (RUNNING_PRIORITY = 1, PEND_INTERRUPT = 0b0000_0000)
+}
+```
```
gnzlbg: From reading the examples, it might be clearer to use `VolatileCell` instead of `UnsafeCell` for the registers.

wip/beam-spec.md
```diff
@@ -0,0 +1,437 @@
...
+execution of the function that was preempted by the interrupt handler.)
+
+- BEAM's interrupt handling capabilities can be summarized as: prioritization of
+  interrupts, individual interrupt masking and global interrupt masking.
+
+- The BEAM has 8 *interrupts* whose *handlers* are named: `INTERRUPT0`,
+  `INTERRUPT1`, all the way to `INTERRUPT7`.
+
+- The signature of all interrupt handlers must be `[unsafe] fn()`. That is these
+  handlers don't take arguments nor can they return values.
```
gnzlbg: To me this document feels like a step towards specifying what are interrupt handlers allowed / not allowed to do. 

I think that is worth specifying and we might be able to do that without having to speak about "masking" and many other concepts touched in this document. 

From that point-of-view, this single point is one of the most important points of the document to me. 

This point mentions that their type must be `fn() -> (), yet if they are not defined, they are implemented as `fn interrupt() { loop {} }`, so I wonder, can their type also be `fn() -> !` ? 

Also, if they return, they either have no effects (e.g. `fn interrupt() {}`) or they have a side-effect. Which side-effects are they allowed to have ? For example, suppose we have:

```rust
static PTR: UnsafeCell<*mut i32>;
unsafe fn interrupt() { PTR.get().write_volatile(42); }
fn foo() {
    let mut x = 0;
    PTR.set(&mut x);
    // preempted to `interrupt` here
    dbg!(x);  // Does this print `0` or `42` ?
}
```

`foo` does not contain any `unsafe { }` block, yet its execution can be preemtpted to an `unsafe fn` call, which can modify `foo` stack. Can `dbg!(x)` be optimized by the compiler to `dbg!(0)` ?

wip/beam-spec.md
petrochenkov: >consider a single instruction like add in an architecture in which it returns a result in one register, and sets a flag in another register indicating whether the add overflowed. Now consider two arguments for which the add overflows. Could the execution of this instruction be preempted after setting the flag register but before writing the result value?

I'm not sure about specifics of embedded hardware, but this seems to be pretty unlikely behavior.

CPU ISAs are normally specified in terms of instructions following each other in-order and changing observable architectural state like registers atomically (unless explicitly said otherwise).

Even if CPU works is out-of-order fashion internally, instructions as defined by ISA are still retired in-order and maintain architectural state at retirement.

That's why taking of interrupts/exceptions is also tied to instruction retirement, and the first instruction of exception handler will see consistent state left by the previous retired instruction (this is called "precise interrupts").

TLDR: the granularity is macro instructions as defined by respective CPU's ISA.

gnzlbg: @petrochenkov Yeah, I guess something like MIR, where `add_with_overflow` is an actual intrinsic, would be more interesting to consider than an actual ISA then. MIR intrinsics are going to get lowered into multiple CPU instruction, and preemption can happen across any of them.

RalfJung: This is getting into the interaction of hardware-level interrupts and language abstract machines... that's a *hard* problem.^^  This is why C says that in an interrupt, the content of all memory is unspecified except for volatile and atomic stuff. We likely need something similar.

wip/beam-spec.md
gnzlbg: In a sense, `interrupt`s always run in a different thread of execution, so even thought the application is single threaded, they should move into them any types that are not `Send`, should not read / write through references to types that are not `Sync`, etc. 

Ideally, we would have an `#[interrupt]` annotation in the language that would enforce these invariants at the type level and produce proper errors.

RalfJung: > In a sense, interrupts always run in a different thread of execution,

I agree that this is the best way to think about them.
In user space, the same happens with signal handlers.

RalfJung:
@japaric final reminder that we are going to close the BEAM PRs soon and replace them by an issue. I'd really like to land this general description of the BEAM machine, but we can't do it alone.

RalfJung:

PR #113, opened 2019-04-14
[BEAM] Unsynchronized access to a shared `static mut` variable

japaric:

[Rendered discussion document](https://github.com/japaric/unsafe-code-guidelines/blob/beam-static-unsynchronized/active_discussion/embedded/static-unsynchronized.md)

### DIFF
--- a/active_discussion/embedded/static-unsynchronized.md
+++ b/active_discussion/embedded/static-unsynchronized.md
@@ -1,0 +1,82 @@
+# Unsynchronized access to a shared `static mut` variable
+
+Sometimes interrupt handlers need to share state. This is done using a static
+variable. We claim that *no* synchronization is needed to access the shared
+static variable if the interrupt handlers run at the same priority.
+
+Note that:
+
+- The programs in this document target the [Basic Embedded Abstract Machine
+  (BEAM)][beam]. Please become familiar with the linked specification before you
+  read the rest of this document.
+
+[beam]: https://github.com/rust-lang/unsafe-code-guidelines/pull/111
+
+- In these programs we assume that [rust-lang/rfcs#2585][rfc2585] has been
+  accepted and implemented.
+
+[rfc2585]: https://github.com/rust-lang/rfcs/pull/2585
+
+## Example program
+
+Consider this program
+
+``` rust
+#![no_std]
+
+#[no_mangle]
+unsafe fn main() -> ! {
+    unsafe {
+        asm!("ENABLE_INTERRUPTS" : : : : "volatile");
+    }
+
+    loop {
+        // .. any safe code ..
+    }
+}
+
+static mut X: u128 = 0;
+
+#[no_mangle]
+unsafe fn INTERRUPT0() {
+    let x: &mut u128 = unsafe { &mut X };
+
+    // .. any safe code ..
+}
+
+#[no_mangle]
+unsafe fn INTERRUPT1() {
+    let x: &mut u128 = unsafe { &mut X };
+
+    // .. any safe code ..
+}
+```
+
+Note that "any safe code" can *not* call `main`, `INTERRUPT0` or `INTERRUPT1`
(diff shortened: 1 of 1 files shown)


active_discussion/embedded/static-unsynchronized.md
```diff
@@ -0,0 +1,82 @@
...
+    }
+}
+
+static mut X: u128 = 0;
+
+#[no_mangle]
+unsafe fn INTERRUPT0() {
+    let x: &mut u128 = unsafe { &mut X };
+
+    // .. any safe code ..
```
gnzlbg: Can interrupt0 be interrupted here again by interrupt1 ?

active_discussion/embedded/static-unsynchronized.md
```diff
@@ -0,0 +1,82 @@
...
+}
+```
+
+Note that "any safe code" can *not* call `main`, `INTERRUPT0` or `INTERRUPT1`
+(because they are `unsafe` functions), use `asm!` or access registers.
+
+**Claim**: this program is well-defined / sound. Rust aliasing rules are
+preserved because no preemption is possible between `INTERRUPT0` and
+`INTERRUPT1`.
```
gnzlbg: IDK exactly what you mean by "no preemption is possible between `INTERRUPT0` and
`INTERRUPT1`.", but from the BEAM document, if `INTERRUPT1` has higher priority than `INTERRUPT0`, it can preempt `INTERRUPT0`.

active_discussion/embedded/static-unsynchronized.md
japaric: Not because:

- they start with the same priority (`1`) when `main` starts executing
- their priority has not been changed in `main` (or elsewhere)
- changing the priority is an `unsafe` operation, which ".. any safe code .." cannot perform.

active_discussion/embedded/static-unsynchronized.md
japaric: Changing the priority is an `unsafe` operation. The invariant required to keep this safe / sound is that interrupt priorities won't be changed.

I didn't paint the whole picture here, sorry about that. In practice, priorities are exposed as something that the user sets at compile time and then never changes at runtime. Before the user code runs some runtime / framework configures the interrupt priorities before globally enabling interrupts (both operations are `unsafe`). The user code is not meant to change interrupt priorities; the API to do that is `unsafe` because it can cause UB if it enables preemption where previously there wasn't.

active_discussion/embedded/static-unsynchronized.md
```diff
@@ -0,0 +1,82 @@
...
+    } else if random() {
+        INTERRUPT1(&mut X);
+    }
+}
+```
+
+## Questions
+
+- Can this program be misoptimized given that the compiler has *no* information
+  about `INTERRUPT0` and `INTERRUPT1` executing "cooperatively"?
```
RalfJung: The crucial aliasing point here is: the moment `INTERRUPT1` starts executing, any previously created pointer gets invalidated.

So, safety relies on the assumption that the safe code cannot somehow leak the `&'static mut` that it got -- for example, through a `static X: RefCell<Option<&'static mut u128>>` or so. I assume that is possible for safe code even on BEAM, so this would not be a safe abstraction. But if you change it as follows, I think this hole is plugged:
```rust
#[no_mangle]
unsafe fn INTERRUPT1() {
    let x: &mut u128 = unsafe { &mut X };
	inner(x);

	// Crucially, inner is generic in the lifetime and hence
    // cannot leak the reference.
    fn inner(x: &mut u128) {
      // .. any safe code ..
    }
}
```

Besides this point, I agree that the program is fine as far as Stacked Borrows is concerned. I can only hope LLVM agrees with that. ;) The compiler cannot know that the functions cooperate, but when it doubt it has to assume that they do.

RalfJung:

PR #115, opened 2019-04-14
[BEAM] `&'static mut` references to static variables and LLVM `noalias`

japaric:

[Rendered discussion document](https://github.com/japaric/unsafe-code-guidelines/blob/beam-static-mut-references/active_discussion/embedded/static-mut-references.md)

### DIFF
--- a/active_discussion/embedded/static-mut-references.md
+++ b/active_discussion/embedded/static-mut-references.md
@@ -1,0 +1,167 @@
+# `&'static mut` references to static variables and LLVM `noalias`
+
+Some programs need to move values between interrupt handlers (or between `main`
+and an interrupt handler); this operation has to be done using static variables.
+Performing this send operation "by value" can result in expensive `memcpy`s
+(from handler A to the static variable and then from the variable to handler B).
+A way to reduce the runtime cost is indirection: instead of memcpy-ing the value
+(`T`) one can send (memcpy) a pointer that owns the value, like a `Box<T>`.
+
+Sometimes heap allocators are avoided in embedded code for performance or
+reliability reasons (e.g. a potential OOM condition is undesirable). In this
+scenario `Box`, `Rc`, `Arc` and other pointers are not available to the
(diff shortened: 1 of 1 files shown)


comex:
`noalias` was originally used to implement C `restrict` and has roughly equivalent semantics.  It basically means (slightly oversimplified): "as long as this pointer is in scope, I promise not to read or write the memory behind this pointer using anything but the pointer itself (or pointers derived from it)".

It's not UB to create a `noalias` pointer to a global variable.  It's only UB if you refer to the original global variable a second time while the pointer is still in scope.

I think the example should also be fine under @RalfJung's proposed "stacked borrows" memory model for Rust, though he's the expert on that.

RalfJung:
Meta-comment: in terms of process, "active_discussion" is for things in our current "area of discussion". The idea is to keep active discussion focused around one area or else we'll never get anywhere.  What you posted as a PR here would, I think, make more sense as the body of an issue.

RalfJung:
For your actual question, I concur with @comex: `&'static mut T` means "this is a unique pointer and no other pointer (not derived from this one) will be used to access said memory". It's a bit like `Box<T>` but without the heap/deallocation part. So, that seems to be exactly what you want.

`noalias_pointer` was derived from `x` in the implementation of `BumpPointerAlloc`, so that seems fine as well.

active_discussion/embedded/static-mut-references.md
```diff
@@ -0,0 +1,167 @@
...
+
+        &mut X
+    };
+
+    let alloc = BumpPointerAlloc::new(x);
+    let noalias_pointer: &'static mut u32 = unsafe {
+        &mut *(alloc.alloc(Layout::from_size_align(4, 4)) as *mut u32)
+    };
+
+    loop {
```
RalfJung: As an example for something that would *not* be okay here: even just reading from `x` or `X` will invalidate overlapping parts of `noalias_pointer`. So if the implementation of `BumpPointerAlloc` would do that, that would be a problem.

RalfJung:

PR #116, opened 2019-04-14
[BEAM] The meaning of Send and Sync in the context of interrupt handlers

japaric:

[Rendered discussion document](https://github.com/japaric/unsafe-code-guidelines/blob/beam-send-sync/active_discussion/embedded/send-sync.md)

### DIFF
--- a/active_discussion/embedded/send-sync.md
+++ b/active_discussion/embedded/send-sync.md
@@ -1,0 +1,188 @@
+# The meaning of Send and Sync in the context of interrupt handlers
+
+The [`Send`] and [`Sync`] traits are defined in `core` as:
+
+[`Send`]: https://doc.rust-lang.org/core/marker/trait.Send.html
+[`Sync`]: https://doc.rust-lang.org/core/marker/trait.Sync.html
+
+> `Send`: types that can be transferred across thread boundaries.
+
+> `Sync`: types for which it is safe to share references between threads.
+
+These definitions don't provide good guidance to embedded (microcontroller)
+developers because there may or may not be (OS) threads in embedded code. Even
+if you consider an interrupt handler to be semantically equivalent to a thread
+these definitions would still end up being more restrictive than they need to
+be.
+
+The keyword that's missing in these definitions, in my opinion, is *preemption*.
+Within an OS, preemption (context switch) could occur between any two threads
+depending on how the thread scheduler is implemented. As the program can make no
+assumptions about the scheduler always requiring the `Send` / `Sync` bounds is
+the right call. In embedded code one can statically know whether preemption will
+occur between any two interrupt handlers; if no preemption is possible then
+neither of these two bounds is required for memory safety.
+
+Note that:
+
+- The programs in this document target the [Basic Embedded Abstract Machine
+  (BEAM)][beam]. Please become familiar with the linked specification before you
+  read the rest of this document.
+
+[beam]: https://github.com/rust-lang/unsafe-code-guidelines/pull/111
+
+- In these programs we assume that [rust-lang/rfcs#2585][rfc2585] has been
+  accepted and implemented.
+
+[rfc2585]: https://github.com/rust-lang/rfcs/pull/2585
+
+## Cooperative interrupt handlers
+
+Consider this program, which is more general version of [#113]
+
+[#113]: https://github.com/rust-lang/unsafe-code-guidelines/pull/113
+
+``` rust
+#![no_std]
+
+#[no_mangle]
+unsafe fn main() -> ! {
+    unsafe {
+        asm!("ENABLE_INTERRUPTS" : : : : "volatile");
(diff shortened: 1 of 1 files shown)


active_discussion/embedded/send-sync.md
```diff
@@ -0,0 +1,188 @@
...
+}
+```
+
+Note that "any safe code" can *not* call `main`, `INTERRUPT0`, `INTERRUPT1` or
+`INTERRUPT2` (because they are `unsafe` functions), use `asm!` or access
+registers.
+
+**Claim**: this program is well-defined / sound for any `Type`. Rust aliasing
+rules are preserved because no preemption (or context switching) can occur
+between `INTERRUPT0`, `INTERRUPT1` and `INTERRUPT2`.
```
gnzlbg: Same issue here, if `INTERRUPT1` is higher priority than `INTERRUPT0`, then it can preempt it, and you might end with two `&mut X`, one in `INTERRUPT0` and one in `INTERRUPT1`, where `INTERRUPT0` will be optimized under the assumption that this cannot happen.

active_discussion/embedded/send-sync.md
```diff
@@ -0,0 +1,188 @@
...
+#[no_mangle]
+unsafe fn INTERRUPT1() {
+    let x: &Type = unsafe { &X };
+
+    // .. any safe code ..
+}
+```
+
+**Claim**: this program is well-defined / sound if and only if `Type`
+implements the `Sync` trait.
```
gnzlbg: This LGTM.

active_discussion/embedded/send-sync.md
```diff
@@ -0,0 +1,188 @@
...
+
+Example that shows that the bound is required: `type Type = Rc<u8>` could result
+in an unsound program (data race between `main` and `INTERRUPT0`).
+
+## Questions
+
+- Can these programs be misoptimized by the compiler?
+
+- Should we redefine `Send` and `Sync` (as documented in doc.rust-lang.org/std)
+  in terms of preemption?
```
gnzlbg: I don't see why this would be necessary. Interrupts are just a concurrency mechanism, and whether they are executed in the same thread or in a different one matters for the exact synchronization primitive that you need to avoid data-races (e.g. atomics vs compiler_fence, Arc vs some compiler-fenced Rc-variant, etc.) but I don't think it matters for `Sync` or `Send`.

For instance, consider: main starts on thread0, thread0 gets paused, thread1 executes interrupt0, thread1 gets paused, thread2 executes interrupt1, ..., thread1 terminates, thread0 is resumed. Whether this is how execution happens (using N concurrent threads), or whether it happens in a single thread, does not matter AFAICT for how the code running on each "thread" will be optimized. If accesses are not synchronized, the optimizer will optimize under the assumption that data-races cannot happen. If accesses are synchronized (e.g. using `compiler_fence`) the optimizer will optimize under the assumption that they are properly synchronized.

gnzlbg: Or in other words, we could define that interrupts are executed in a different thread of execution, and just note that this thread happens to be scheduled in the same OS thread that that "<whatever-condition fits here>", such that everything that applies to threads, would apply for interrupts as well. 

@RalfJung have you thought about how to model interrupts in miri  ? (e.g. under the assumption that we could make them deterministic, e.g., via a CLI option that deterministically raises a signal at some fixed point in the program execution? )

active_discussion/embedded/send-sync.md
japaric: These three interrupts have the same priority in this program. When the BEAM starts executing `main` all interrupts have a priority of `1`, and the program hasn't change this priority configuration. ".. any safe code .." cannot change the priorities because writing to the relevant register requires `unsafe` code.

RalfJung: The same issue wrt. leaking `'static` references as in https://github.com/rust-lang/unsafe-code-guidelines/pull/115 applies.

active_discussion/embedded/send-sync.md
RalfJung: No I have not.^^

I also agree with @gnzlbg that it is probably easier to define preempted interrupts in terms of threads than vice versa. So all interrupts with the same priority would run in the same thread, and as such do not have to worry about `Send`/`Sync` *between themselves*. Interrupts are the niche case here so if we can avoid it I'd rather not fit the definition to to that -- fitting it to the common case is much likely to be understood correctly by crate authors.

That said -- what kind of formulation in terms of preemption did you have in mind?

active_discussion/embedded/send-sync.md
RalfJung: The comment is marked as resolved, but I think the text could be improved to state the priorities at which the code will run explicitly.

RalfJung:

PR #120, opened 2019-04-19
Editorial changes.

crlf0710:
I'm not happy with several wordings within the text, so i did some small modifications.

Beyond this, i think `&dyn Trait + <auto trait>s + <lifetime bounds>` are actually guaranteed, but this text didn't cover it somehow. And i don't really know the correct way to write them out. Besides this, i'm still puzzled about things like the validity of `&dyn Send + Sync` and `&dyn Sync + Send`...

### DIFF
--- a/reference/src/layout/pointers.md
+++ b/reference/src/layout/pointers.md
@@ -2,18 +2,17 @@
 
 ### Terminology
 
-Reference types are types of the form `&T`, `&mut T` or `&dyn T`.
+Reference types are types of the form `&T`, `&mut T`.
 
 Raw pointer types are types of the form `*const T` or `*mut T`.
-We write `*T` when the mutability attribute is unimportant.
 
 ### Representation
 
 The alignment of `&T`, `&mut T`, `*const T` and `*mut T` are the same,
 and are at least the word size.
 
-* If `T` is a trait, then the alignment of `&dyn T` is the word size.
 * If `T` is a sized type then the alignment of `&T` is the word size.
+* The alignment of `&dyn Trait` is the word size.
 * The alignment of `&[T]` is the word size.
 * The alignment of `&str` is the word size.
 * Alignment in other cases may be more than the word size (e.g., for other dynamically sized types).
@@ -21,15 +20,15 @@ and are at least the word size.
 The sizes of `&T`, `&mut T`, `*const T` and `*mut T` are the same,
 and are at least one word.
 
-* If `T` is a trait, then the size of `&dyn T` is two words.
 * If `T` is a sized type then the size of `&T` is one word.
+* The size of `&dyn Trait` is two words.
 * The size of `&[T]` is two words.
 * The size of `&str` is two words.
 * Size in other cases may be more than one word (e.g., for other dynamically sized types).
 
 ### Notes
 
-The layouts of `&T`, `&mut T` and `*T` are the same.
+The layouts of `&T`, `&mut T`, `*const T` and `*mut T` are the same.
 
 If `T` is sized, references and pointers to `T` have a size and alignment of one
 word and have therefore the same layout as C pointers.
@@ -40,23 +39,23 @@ word and have therefore the same layout as C pointers.
 > or, in the case of `&mut T`, aliasing.
 
 We do not make any guarantees about the layout of
-multi-trait objects `&(dyn T + U)` or references to other dynamically sized types,
+multi-trait objects `&(dyn Tr + Ur)` or references to other dynamically sized types,
 other than that they are at least word-aligned, and have size at least one word.
 
-The layout of `&dyn T` when `T` is a trait is the same as that of:
-```rust,ignore
+The layout of `&dyn Trait` when `Trait` is a trait is the same as that of:
+```rust
 #[repr(C)]
 struct DynObject {
-  data: *u8,
-  vtable: *u8,
+  data: *const u8,
+  vtable: *const u8,
 }
 ```
 
 The layout of `&[T]` is the same as that of:
-```rust,ignore
+```rust
 #[repr(C)]
 struct Slice<T> {
-  ptr: *T,
+  ptr: *const T,
   len: usize,
 }
 ```


reference/src/layout/pointers.md
```diff
@@ -40,23 +39,23 @@ word and have therefore the same layout as C pointers.
 > or, in the case of `&mut T`, aliasing.
 
 We do not make any guarantees about the layout of
-multi-trait objects `&(dyn T + U)` or references to other dynamically sized types,
+multi-trait objects `&(dyn Tr + Ur)` or references to other dynamically sized types,
```
gnzlbg: Maybe use `Trait0` and `Trait1` here ?

gnzlbg:
> Beyond this, i think &dyn Trait + <auto trait>s + <lifetime bounds> are actually guaranteed, but this text didn't cover it somehow. 

Where is this guaranteed ? (Can you point to an RFC, the reference, etc.?)

reference/src/layout/pointers.md
crlf0710: Changed to Trait1 and Trait2.

crlf0710:
> Where is this guaranteed ?

Hmm, i never questioned this! No, i don't think i've seen any material explicitly said so. 

I think it is important for this to become a guarantee, or i won't be able to transmute a `&dyn Trait + Send + Sync` to a  `std::raw::TraitObject`...

reference/src/layout/pointers.md
```diff
@@ -2,34 +2,33 @@
...
 * If `T` is a sized type then the size of `&T` is one word.
+* The size of `&dyn Trait` is two words.
 * The size of `&[T]` is two words.
 * The size of `&str` is two words.
 * Size in other cases may be more than one word (e.g., for other dynamically sized types).
 
 ### Notes
 
-The layouts of `&T`, `&mut T` and `*T` are the same.
+The layouts of `&T`, `&mut T`, `*const T` and `*mut T` are the same.
```
RalfJung: Did we never define whether "layout" is just size+align+field offset or whether it includes "niches for enum optimization"?

gnzlbg: I don't think we ever did (do we have a definition of layout somewhere?), but it makes sense to define niches as part of layout. Otherwise when we specify layout recursively for structs, etc. to depend on the layout of their fields, niches cannot play a role, but they do play a role for `Option` and that's already guaranteed.

gnzlbg:
@crlf0710 

> I think it is important for this to become a guarantee, or i won't be able to transmute a &dyn Trait + Send + Sync to a std::raw::TraitObject

I've opened #121 to track that we currently don't guarantee the layout of multi-trait objects, and provide a place to discuss how to solve that.

reference/src/layout/pointers.md
RalfJung: Then the sentence here is wrong though -- these types do not all have the same niche.

gnzlbg: That's right. I don't think this PR is the right place to fix that (this was already broken), but I've opened #122 to track that, and maybe we can discuss that in the next meeting.

crlf0710:
@gnzlbg ok, would you like to merge this, or me to abandon this, or wait until #121 #122 be resolved somehow, and modify this PR?

gnzlbg:
We'll discuss it in the next meeting (next thursday), this will give time to more people to take a look at this.

reference/src/layout/pointers.md
```diff
@@ -40,23 +39,23 @@ word and have therefore the same layout as C pointers.
...
 
-The layout of `&dyn T` when `T` is a trait is the same as that of:
-```rust,ignore
+The layout of `&dyn Trait` when `Trait` is a trait is the same as that of:
+```rust
 #[repr(C)]
 struct DynObject {
-  data: *u8,
-  vtable: *u8,
+  data: *const u8,
```
RalfJung: We deliberately did not have `mut` or `const` here because it kind of depends on whether this is used in a shared or mutable pointer. `const` unconditionally here is wrong.

reference/src/layout/pointers.md
RalfJung: Fair.

RalfJung:
I think it's fine to merge this as long as it doesn't make new guarantees -- and I don't think it does. :)  See my one remaining open comment, though.

reference/src/layout/pointers.md
crlf0710: This is associated with `&dyn T`, so i think it's a shared pointer? or am i totally wrong about this?

By the way, does this mean `std::raw::TraitObject` (which only has a mut variant) is fundamentally broken?

And i feel the associated example in docs: ```data: &other_value as *const _ as *mut (),``` is even more broken https://doc.rust-lang.org/nightly/std/raw/struct.TraitObject.html

gnzlbg: @RalfJung the document only claims that the layout is equivalent to the layout of these structs, not that we internally use these structs to represent stuff.

I would prefer if these could be Rust code, and I wouldn't mind having two types here, `DynObject` and `DynObjectMut` if that's what it takes.

RalfJung: > This is associated with &dyn T, so i think it's a shared pointer? or am i totally wrong about this?

This is, I think, meant to apply to `&dyn T` and `&mut dyn T` alike. It would be weird to have to repeat all that.

Fair. So maybe add a comment saying that `data` might be mutable if this is an `&mut dyn`?

crlf0710: Done.

reference/src/layout/pointers.md
```diff
@@ -40,25 +39,30 @@ word and have therefore the same layout as C pointers.
...
+  vtable: *const u8,
 }
 ```
 
+> **note**: In the layout equivalence of `&dyn mut Trait` the field `data` is of the type `*mut u8`.
```
RalfJung:
```suggestion
> **note**: In the layout of `&dyn mut Trait` the field `data` is of the type `*mut u8`.
```

reference/src/layout/pointers.md
```diff
@@ -40,25 +39,30 @@ word and have therefore the same layout as C pointers.
...
 #[repr(C)]
 struct Slice<T> {
-  ptr: *T,
+  ptr: *const T,
   len: usize,
 }
 ```
 
-The layout of `&str` is the same as that of `&[u8]`.
+> **note**: In the layout equivalence of `&mut [T]` the field `ptr` is of the type `*mut T`.
```
RalfJung:
```suggestion
> **note**: In the layout of `&mut [T]` the field `ptr` is of the type `*mut T`.
```

Centril:
> Beyond this, i think `&dyn Trait + <auto trait>s + <lifetime bounds>` are actually guaranteed

We do *not* guarantee the layout of trait objects in any shape or form *on stable*. As for `raw::TraitObject`, it is only a guarantee on nightly and only if you use that struct (making your own `repr(C)` type on stable does not count), which may go away at any time.

PR #132, opened 2019-05-25
Update README

gnzlbg:
Updates the readme:

* add links / badges to the UCGs book
* updates some links (active discussion topic, etc.)
* updates meeting times
* tries to keep it brief

[Rendered](https://github.com/gnzlbg/unsafe-code-guidelines/blob/readme_update/README.md).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: rust
 sudo: false
 rust: nightly
 cache: cargo
-install: cargo install mdbook
+install: cargo install mdbook || true
 script: bash -c "cd reference && mdbook build && mdbook test"
 env:
   RUSTINFRA_DEPLOY_DIR: reference/book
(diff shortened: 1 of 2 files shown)


RalfJung:
CI is not happy
```
1.07s$ cargo install mdbook
    Updating crates.io index
error: binary `mdbook` already exists in destination as part of `mdbook v0.2.3`
Add --force to overwrite
The command "cargo install mdbook" failed and exited with 101 during .
```

gnzlbg:
CI is green now, it appears that installing a package that's cached can fail.. I really hate that about cargo. If I write `cargo install foo`,`foo` is already installed, and the installed version is exactly the one that would have been installed (not older, etc.), then the command should just succeed, and not fail.

RalfJung:
Seems like you want https://github.com/rust-lang/cargo/pull/6798?

gnzlbg:
Nice! That's great! It currently requires `cargo +nightly install mdbook -Z install-upgrade`, but might become the default behavior soon!

PR #140, opened 2019-06-03
add introduction to book

RalfJung:


reference/book.toml
```diff
@@ -1,5 +1,5 @@
 [book]
-authors = ["Niko Matsakis"]
+authors = ["UCG WG"]
```
gnzlbg:
```suggestion
authors = ["The Rust Project Developers"]
```

RalfJung: Is that what we usually do? Sure.

RalfJung:
Nobody complained for 10 days and one "LGTM", I am going to merge this.

Issue #73, opened 2019-01-10
Validity of unions


RalfJung:
Discussing the validity invariant of unions.

One possible choice here is "none, any bit pattern is allowed no matter which types the fields have, and including uninitialized bits".

We could also decide that e.g. a
```rust
union Foo { a: bool, b: (bool, u8) }
```
must start with the first byte being either the bit-pattern of `false` or the bit-pattern of `true`, because all fields agree on that invariant.

Notice that we *cannot* require the union to be valid for some field: for a union like
```rust
union Mix {
  f1: (bool, u8),
  f2: (u8, bool),
}
```
we want to allow a bit pattern like `0x3 0x3`, which can occur from code like
```rust
let m = Mix { f1: (false, 3) };
m.f2.0 = 3;
```
There is no demonstrated benefit from disallowing such code, and this kind of code seems perfectly reasonable around unions.

Given that, any validity invariant that wants to restrict the set of allowed bit patterns will be rather complicated.  However, such an invariant would enable us to e.g. layout-optimize `Option<Foo>`, whereas the "anything goes"-invariant would prohibit any kind of layout optimization around unions.

nikomatsakis:
> My personal preference is to allow any bit pattern, mostly to keep things simple. Unions are already complex enough, and they only occur in unsafe code, we should make this as simple to use for the programmer as we can.

I feel like @JoshTripplet and @cramertj have both, in the past, expressed strong opinions about this.

I think I agree regarding allowing any bit pattern. That said, I **can** definitely imagine wanting to be able to create a `union` that is "as optimized" as the equivalent `enum`, but that it has no discriminant (because you know you can figure that out from other means).

nikomatsakis:
@RalfJung 

> We could also decide ... must start with the first byte being either the bit-pattern of false or the bit-pattern of true, because all fields agree on that invariant.

It seems like this would be true only if the Rust compiler decided to lay the fields out at offset zero, right? Personally, I sort of think we should just guarantee that the Rust compiler **will** do so. Particularly if we decide that unions are an opaque "bag of bits" from the perspective of the compiler, what is the motivation for the compiler to add extra padding into that bag?

RalfJung:
> It seems like this would be true only if the Rust compiler decided to lay the fields out at offset zero, right? 

Yes, I've been assuming that to be the base.

> I feel like @joshtriplett and @cramertj have both, in the past, expressed strong opinions about this.

[...]

PR #147, opened 2019-06-20
Validity of unions with a zero-sized field

gnzlbg:


reference/src/validity/unions.md
```diff
@@ -0,0 +1,25 @@
+# Validity of unions
+
+**Disclaimer**: This chapter is a work-in-progress. What's contained here
+represents the consensus from issue [#73]. The statements in here are not (yet)
+"guaranteed" not to change until an RFC ratifies them.
+
+The bit `i` of an union is allowed to have value `v` _iff_ there is a variant of
```
rkruppe: This isn't consensus in #73 and so it shouldn't be written down here.

gnzlbg: Latest commit removes this definition and the rationale.

reference/src/validity/unions.md
```diff
@@ -0,0 +1,13 @@
+# Validity of unions
+
+**Disclaimer**: This chapter is a work-in-progress. What's contained here
+represents the consensus from issue [#73]. The statements in here are not (yet)
+"guaranteed" not to change until an RFC ratifies them.
+
+## Validity of unions with zero-sized fields
+
+A union containing a zero-sized field can contain any value. An example of such
```
rkruppe:
```suggestion
A union containing a zero-sized field can contain any bit pattern. An example of such
```

rkruppe: BTW: still not sure if that's the best way to put it but this change is at least an improvement

PR #149, opened 2019-06-22
Update discriminant-elision optimization on Option-like enums

gnzlbg:
This PR updates the discriminant-elision optimization on option-like enums with the semantics specified in https://github.com/rust-lang/rust/pull/60300 

The main change is requiring that option-like enums do not have any explicit `#[repr(...)]` attributes.

### DIFF
--- a/reference/src/glossary.md
+++ b/reference/src/glossary.md
@@ -64,7 +64,8 @@ In this document, *layout* and *representation* are not synonyms.
 The *niche* of a type determines invalid bit-patterns that will be used by layout optimizations.
 
 For example, `&mut T` has at least one niche, the "all zeros" bit-pattern. This
-niche is used by layout optimizations like "`enum` discriminant elision" to
+niche is used by layout optimizations like ["`enum` discriminant
+elision"](layout/enums.html#discriminant-elision-on-option-like-enums) to
 guarantee that `Option<&mut T>` has the same size as `&mut T`.
 
 While all niches are invalid bit-patterns, not all invalid bit-patterns are
--- a/reference/src/glossary.md
+++ b/reference/src/glossary.md
@@ -61,7 +61,7 @@ In this document, *layout* and *representation* are not synonyms.
 
 #### Niche
 
-Invalid bit-patterns that will be used by layout optimizations.
+The *niche* of a type determines invalid bit-patterns that will be used by layout optimizations.
 
 For example, `&mut T` has at least one niche, the "all zeros" bit-pattern. This
 niche is used by layout optimizations like "`enum` discriminant elision" to
--- a/reference/src/glossary.md
+++ b/reference/src/glossary.md
@@ -61,11 +61,16 @@ In this document, *layout* and *representation* are not synonyms.
 
 #### Niche
 
-Invalid bit-patters that will be used by layout optimizations.
+Invalid bit-patterns that will be used by layout optimizations.
 
 For example, `&mut T` has at least one niche, the "all zeros" bit-pattern. This
 niche is used by layout optimizations like "`enum` discriminant elision" to
 guarantee that `Option<&mut T>` has the same size as `&mut T`.
+
+While all niches are invalid bit-patterns, not all invalid bit-patterns are
+niches. For example, the "all bits uninitialized" is an invalid bit-pattern for
+`&mut T`, but this bit-pattern is not used by layout optimizations, and is not a
+niche.
 
 
 ### TODO
--- a/reference/src/glossary.md
+++ b/reference/src/glossary.md
@@ -69,7 +69,7 @@ guarantee that `Option<&mut T>` has the same size as `&mut T`.
 
 While all niches are invalid bit-patterns, not all invalid bit-patterns are
 niches. For example, the "all bits uninitialized" is an invalid bit-pattern for
-`&mut T`, but this bit-pattern is not used by layout optimizations, and is not a
+`&mut T`, but this bit-pattern cannot be used by layout optimizations, and is not a
 niche.
 
 
--- a/reference/src/glossary.md
+++ b/reference/src/glossary.md
@@ -59,9 +59,17 @@ Moreover, the layout of a type records its *function call ABI* (or just *ABI* fo
 Note: Originally, *layout* and *representation* were treated as synonyms, and Rust language features like the `#[repr]` attribute reflect this. 
 In this document, *layout* and *representation* are not synonyms.
 
+#### Niche
+
+Invalid bit-patters that will be used by layout optimizations.
+
+For example, `&mut T` has at least one niche, the "all zeros" bit-pattern. This
+niche is used by layout optimizations like "`enum` discriminant elision" to
+guarantee that `Option<&mut T>` has the same size as `&mut T`.
+
+
 ### TODO
 
-* *niche*
 * *tag*
 * *rvalue*
 * *lvalue*
--- a/reference/src/layout/enums.md
+++ b/reference/src/layout/enums.md
@@ -323,7 +323,8 @@ as a niche.
 
 The niche values of a type are parts of its validity invariant which, as of this
 writing, is the current active discussion topic in the unsafe code guidelines
-process. [rust-lang/rust#60300] specifies that the following types have a niche:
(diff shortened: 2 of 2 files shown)


reference/src/layout/enums.md
```diff
@@ -292,11 +292,15 @@ apply, as described below.
 
 #### Discriminant elision on Option-like enums
 
-(Meta-note: The content in this section is not described by any RFC
-and is therefore "non-normative".)
+(Meta-note: The content in this section is not fully described by any RFC and is
+therefore "non-normative". Parts of it were specified in
+[rust-lang/rust#60300].
```
RalfJung:
```suggestion
[rust-lang/rust#60300]).
```
Missing a closing parenthesis.

reference/src/layout/enums.md
```diff
@@ -313,13 +317,21 @@ values, which are called **niches**. For example, a value of type `&T`
...
-description of their validity invariant, which -- as of this writing
--- are the next topic up for discussion in the unsafe code guidelines
-process.)
+as a niche. 
 
 [`NonZeroU8`]: https://doc.rust-lang.org/std/num/struct.NonZeroU8.html
 
+The niche values of a type are parts of its validity invariant which, as of this
+writing, is the current active discussion topic in the unsafe code guidelines
+process. [rust-lang/rust#60300] specifies that the following types have a niche:
```
RalfJung: Should this maybe say what the niche is? I think we technically only fixed a lower bound, but still.

RalfJung:
This still does not specify, though, that transmuting between e.g. `u32` and `Option<NonZeroU32>` is safe, right? For that we have to say that the `Some` case of that option is represented exactly like the inner type, and the `None` case is 0. I guess this works for types that have a "single-value niche", which is the case for the `NonZero*`/`NonNull` types. For `&T`, their niche *currently* is just 0, but in the future might grow to also include some unaligned values (e.g., `&u32` could have a niche consisting of the values `0..=3`).

reference/src/layout/enums.md
gnzlbg: There is a TODO in the glossary for that.

RalfJung: No I don't mean define what the term niche means. I mean say concretely something like "The niche of `NonZeroU8` is the value 0".

gnzlbg: Ah, I misunderstood. We can try to explicitly say that, e.g., for `NonZero`, `NonNull`, etc. But as you mentioned elsewere this might be harder for `&T`, where zero is definitely a niche, but there might be others, and those are currently unclear.

RalfJung: We can say the niche includes at least "0". That's probably something we want to guarantee.

That's why I said we only have a lower bound.

gnzlbg: Fixed

reference/src/layout/enums.md
```diff
@@ -313,13 +317,22 @@ values, which are called **niches**. For example, a value of type `&T`
...
 and friends may never be zero, and hence also define the value of `0`
-as a niche. (Types that define niche values will say so as part of the
-description of their validity invariant, which -- as of this writing
--- are the next topic up for discussion in the unsafe code guidelines
-process.)
+as a niche. 
 
 [`NonZeroU8`]: https://doc.rust-lang.org/std/num/struct.NonZeroU8.html
 
+The niche values of a type are parts of its validity invariant which, as of this
```
RalfJung: The niche is not "part of the invariant", I don't even know what that means.

The niche values are *some of* values that variables of this type can never have. For example, a `&mut u32` can also never have value 1337, and yet it is unlikely that that will be in the niche. Currently it certainly is not.

If you want to draw a connection to validity, you can say that the niche values must be disjoint from the values allowed by the validity invariant.

RalfJung: Oh and if you are anyway defining here what "niche" means, wouldn't it make sense to put that into the glossary?

gnzlbg: > If you want to draw a connection to validity, you can say that the niche values must be disjoint from the values allowed by the validity invariant

Done.

> a &mut u32 can also never have value 1337, and yet it is unlikely that that will be in the niche. 

Why is this unlikely? Not all invalid values can be niches (e.g. the "all-bits-uninitialized" bit-pattern), but for references, unaligned addresses are invalid and could be niches. I don't see a reason why they shouldn't be.

gnzlbg: > what "niche" means, wouldn't it make sense to put that into the glossary?

I'm not sure how to briefly define niche. Maybe something like "An invalid bit-pattern that can be used by layout optimizations.

RalfJung: > Why is this unlikely? Not all invalid values can be niches (e.g. the "all-bits-uninitialized" bit-pattern), but for references, unaligned addresses are invalid and could be niches. I don't see a reason why they shouldn't be.

It doesn't fit the current architecture very well.
I guess long-term there's no problem with it, and certainly not in terms of spec, but short-term it seems unlikely.

> I'm not sure how to briefly define niche.

You are doing that already, right here. ;)

> Maybe something like "An invalid bit-pattern that can be used by layout optimizations".

Or "*will* be used"? That's what differentiates it from bit patterns that are ruled out by the validity invariant but *not* in the niche.

reference/src/layout/enums.md
```diff
@@ -313,13 +317,23 @@ values, which are called **niches**. For example, a value of type `&T`
...
-as a niche. (Types that define niche values will say so as part of the
-description of their validity invariant, which -- as of this writing
--- are the next topic up for discussion in the unsafe code guidelines
-process.)
+as a niche. 
 
 [`NonZeroU8`]: https://doc.rust-lang.org/std/num/struct.NonZeroU8.html
 
+The niche values must be disjoint from the values allowed by the validity
+invariant. The validity invariant is, as of this writing, is the current active
```
RalfJung:
```suggestion
invariant. The validity invariant is, as of this writing, the current active
```

reference/src/layout/enums.md
gnzlbg: I've added a definition of niche to the glossary. I'm resolving this, we can discuss that definition there.

reference/src/glossary.md
```diff
@@ -59,9 +59,22 @@ Moreover, the layout of a type records its *function call ABI* (or just *ABI* fo
 Note: Originally, *layout* and *representation* were treated as synonyms, and Rust language features like the `#[repr]` attribute reflect this. 
 In this document, *layout* and *representation* are not synonyms.
 
+#### Niche
+
+Invalid bit-patterns that will be used by layout optimizations.
```
```suggestion
The *niche* of a type determines invalid bit-patterns that will be used by layout optimizations.
```

This is not a proper sentence. Also text should make sense if you skip the headers.

reference/src/glossary.md
```diff
@@ -59,9 +59,22 @@ Moreover, the layout of a type records its *function call ABI* (or just *ABI* fo
...
+
+For example, `&mut T` has at least one niche, the "all zeros" bit-pattern. This
+niche is used by layout optimizations like "`enum` discriminant elision" to
```
RalfJung: Can this link to the enum layout document where we describe that layout optimization?

reference/src/glossary.md
```diff
@@ -59,9 +59,22 @@ Moreover, the layout of a type records its *function call ABI* (or just *ABI* fo
...
+guarantee that `Option<&mut T>` has the same size as `&mut T`.
+
+While all niches are invalid bit-patterns, not all invalid bit-patterns are
+niches. For example, the "all bits uninitialized" is an invalid bit-pattern for
+`&mut T`, but this bit-pattern is not used by layout optimizations, and is not a
```
```suggestion
`&mut T`, but this bit-pattern cannot be used by layout optimizations, and is not a
```

reference/src/layout/enums.md
```diff
@@ -313,13 +317,23 @@ values, which are called **niches**. For example, a value of type `&T`
 may never be `NULL`, and hence defines a niche consisting of the
```
RalfJung: This paragraph should somewhere link to the glossary.

gnzlbg:
The last commit hyperlinks everything.

PR #153, opened 2019-06-26
Make niches part of layout and expand call ABI

### LINKED ISSUE #122: Should niches/ABI be part of the layout of a type?
The current definition of layout (https://github.com/rust-lang/unsafe-code-guidelines/blob/master/reference/src/glossary.md#layout) does not consider "niches" part of the type layout. 

In this thread (https://github.com/rust-lang/unsafe-code-guidelines/pull/120#discussion_r276975529) it was argued that maybe we might want to change that and make them part of the layout of a type. 

If we do that, we need to change the glossary, and distinguish that `&mut T` and `*mut T` don't have the same layout, because they don't have the same "niches".

cc @eddyb

gnzlbg:
Closes #122 

cc @eddyb  

---

(Once the other PRs are merged I'll update the links here).

### DIFF
--- a/reference/src/glossary.md
+++ b/reference/src/glossary.md
@@ -58,8 +58,8 @@ The *layout* of a type defines:
 * its size,
 * its alignment,
 * its field offsets (e.g. fields of structs, union, enums, arrays, ...)
-* its [niches][Niche],
-* its call abi.
+* its [niches](glossary.html#niche),
+* its [call abi](glossary.html#call-abi).
 
 Note: due to discriminant-elision optimizations, niches are required to compute
 the layout of, e.g., `Option<T>`, from the layout of `T`.
--- a/reference/src/glossary.md
+++ b/reference/src/glossary.md
@@ -53,8 +53,13 @@ For some more information, see [this blog post](https://www.ralfj.de/blog/2018/0
 
 #### Layout
 
-The *layout* of a type defines its size and alignment as well as the offsets of its subobjects (e.g. fields of structs/unions/enum/... or elements of arrays).
-Moreover, the layout of a type records its *function call ABI* (or just *ABI* for short): how the type is passed *by value* across a function boundary.
+The *layout* of a type defines:
+
+* its size,
+* its alignment,
+* its field offsets (e.g. fields of structs, union, enums, arrays, ...)
+* its [niches][Niche],
+* its call abi.
 
 Note: Originally, *layout* and *representation* were treated as synonyms, and Rust language features like the `#[repr]` attribute reflect this. 
 In this document, *layout* and *representation* are not synonyms.
@@ -73,6 +78,28 @@ niches. For example, the "all bits uninitialized" is an invalid bit-pattern for
 `&mut T`, but this bit-pattern cannot be used by layout optimizations, and is not a
 niche.
 
+#### Call ABI
+
+The call ABI is how a type is passed *by value* across a function boundary.
+
+Note: The call ABI is not stable. Currently, it is one of:
+
+```rust,norun
+pub enum Abi {
+    Uninhabited,
+    Scalar(Scalar),
+    Vector {
+        element: Scalar,
+        count: u64
+    },
+    Aggregate {
+        // If true, the layout size of the aggregate is exact.
+        // Otherwise, the layout size of the aggregate is
+        // a lower bound.
+        sized: bool,
+    }
+}
+```
 
 ### TODO
 
--- a/reference/src/glossary.md
+++ b/reference/src/glossary.md
@@ -85,7 +85,7 @@ the layout of, e.g., `Option<T>`, from the layout of `T`.
 
 #### Call ABI
 
-The call ABI is how a type is passed *by value* across a function boundary.
+The *call ABI* determines how a type is passed *by value* across a function boundary.
 
 Note: The call ABI is not stable. Currently, it is one of:
 
--- a/reference/src/glossary.md
+++ b/reference/src/glossary.md
@@ -89,8 +89,8 @@ The *call ABI* determines how a type is passed *by value* across a function boun
 
 Note: The call ABI is not stable. Currently, it is one of:
 
-```rust,norun
-pub enum Abi {
+```rust,ignore
+enum Abi {
     Uninhabited,
     Scalar(Scalar),
     Vector {
--- a/reference/src/glossary.md
+++ b/reference/src/glossary.md
@@ -58,8 +58,8 @@ The *layout* of a type defines:
 * its size,
 * its alignment,
 * its field offsets (e.g. fields of structs, union, enums, arrays, ...)
-* its [niches](glossary.html#niche),
-* its [call abi](glossary.html#call-abi).
+* its [niches](#niche),
+* its [call abi](#call-abi).
 
 Note: due to discriminant-elision optimizations, niches are required to compute
 the layout of, e.g., `Option<T>`, from the layout of `T`.
--- a/reference/src/glossary.md
+++ b/reference/src/glossary.md
@@ -106,6 +106,10 @@ pub enum Abi {
 }
 ```
 
+For example, the call ABI of `i32` is `Scalar`. The call ABI of `struct
+Wrapper(i32);` is also `Scalar`. But the call ABI of `#[repr(C)] struct
+Wrapper(i32)l` is `Aggregate { sized: true }`.
+
 ### TODO
 
 * *tag*
--- a/reference/src/glossary.md
+++ b/reference/src/glossary.md
@@ -61,10 +61,12 @@ The *layout* of a type defines:
 * its [niches][Niche],
 * its call abi.
 
+Note: due to discriminant-elision optimizations, niches are required to compute
+the layout of, e.g., `Option<T>`, from the layout of `T`.
+
 Note: Originally, *layout* and *representation* were treated as synonyms, and Rust language features like the `#[repr]` attribute reflect this. 
 In this document, *layout* and *representation* are not synonyms.
 
-<<<<<<< HEAD
 #### Niche
 
 The *niche* of a type determines invalid bit-patterns that will be used by layout optimizations.
@@ -78,16 +80,12 @@ While all niches are invalid bit-patterns, not all invalid bit-patterns are
 niches. For example, the "all bits uninitialized" is an invalid bit-pattern for
 `&mut T`, but this bit-pattern cannot be used by layout optimizations, and is not a
 niche.
-=======
-Note: due to discriminant-elision optimizations, niches are required to compute
-the layout of, e.g., `Option<T>`, from the layout of `T`.
->>>>>>> Add note about why niches need to be part of the layout
 
 #### Call ABI
 
 The *call ABI* determines how a type is passed *by value* across a function boundary.
 
-Note: The call ABI is not stable. Currently, it is one of:
+Note: The set of possible call ABIs is not stable. Currently, it consists of:
 
 ```rust,ignore
 enum Abi {
--- a/reference/src/glossary.md
+++ b/reference/src/glossary.md
@@ -106,34 +106,30 @@ pub enum Abi {
 }
 ```
 
-where
+where:
 
 ```rust,norun
 enum Integer { I8, I16, I32, I64, I128, }
 enum FloatTy { F32, F64, }
 
-enum Primitive {
+enum Scalar {
     Int(Integer, /*signedness:*/ bool),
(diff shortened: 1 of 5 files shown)


reference/src/glossary.md
```diff
@@ -53,12 +53,47 @@ For some more information, see [this blog post](https://www.ralfj.de/blog/2018/0
...
 
 Note: Originally, *layout* and *representation* were treated as synonyms, and Rust language features like the `#[repr]` attribute reflect this. 
 In this document, *layout* and *representation* are not synonyms.
 
+Note: due to discriminant-elision optimizations, niches are required to compute
+the layout of, e.g., `Option<T>`, from the layout of `T`.
+
+#### Call ABI
+
+The call ABI is how a type is passed *by value* across a function boundary.
```
```suggestion
The *call ABI* is how a type is passed *by value* across a function boundary.
```

reference/src/glossary.md
```diff
@@ -53,12 +53,47 @@ For some more information, see [this blog post](https://www.ralfj.de/blog/2018/0
...
+    Aggregate {
+        // If true, the layout size of the aggregate is exact.
+        // Otherwise, the layout size of the aggregate is
+        // a lower bound.
+        sized: bool,
+    }
+}
+```
+
+For example, the call ABI of `i32` is `Scalar`. The call ABI of `struct
```
RalfJung: `Scalar` of what though? That variant has a field.

reference/src/glossary.md
```diff
@@ -53,12 +53,47 @@ For some more information, see [this blog post](https://www.ralfj.de/blog/2018/0
...
+Wrapper(i32);` is also `Scalar`. But the call ABI of `#[repr(C)] struct
```
RalfJung: Maybe use `#[repr(transparent)] struct Wrapper`? And then say that for `repr(Rust)` the ABI is not guaranteed?

gnzlbg: The call ABI of `repr(Rust)` types is never guaranteed AFAICT, but the call ABI of `struct Wrapper(i32);` (without `repr(transparent)`) is "scalar", not "aggregate".

gnzlbg:
So I've added some clarifications to the ABI and the examples.

reference/src/glossary.md
```diff
@@ -53,12 +53,72 @@ For some more information, see [this blog post](https://www.ralfj.de/blog/2018/0
...
+
+enum Primitive {
+    Int(Integer, /*signedness:*/ bool),
+    Float(FloatTy),
+    Pointer
+}
+
+struct Scalar {
+    value: Primitive,
+    valid_range: RangeInclusive<u128>,
```
RalfJung: This already encodes a niche, though...

reference/src/glossary.md
```diff
@@ -53,12 +53,72 @@ For some more information, see [this blog post](https://www.ralfj.de/blog/2018/0
...
+}
+```
+
+For example, right now, the call ABI of:
+
+* `i32` is `Scalar`,
+* `struct Wrapper(i32);` (without `repr(transparent)`) is also `Scalar` (only larger `struct`s are aggregates),
```
RalfJung: Given that you mention `repr(Rust)` below, I'd also write `#[repr(Rust)] struct Wrapper` here. Otherwise people have to know that this is the default to even make the connection.

reference/src/glossary.md
```diff
@@ -53,12 +53,72 @@ For some more information, see [this blog post](https://www.ralfj.de/blog/2018/0
...
+* `#[repr(C)] struct Wrapper(i32);` is `Aggregate { sized: true }` - required by `repr(C****.
+
+**NOTE**: the call ABI of `repr(Rust)` types is unspecified, so the "call ABI"
+of `repr(Rust)` types is not guaranteed and `unsafe` code shall not rely on it.
+That is, code that requires `struct Wrapper(i32);` to have the same ABI as `i32`
+needs to use `repr(transparent)`. f
```
RalfJung: The example list mixes some things that are guaranteed and some that are not. I don't think that's a good idea.

Also you have a spurious ` f` at the end here.

reference/src/glossary.md
RalfJung: But really I wouldn't add all this detail, that's just distracting.

reference/src/glossary.md
```diff
@@ -53,12 +53,72 @@ For some more information, see [this blog post](https://www.ralfj.de/blog/2018/0
...
+
+struct Scalar {
+    value: Primitive,
+    valid_range: RangeInclusive<u128>,
+}
+```
+
+For example, right now, the call ABI of:
+
+* `i32` is `Scalar`,
```
RalfJung: You added all the detail and still don't say "`Scalar` or what" here. ;)

What about "`Scalar(s)`, where `s` encodes the size and signedness of `i32`".

reference/src/glossary.md
gnzlbg: cc @eddyb - so apparently niches are part of the call ABI, at least for scalars

gnzlbg: I've simplified this a bit, but I'd prefer to remove all this detail. I would still to clarify the question of "What is Scalar" though. I can probably say "where `Scalar` is a primitive integer, floating point, or raw pointer to a sized type.

RalfJung: I'm all for removing the definition of the `Scalar` type.

RalfJung: Oh, you did. Great. :)

reference/src/glossary.md
```diff
@@ -53,12 +53,58 @@ For some more information, see [this blog post](https://www.ralfj.de/blog/2018/0
...
 In this document, *layout* and *representation* are not synonyms.
 
+Note: due to discriminant-elision optimizations, niches are required to compute
+the layout of, e.g., `Option<T>`, from the layout of `T`.
+
+#### Call ABI
+
+The *call ABI* determines how a type is passed *by value* across a function boundary.
+
+Note: The call ABI is not stable. Currently, it is one of:
```
```suggestion
Note: The set of possible call ABIs is not stable. Currently, it consists of:
```

RalfJung:
@gnzlbg this PR has conflicts now. Can you resolve that, and fix the nit? Then I am going to merge. This has sat long enough with an "approved" seal and nobody complaining.

eddyb:
Kind of related, we can do the same in the compiler too: https://github.com/rust-lang/rust/pull/62692
(due to initialization concerns, we only can ever use at most once niche)

gnzlbg:
Done.

reference/src/glossary.md
```diff
@@ -53,8 +53,16 @@ For some more information, see [this blog post](https://www.ralfj.de/blog/2018/0
...
 
-The *layout* of a type defines its size and alignment as well as the offsets of its subobjects (e.g. fields of structs/unions/enum/... or elements of arrays).
-Moreover, the layout of a type records its *function call ABI* (or just *ABI* for short): how the type is passed *by value* across a function boundary.
+The *layout* of a type defines:
+
+* its size,
+* its alignment,
+* its field offsets (e.g. fields of structs, union, enums, arrays, ...)
+* its [niches][Niche],
+* its call abi.
```
RalfJung: This could link to the call ABI definition below.

RalfJung:
Hm okay I am having second thoughts about merging this without hearing from anyone else in the WG. After all the (re)defines a key term.

@rust-lang/wg-unsafe-code-guidelines please share your thoughts. :)

Also, do we need to update any of the other existing text that talks about "layout"? I am thinking in particular of statements of the form "A and B have the same layout".

reference/src/glossary.md
```diff
@@ -53,8 +53,16 @@ For some more information, see [this blog post](https://www.ralfj.de/blog/2018/0
...
 #### Layout
 
-The *layout* of a type defines its size and alignment as well as the offsets of its subobjects (e.g. fields of structs/unions/enum/... or elements of arrays).
-Moreover, the layout of a type records its *function call ABI* (or just *ABI* for short): how the type is passed *by value* across a function boundary.
+The *layout* of a type defines:
+
+* its size,
+* its alignment,
+* its field offsets (e.g. fields of structs, union, enums, arrays, ...)
+* its [niches][Niche],
```
RalfJung: Where is the `Niche` target defined in the markdown?

gnzlbg:
> Also, do we need to update any of the other existing text that talks about "layout"?

We probably should open an issue to go through all uses of "layout" in the text and update them. We could already start doing that, but the `union`s discussions might end up adding some form of "padding" component to layout, so I'd rather settle that first before going through the whole text updating things.

RalfJung:
Saying incorrect things until the update has been carried out seems bad, though.

gnzlbg:
So I went through all uses of "layout" and updated everything except for struct and enum documents. 

I don't think anything "incorrect" in these two documents, but they mix the term "layout" with "memory layout" quite freely, and they also use the term C-compatible layout, but AFAICT C has no notion of niches, so I'm unsure what to do there.

RalfJung:
Thanks! That looks great.

"C-compatible layout" might be worth defining here? But leaving that for later is also okay.

However, I *would* like to see an ack of some other WG member before merging.

gnzlbg:
> "C-compatible layout" might be worth defining here? 

I think we need a different term for "size+alignment" (e.g. "memory layout" ?), "size+alignment+call AB" (e.g. "abi" ?), and what Rust uses for layout optimizations (size+alignment+abi+niche, e.g., just "layout"). 

Otherwise being explicit about each term gets repetitive and error prone, and it is hard to refactor as we iterate on "layout"'s definition.

RalfJung:
> "size+alignment" (e.g. "memory layout" ?),

I like "shape", but I am not sure what exactly that should include.

ehsanmok:
How about

1. "size+alignment" = memory layout
2. memory layout + call abi = call layout
3. call layout + niches = type layout ?

Lokathor:
I'd maybe put call layout at the end of the list. There's many data transforms you might care to do that worry about size/align/niche but don't care about call

gnzlbg:
FWIW this isn't stale. The PRs #159, #161, and #164 all use "layout" in the same way that the glossary does now. 

I prefer to wait for some version of those to be merged before replacing "layout" with something more specific everywhere, since the diff of doing that might help us tune things a bit.

reference/src/glossary.md
```diff
@@ -73,6 +81,43 @@ niches. For example, the "all bits uninitialized" is an invalid bit-pattern for
...
+#### Call ABI
+
+The *call ABI* determines how a type is passed *by value* across a function boundary.
+
+Note: The set of possible call ABIs is not stable. Currently, it consists of:
+
+```rust,ignore
+enum Abi {
+    Uninhabited,
+    Scalar(Scalar),
```
bjorn3: `ScalarPair` is missing.

RalfJung:
Also see https://github.com/rust-lang/unsafe-code-guidelines/issues/122#issuecomment-520225053, where we are going back to the drawing board and considering calling these sub-aspects of "layout" "ABIs". So there would be a bunch of ABIs (call ABI, memory ABI, maybe something like "nesting ABI" that includes a niche and means things are the same even when nested in other types?"), and all of that together would be called "layout". Or so.

reference/src/glossary.md
gnzlbg: That's intentional. ScalarPair is an implementation detail useful for some optimizations that we don't want to guarantee it exists - right @eddyb ?

gnzlbg:
> Or so.

Yes. Just to confirm, the temporary plan until we have a better solution is to continue to call this "Layout", but writing down for each case what exactly do we guarantee (SAN-layout), right ?

RalfJung:
That was the plan before "X ABI" got proposed and gained some traction. Not sure if it still is the plan.

gnzlbg:
I think it would be at least helpful to know what do we precisely guarantee where for coming up with the "X" in the "X ABI" plan.

reference/src/glossary.md
eddyb: Yeah, it's just an aggregate that we pass around as two scalars *only* in Rust ABIs and within functions.

RalfJung:
@gnzlbg given that this will need some rewording and rebasing etc. anyway, are you fine with going ahead with https://github.com/rust-lang/unsafe-code-guidelines/issues/191 while this PR is still open?

gnzlbg:
Sure @Lokathor go ahead, this will need many changes anyways.

RalfJung:
There was no activity here for quite a while so I am going to close this PR; looks like more discussion in https://github.com/rust-lang/unsafe-code-guidelines/issues/122 is needed.

Issue #9, opened 2018-08-30
Representation of bool, integers and floating points


nikomatsakis:
This issue is to discuss the memory layout for integral and floating point types:

- `bool`
- `u8`..`u128`, `i8`..`i128`
- `usize`, `isize`
- `f32`, `f64`

For the most part, these are relatively uncontroversial. However, there are some interesting things worth discussing:

- Unlike other types, there are no `#[repr(C)]` vs `#[repr(Rust)]` variants here. The size is always fixed and well-defined across FFI boundaries. The types map to their corresponding integral types in the surrounding C ABI. 
- Prior discussions ([#46156][], [#46176][]) documented bool as a single byte that is either 0 or 1.
- How is `usize` intended to be defined on various platforms?
    - the native size of a pointer?
    - [the max of various other considerations](https://github.com/rust-rfcs/unsafe-code-guidelines/pull/5#discussion_r212702266)?
    - other edge cases to consider?
- Rust currently states that the maximum size of any single value must fit in with `isize`
      - Can we say a bit more about why? (e.g., [ensuring that "pointer diff" is representable](https://github.com/rust-rfcs/unsafe-code-guidelines/pull/5#discussion_r212703192)
- Do we want to discuss signaling NaN at all? Specifically: why is it potentially of concern, and are there things that unsafe authors or other folks need to be aware of? (@Gankro, for example, wrote that ["NaN masking is unnecessary from LLVM's perspective"](https://github.com/rust-lang/rust/issues/40470#issuecomment-343803381), but I don't really know what that means. =)

joshtriplett:
Don't forget `u128` and `i128`.

Another topic: Does FFI code *need* to use types like `c_uint64`, or is it safe to just use `u64`?

nikomatsakis:
@joshtriplett 

> Don't forget `u128` and `i128`.

Edited, thanks!

> Another topic: Does FFI code need to use types like `libc::uint64_t`, or is it safe to just use u64?

Good question! I presume it does not, but I'd be curious if there is another side to the discussion.

**Related note:* clearly `usize` and C's `unsized` are not equivalent, but it's worth stating this explicitly.

Gankro:
> Another topic: Does FFI code need to use types like c_uint64, or is it safe to just use u64?

Much of the rust ecosystem (e.g. webrender and therefore firefox) assumes uint64_t and u64 are the same ABI-wise, and I'm unaware of any reason to prevent that assumption.

[...]

PR #154, opened 2019-07-01
Document that isize/usize are at least 16-bit wide on all platforms we currently support

gnzlbg:
This is something that we already document, but was not explicitly spelled out.

cc @Gankro

### DIFF
--- a/reference/src/layout/scalars.md
+++ b/reference/src/layout/scalars.md
@@ -69,7 +69,11 @@ range of values that the pointer-holding integer types must support:
 * `intptr_t`: `[INTPTR_MIN, INTPTR_MAX] = [-(2^15 - 1), 2^15 - 1]` 
 * `uintptr_t`: `[0, UINTPTR_MAX] = [0, 2^16 - 1]`
 
-That is, `isize`/`usize` are at least 16-bit wide.
+That is, `isize`/`usize` are at least 16-bit wide. `libcore` relies on
+this guarantee and unconditionally provides:
+
+* `impl From<i16> for isize`,
+* `impl From<u16> for usize`.
 
 </details>
 
--- a/reference/src/layout/scalars.md
+++ b/reference/src/layout/scalars.md
@@ -37,11 +37,11 @@ The `isize` and `usize` types are pointer-sized signed and unsigned integers.
 They have the same layout as the [pointer types] for which the pointee is
 `Sized`, and are layout compatible with C's `uintptr_t` and `intptr_t` types.
 
-> **Note**: on all platforms that we currently support, `usize`/`isize` are at
-> least 16-bit wide. C99
-> [7.18.2.4](https://port70.net/~nsz/c/c99/n1256.html#7.18.2.4) requires
-> `uintptr_t` and `intptr_t` to be at least 16-bit wide. `libcore` relies on
-> this to unconditionally provide impls of `From<i16>`/`From<u16>` for
+> **Note**: C99 [7.18.2.4](https://port70.net/~nsz/c/c99/n1256.html#7.18.2.4)
+> requires `uintptr_t` and `intptr_t` to be at least 16-bit wide. That is, on
+> all platforms that we currently support, which have a C platform,
+> `isize`/`usize` are at least 16-bit wide. `libcore` relies on this guarantee
+> to unconditionally provide impls of `From<i16>`/`From<u16>` for
 > `isize`/`usize`.
 
 > **Note**: Rust's `usize` and C's `unsigned` types are **not** equivalent. C's
--- a/reference/src/layout/scalars.md
+++ b/reference/src/layout/scalars.md
@@ -35,8 +35,14 @@ of `char` is _implementation-defined_.
 
 The `isize` and `usize` types are pointer-sized signed and unsigned integers.
 They have the same layout as the [pointer types] for which the pointee is
-`Sized`, and are layout compatible with C's `uintptr_t` and `intptr_t` types,
-and are therefore at least 16-bit wide.
+`Sized`, and are layout compatible with C's `uintptr_t` and `intptr_t` types.
+
+> **Note**: on all platforms that we currently support, `usize`/`isize` are at
+> least 16-bit wide. C99
+> [7.18.2.4](https://port70.net/~nsz/c/c99/n1256.html#7.18.2.4) requires
+> `uintptr_t` and `intptr_t` to be at least 16-bit wide. `libcore` relies on
+> this to unconditionally provide impls of `From<i16>`/`From<u16>` for
+> `isize`/`usize`.
 
 > **Note**: Rust's `usize` and C's `unsigned` types are **not** equivalent. C's
 > `unsigned` is at least as large as a short, allowed to have padding bits, etc.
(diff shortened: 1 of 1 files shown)


gnzlbg:
TL;DR: this documents that for all practical purposes `isize`/`usize` are at least 16-bit wide, but leaves the door open for 8-bit platform support. I think that any stronger wording would make supporting those platforms illegal, and I don't know if this is something we want. I don't see what that would buy us.

---

So I started guaranteeing that `isize`/`usize` are at least 16-bit wide, and falled back to just documenting that, for all platforms we currently support, this is the case.

The current guarantee says that the layout of `isize`/`usize` matches that of `intptr_t`/`uintptr_t`, so if those are 16-bit wide, which they are on all platforms that we currently support, then this holds for `isize`/`usize` as well. 

However, this also means that if our platform doesn't have a C platform, or a C standard conforming platform, not much about the layout of `isize`/`usize` is guaranteed. This could be useful, if somebody wanted to ever support 8-bit platforms. C does not support them, so Rust could add support for them in whatever way it sees fit, and still remain C-compatible in 16-bit platforms and up. 

There is the "issue" of `libcore` providing unconditional impls of `From<i16>`/`From<u16>` for `isize`/`usize`. These impls are correct for all platforms we currently support, so that's fine. An 8-bit platform can just `cfg` these impls out. Code that relies on them would just fail to compile, and AFAICT, that's ok.

comex:
Note that even "8-bit" CPUs mostly have 16-bit address spaces, even though that means that a memory address cannot fit into a single register.  For example, in the Intel 8008 and derivatives, addresses are specified using two 8-bit registers 'H' and 'L' (high and low), which are concatenated to form a 16-bit address.

gnzlbg:
Good point. We can drop the last commit, and make this a real guarantee. I'm not sue what the full implications of that are, or whether we are already making this guarantee, so I changed that to at least document current behavior. But if we can guarantee this without draw-backs, we should.

8-bit pointers sound like a extremely limiting concept to me. If one is restricted by that, writing code for such a target probably won't look like Rust at all.

reference/src/layout/scalars.md
```diff
@@ -37,6 +37,13 @@ The `isize` and `usize` types are pointer-sized signed and unsigned integers.
 They have the same layout as the [pointer types] for which the pointee is
 `Sized`, and are layout compatible with C's `uintptr_t` and `intptr_t` types.
 
+> **Note**: C99 [7.18.2.4](https://port70.net/~nsz/c/c99/n1256.html#7.18.2.4)
+> requires `uintptr_t` and `intptr_t` to be at least 16-bit wide. That is, on
+> all platforms that we currently support, which have a C platform,
+> `isize`/`usize` are at least 16-bit wide. `libcore` relies on this guarantee
+> to unconditionally provide impls of `From<i16>`/`From<u16>` for
```
petrochenkov: `libcore` doesn't really rely on this, it just effectively has `#[cfg(all_platforms_supported_right_now)]` on the impl.
If we didn't support 16-bit platforms yet, `From<i32>` would also be unconditional.
`cfg`-s can be added on every new platform support if necessary.

I don't understand why this guarantee is useful, other libraries are free to support any platforms they want, I'm pretty sure many of them are not tested on 16-bit either and therefore not expected to work without fixes.

gnzlbg: First, we do agree. 

> libcore doesn't really rely on this, it just effectively has #[cfg(all_platforms_supported_right_now)] on the impl.
If we didn't support 16-bit platforms yet, From<i32> would also be unconditional.
cfg-s can be added on every new platform support if necessary.

The main argument for adding that API was that C guarantees that `uintptr_t`/`intptr_t` are at least 16-bit wide (the source code even calls it as such). I mentioned [here](https://github.com/rust-lang/unsafe-code-guidelines/pull/154#issuecomment-507142684), that there is no technical reason we couldn't cfg them out if we ever added support for an 8-bit platform: that wouldn't break any code there, because no code there exists. Still, all non-portable APIs in libcore are in `core::arch`, and this one isn't. Whether this should be available for all Rust programs or not, is kind of scope for the UCGs (more of a T-libs/T-lang/T-compiler issue). 

> I don't understand why this guarantee is useful, other libraries are free to support any platforms they want, I'm pretty sure many of them are not tested on 16-bit either and therefore not expected to work without fixes.

Note that this PR does not guarantee anything. It documents something that is true for all platforms that Rust supports today, and explicitly calls it as such. 

Why would guaranteeing this be useful ? It would make this code "`if size_of::<usize>() < 2 { ... }`" dead for all Rust programs, removing one corner case / bit-of-trivia that programmers have to remember, or that can be used to accuse code of "not being portable". I do see some value in this. 

Are we already guaranteeing this ? I don't know, I don't think so. [In previous discussions](https://github.com/rust-lang/unsafe-code-guidelines/issues/9), it appeared to be common folklore that there are some platforms that we never intend to support, like platforms where `CHAR_BITS != 8`, but 8-bit platforms were not mentioned there. 

This is why the wording is so "generic". `usize`/`isize` correspond to `uintptr_t`/`intptr_t` of the C platform, iff a C platform exists. A standard compliant C platform cannot exist for 8-bit architectures, so there we can do something else, as long as we don't break targets where a C platform exists. Right now, that something else is unspecified.

Lokathor: I have been told that once the [portability lint](https://github.com/rust-lang/rust/issues/41619) feature gets farther, then we'll be in a much better position to make "potentially not portable" code in core and std. One of the exact examples given is that, with the portability lint ability in place, we'd be able to add From/Into for usize/isize/u32/i32/u64/i64 and they get cfg'd in based on your platform width, and so on.

Of course, it's had basically no progress since 2017 it looks like, so rip.

rkruppe: Yeah, there's plausible ways for libcore to be ported to a new platform and lose the impl there, so let's not say it "relies" on this. Just noting the relation to C is good enough for now.

gnzlbg: I've reworded this a bit - is it better now?

rkruppe: LGTM

reference/src/layout/scalars.md
```diff
@@ -37,6 +37,12 @@ The `isize` and `usize` types are pointer-sized signed and unsigned integers.
 They have the same layout as the [pointer types] for which the pointee is
 `Sized`, and are layout compatible with C's `uintptr_t` and `intptr_t` types.
 
+> **Note**: C99 [7.18.2.4](https://port70.net/~nsz/c/c99/n1256.html#7.18.2.4)
+> requires `uintptr_t` and `intptr_t` to be at least 16-bit wide. That is, on
+> all platforms that we currently support, which have a C platform,
```
RalfJung: Are you saying "if we support a platform and it has a C platform, then they are at least 16 bits wide", or "if we support a platform, then it has a C platform, and hence they are at least 16 bits wide"? Please clarify the wording.

Suggestion:
```
C99 [7.18.2.4](https://port70.net/~nsz/c/c99/n1256.html#7.18.2.4)
requires `uintptr_t` and `intptr_t` to be at least 16-bit wide.
All platforms we currently support have a C platform, and as a consequence,
`isize`/`usize` are at least 16-bit wide for all of them.
```

briansmith:
This doesn't belong in the unsafe code guidelines. It belongs in the core language documentation. In particular, it affects whether or not we can implement `From<u16> for usize` which doesn't have anything to do with unsafe code.

reference/src/layout/scalars.md
gnzlbg: Done

gnzlbg:
>  It belongs in the core language documentation.

@briansmith feel free to open a PR to the core language documentation to document this there as well. although I'm not sure where you could put this there. Probably in the `usize`/`isize` docs, the reference part talking about integers, and the nomicon part talking about integers.

RalfJung:
@briansmith note that this here is just a statement about "all platforms we support right now". It doesn't exclude the possibility of adding support for other platforms in the future; that is beyond the WG's power.

Issue #37, opened 2018-10-11
defining zero-sized structs


nikomatsakis:
From #31:

"If you have a struct which -- transitively -- contains no data of non-zero size, then the size of that struct will be zero as well. These zero-sized structs appear frequently as exceptions in other layout considerations (e.g., single-field structs). An example of such a struct is `std::marker::PhantomData`."

Is that a sufficient definition for zero-sized structs? This seems like an important guarantee that we frequently rely upon for performance and other purposes, so it is worth specifying.

rkruppe:
I'd rather spell out exactly the structural recursion that is going rather than hand wave it away with "transitively", but yes I think we should guarantee:

- Arrays are ZSTs if the count is zero or the element type is ZST
- A tuple is ZST if all elements are ZST
- A struct is ZST if all fields are ZST
- Empty enums are ZSTs

Enums are interesting because we could try to be smart around enums with a single variant but that will likely depend on the `repr` attribute which doesn't seem the case for structs (`#[repr(C)] struct Foo {}` is currently a ZST).

asajeffrey:
What about a one-case enum whose body is zero-sized, e.g.:
```
enum Foo { OnlyCase(SomethingZeroSized) }
```
or a union all of whose cases are zero-sized?

rkruppe:
The only-variant-containing-ZST case could also be defined as ZST depending on repr attributes, but doesn't seem very useful.

Unions, good point, could probably be treated the same as structs, i.e., could be ZSTs even with repr(C).

rodrimati1992:
Generic enums with 2 or more variants are zero-sized if every variant is Void except for a zero-sized variant.
https://play.rust-lang.org/?gist=0b3278bdd497b1bc3c9bccbb4ca828df&version=stable&mode=debug&edition=2015

Mark-Simulacrum:
> Empty enums are ZSTs

Aren't empty enums considered uninhabited? I'm not sure if those types are technically ZSTs since there's no actual valid representation.

rkruppe:
@rodrimati1992 True, but I'd rather leave that sort of reasoning out of the picture for the initial step because:

- We would presumably want to generalize it to uninhabited types more generally rather than special case `enum Void {}`, but the exact definition of uninhabitedness is still under discussion (e.g., https://github.com/rust-lang/rust/pull/54125)
[...]

PR #159, opened 2019-07-08
Resolve layout of single variant enums

### LINKED ISSUE #79: Layout of single-variant enums
Enums that contain a **single variant** and which do not have an
explicit `#[repr]` annotation are an important special case. Since
there is only a single variant, the enum must be instantiated with
that variant, which means that the enum is in fact equivalent to a
struct. The question then is to what extent we should **guarantee**
that the two share an equivalent layout, and also how to define the
interaction with uninhabited types.

As presently implemented, the compiler will use the same layout for
structs and for single variant enums (as long as they do not have a
`#[repr]` annotation that overrides that choice). So, for example, the
struct `SomeStruct` and the enum `SomeEnum` would have an equivalent
layout ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=3697ac684c3d021892694956df957653))::

```rust
struct SomeStruct;
enum SomeEnum {
  SomeVariant,
}
```

Similarly, the struct `SomeStruct` and the enum `SomeVariant` in this
example would also be equivalent in their layout
([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=924724764419f846c788a8763da45992)):

```rust
struct SomeStruct { x: u32 }
enum SomeEnum {
  SomeVariant { x: u32 },
}
```

In fact, the compiler will use this optimized layout even for enums
that define multiple variants, as long as all but one of the variants
is uninhabited
([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=3cc1484c5b91097f3dc2015b7c207a0e)):

```rust
struct SomeStruct { x: u32 }
enum SomeEnum {
  SomeVariant { x: u32 },
  UninhabitedVariant { y: Void },
}
```

gnzlbg:
Addresses part of #79. Addresses part of #37 .

I don't see any reasons not to guarantee this, and AFAICT nobody has mentioned any either.

---

EDIT: Note that this doesn't say anything about the layout of enums without variants, or where all variants are uninhabited. That should better be tackled in a different PR.

### DIFF
--- a/reference/src/layout/enums.md
+++ b/reference/src/layout/enums.md
@@ -380,6 +380,8 @@ enum Enum2<T> {
 
 ### Layout of single variant enums
 
+> **NOTE**: the guarantees in this section have not been approved by an RFC process.
+
 **Single variant data-carrying*** enums without a `repr()` annotation have the
 same layout as the variant field. **Single variant fieldless** enums have the
 same layout as a unit struct.
@@ -419,6 +421,8 @@ the single-variant data-carrying enum `SingleVariantDataCarrying` has the same
 layout as `SomeStruct`.
 
 ### Layout of multi-variant enums with one inhabited variant
+
+> **NOTE**: the guarantees in this section have not been approved by an RFC process.
 
 The layout of **multi-variant** enums with **one inhabited variant** is the same
 as that of the single-variant enum containing that same inhabited variant.
--- a/reference/src/layout/enums.md
+++ b/reference/src/layout/enums.md
@@ -378,90 +378,72 @@ enum Enum2<T> {
 
 [niche]: ../glossary.html#niche
 
-## Unresolved questions
-
 ### Layout of single variant enums
 
-[Issue #79.](https://github.com/rust-rfcs/unsafe-code-guidelines/issues/79)
+**Single variant data-carrying*** enums without a `repr()` annotation have the
+same layout as the variant field. **Single variant fieldless** enums have the
+same layout as a unit struct.
 
-Enums that contain a **single variant** and which do not have an
-explicit `#[repr]` annotation are an important special case. Since
-there is only a single variant, the enum must be instantiated with
-that variant, which means that the enum is in fact equivalent to a
-struct. The question then is to what extent we should **guarantee**
-that the two share an equivalent layout, and also how to define the
-interaction with uninhabited types.
-
-As presently implemented, the compiler will use the same layout for
-structs and for single variant enums (as long as they do not have a
-`#[repr]` annotation that overrides that choice). So, for example, the
-struct `SomeStruct` and the enum `SomeEnum` would have an equivalent
-layout ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=3697ac684c3d021892694956df957653)):
+Here:
 
 ```rust
-struct SomeStruct;
-enum SomeEnum {
-  SomeVariant,
-}
-
+# use std::mem::{size_of, align_of};
+struct UnitStruct;
+enum SingleVariantFieldless { FieldlessVariant }
 # fn main() {
-# use std::mem;
-let x = SomeStruct;
-let y = SomeEnum::SomeVariant;
-assert_eq!(
-    mem::size_of_val(&x), 
-    mem::size_of_val(&y),
-    "types have different sizes",
-);
-println!("types both have size {}", std::mem::size_of_val(&x));
+assert_eq!(size_of::<SingleVariantFieldless>(), size_of::<UnitStruct>());
+assert_eq!(align_of::<SingleVariantFieldless>(), align_of::<UnitStruct>());
+// assert_eq!(call_abi_of::<SingleVariantFieldless>(), call_abi_of::<UnitStruct>());
+// assert_eq!(niches_of::<SingleVariantFieldless>(), niches_of::<UnitStruct>());
 # }
 ```
 
-Similarly, the struct `SomeStruct` and the enum `SomeVariant` in this
-example would also be equivalent in their layout
-([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=924724764419f846c788a8763da45992)):
+the single-variant fieldless enum `SingleVariantFieldless` has the same layout
+as `UnitStruct`.
+
+Here: 
 
 ```rust
+# use std::mem::{size_of, align_of};
 struct SomeStruct { x: u32 }
-enum SomeEnum {
-  SomeVariant { x: u32 },
+enum SingleVariantDataCarrying {
+  DataCarryingVariant(SomeStruct),
 }
-
 # fn main() {
-# use std::mem;
-let x = SomeStruct { x: 22 };
-let y = SomeEnum::SomeVariant { x: 22 };
-assert_eq!(
-    mem::size_of_val(&x), 
-    mem::size_of_val(&y),
-    "types have different sizes",
-);
-println!("types both have size {}", mem::size_of_val(&x));
-}
-```
-
-In fact, the compiler will use this optimized layout even for enums
-that define multiple variants, as long as all but one of the variants
-is uninhabited
-([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=3cc1484c5b91097f3dc2015b7c207a0e)):
-
-```rust
-# enum Void { }
-struct SomeStruct { x: u32 }
-enum SomeEnum {
-  SomeVariant { x: u32 },
-  UninhabitedVariant { y: Void },
-}
-
-# fn main() {
-# use std::mem;
-let x = SomeStruct { x: 22 };
-let y = SomeEnum::SomeVariant { x: 22 };
-assert_eq!(
-    mem::size_of_val(&x), 
-    mem::size_of_val(&y),
-    "types have different sizes",
-);
-println!("types both have size {}", mem::size_of_val(&x));
+# assert_eq!(size_of::<SingleVariantDataCarrying>(), size_of::<SomeStruct>());
+# assert_eq!(align_of::<SingleVariantDataCarrying>(), align_of::<SomeStruct>());
 # }
 ```
+
+the single-variant data-carrying enum `SingleVariantDataCarrying` has the same
+layout as `SomeStruct`.
+
+### Layout of multi-variant enums with one inhabited variant
+
+The layout of **multi-variant** enums with **one inhabited variant** is the same
+as that of the single-variant enum containing that same inhabited variant.
+
+Here:
+
+```rust
+# use std::mem::{size_of, align_of};
+struct SomeStruct { x: u32 }
+enum SingleVariantDataCarrying {
+    DataCarryingVariant(SomeStruct),
+}
+enum Void0 {}
+enum Void1 {}
+enum MultiVariantSingleHabited {
+  DataCarryingVariant(SomeStruct),
+  Uninhabited0(Void0),
+  Uninhabited1(Void1),
+}
+# fn main() {
+# use std::mem;
+# assert_eq!(size_of::<SingleVariantDataCarrying>(), size_of::<MultiVariantSingleHabited>());
+# assert_eq!(align_of::<SingleVariantDataCarrying>(), align_of::<MultiVariantSingleHabited>());
+# }
+```
+
+the `MultiVariantSingleHabited` enum has the same layout as
+`SingleVariantDataCarrying`.
--- a/reference/src/layout/enums.md
+++ b/reference/src/layout/enums.md
@@ -405,4 +405,4 @@ enum DataCarryingSingleVariant {
 
 See [Issue #79.](https://github.com/rust-rfcs/unsafe-code-guidelines/issues/79):
   
-* Layout of multi-variant enums with one inhabited variant.
+* Layout of multi-variant enums where only one variant is inhabited.
--- a/reference/src/layout/enums.md
+++ b/reference/src/layout/enums.md
@@ -46,10 +46,10 @@ As in C, discriminant values that are not specified are defined as
 either 0 (for the first variant) or as one more than the prior
 variant.
 
-**Data-carrying enums.** Enums whose variants have fields are called
-"data-carrying" enums. Note that for the purposes of this definition,
-it is not relevant whether those fields are zero-sized. Therefore this
-enum is considered "data-carrying":
+**Data-carrying enums.** Enums with at least one variant with fields are called
+"data-carrying" enums. Note that for the purposes of this definition, it is not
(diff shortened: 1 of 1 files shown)


danielhenrymantilla:
What about `#[non_exhaustive]` `enum`s?

gnzlbg:
AFAICT `#[non_exhaustive]` does not interact with layout.

rkruppe:
To elaborate: users of non-exhaustive enums can't rely on the number of variants not changing under their feet, so for their purposes this clause would not guarantee anything about the layout of the non-exhaustive enum, since it is not guaranteed to meet the precondition (only one variant). Similarly, most guarantees about the layout of structs automatically wouldn't apply to non-exhaustive structs since fields might be added to them.

(Within the same crate/module that defines the non-exhaustive enum, it would technically be allowed, though I guess it's a bit fragile to rely on it when you intend to expend the set of variants later.)

gnzlbg:
@rkruppe 

I don't know what value would we add by not allowing such code to rely on the `#[non_exhaustive]` enums layout; AFAICT there aren't any extra optimizations that this would allow. 

So I think I disagree (you kind of hint things both ways) ? I think that the layout guarantees specified here apply to both `#[non_exhaustive]` and "exhaustive" enums because `#[non_exhaustive]` does not affect layout and `unsafe` code is allowed to rely on these.

Whether that `unsafe` code is safe or not would be a different issue, but `unsafe` code written in the same module as a `#[non_exhaustive]` enum can make use of layout guarantees _and_ be safe.

rkruppe:
I didn't mean to imply layout computation or what reliance on layout is considered UB by the Rust language should be affected by non_exhaustive. I was just speaking of contracts between libraries: an external library that assumes for soundness that a non_exhaustive struct or enum has a certain layout (based on information that is strictly ) is just as wrong as if it relies on private fields or other information not part of the API. As we both noted, code within the same API boundary as the non-exhaustive type is not subject to the same restrictions.

gnzlbg:
@rkruppe 

> I was just speaking of contracts between libraries: an external library that assumes for soundness that a non_exhaustive struct or enum has a certain layout (based on information that is strictly ) is just as wrong as if it relies on private fields or other information not part of the API. 

Such a library is ugly, brittle, non-portable (to different library versions / toolchain), etc. but it is only wrong if its assumptions are incorrect.

If a Rust library exposes a `#[non_exhaustive] #[repr(C)] enum`, and one wants to interface with that library from C, then one can make it work, e.g., by just fixing the exact version of the dependency and toolchain and manually inspecting the variants of the enum.

Ideally, nobody would ever need to do this, and such APIs are obviously bad, but doing this types of things is the bread and butter of, for example, the `libc` crate. 

I don't think we can do much about this either. Making a type public leaks its layout, and while features like `#[non_exhaustive]` protect users from accidental misuse, they can't prevent a motivated user from exploiting layout for profit, nor should that be their goal. The only thing I can think of that we could make to help, would be to make the rules clearer, so that those who need to exploit API details know where they stand, and maybe give them better tools, so that they don't have to rely on undefined behavior for their hacks.

reference/src/layout/enums.md
```diff
@@ -351,90 +351,72 @@ enum Enum1<T> {
...
-[Issue #79.](https://github.com/rust-rfcs/unsafe-code-guidelines/issues/79)
-
-Enums that contain a **single variant** and which do not have an
-explicit `#[repr]` annotation are an important special case. Since
-there is only a single variant, the enum must be instantiated with
-that variant, which means that the enum is in fact equivalent to a
-struct. The question then is to what extent we should **guarantee**
-that the two share an equivalent layout, and also how to define the
-interaction with uninhabited types.
+**Single variant data-carrying*** enums without a `repr()` annotation have the
```
RalfJung: Have we defined this precisely enough that people will not confuse this with "enums where only a single variant carris data"?

reference/src/layout/enums.md
```diff
@@ -351,90 +351,72 @@ enum Enum1<T> {
...
-assert_eq!(
-    mem::size_of_val(&x), 
-    mem::size_of_val(&y),
-    "types have different sizes",
-);
-println!("types both have size {}", std::mem::size_of_val(&x));
+assert_eq!(size_of::<SingleVariantFieldless>(), size_of::<UnitStruct>());
+assert_eq!(align_of::<SingleVariantFieldless>(), align_of::<UnitStruct>());
+// assert_eq!(call_abi_of::<SingleVariantFieldless>(), call_abi_of::<UnitStruct>());
+// assert_eq!(niches_of::<SingleVariantFieldless>(), niches_of::<UnitStruct>());
```
RalfJung: What is the point of these commented-out assertions?

reference/src/layout/enums.md
```diff
@@ -351,90 +351,72 @@ enum Enum1<T> {
...
-In fact, the compiler will use this optimized layout even for enums
-that define multiple variants, as long as all but one of the variants
-is uninhabited
-([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=3cc1484c5b91097f3dc2015b7c207a0e)):
+the single-variant data-carrying enum `SingleVariantDataCarrying` has the same
+layout as `SomeStruct`.
+
+### Layout of multi-variant enums with one inhabited variant
+
+The layout of **multi-variant** enums with **one inhabited variant** is the same
```
RalfJung: Have we defined "inhabited" sufficiently precisely for this?

I am a bit worried about the details of this definition.

RalfJung:
The file says at the beginning that we try to mark what is RFC'ed and what not. Should these new sections get such comments?

reference/src/layout/enums.md
gnzlbg: The definition of data-carrying and fieldless enums is at the beginning of the document. This PR uses "data-carrying" and "fieldless" with the meaning of those definitions. 

Whether the definitions themselves are clear enough or not, I don't know. They were clear to me, e.g., they explicitly call out that `enum E { Varian0(Data) }` is data-carrying even if `Data` is a ZST and `E` has size zero, and they also call out that `enum F { Variant0, Variant1 }` is fieldless even though its size is not zero. 

One can always improve the definition to make them more clear, but that should probably happen in a different PR. I kind of wish that these two definitions had their place in the glossary as well.

reference/src/layout/enums.md
gnzlbg: We don't have any way to check those aspects of layout in code, so they are pseudo-code for that. I can remove them, since these are already implied when the text says "have the same layout". I thought the assertions made it clearer that by layout we meant _layout_.

reference/src/layout/enums.md
gnzlbg: There is no definition of uninhabited or inhabited types anywhere in the reference. 

What worries you about this definition ? If only one variant is inhabited, all other variants are uninhabited, and there can be no value of the enum that contains any of those variants, according to all definitions of uninhabited types that I can imagine.

gnzlbg:
> Should these new sections get such comments?

They probably should, I'll add that these sections have not gone through the RFC process, although the other sections more explicitly call what has gone through the process.

Before turning the reference or a subset of it into an RFC, we are going to have to do multiple rounds of making everything consistent, and that will include deciding to call out either what has been RFC'ed already, what hasn't, or both.

As layout has shown, some of the fundamental definitions evolve over time as we work through newer topics like validity. Each time they change, we kind of have to go through everything.

reference/src/layout/enums.md
Lokathor: "data carrying" on its own isn't very clear, but when paired with "fieldless" it becomes (i think) very clear.

reference/src/layout/enums.md
Lokathor: They're good asserts, and hopefully some day they'll be real rust functions as well

reference/src/layout/enums.md
gnzlbg: I'm resolving this conversation. If someone wants to improve the definitions of "fieldless" and "data-carrying" at the beginning of this document or in the glossary PRs welcome. If someone wants to discuss renaming "data-carrying" enums to something else, please open an issue. This PR is not the place to do either.

reference/src/layout/enums.md
RalfJung: The fact that they are commented out could also be interpreted as "this does not currently hold", as one would do in a test suite.

So, it should be made clearer that they are commented out just because the methods don't exist. Probably stating "these things are the same" as Rust asserts is not the clearest way anyway, and it *is* very verbose.

reference/src/layout/enums.md
RalfJung: There are at least three different kinds of "inhabited":
* Type tas a call ABI other than `Uninhabited`.
* The type's validity invariant is satisfiable.
* The type's safety invariant is satisfiable.

So, the term is ambiguous without further clarification.

gnzlbg: Are there types for which "The type's validity invariant is satisfiable" and the call ABI is `Uninhabited` ?

RalfJung: Good question! I don't know. `&!` might be such a type, depending on how the validity invariant discussions turn out.

Declaring that these two notions (the first two in my list) are equal would put a restriction on the validity invariant discussion: we would have to make sure that satisfiability is trivially computable and can be put into the rustc layout code. I think this would be a nice property to have, but I wouldn't want to assert now that we will end up there.

gnzlbg: We defined that `&T where T: Sized` has the same alignment and size as an `usize`. _I think_ this means that the call ABI of `&T where T: Sized` is `Scalar`. Is this correct? If so, for `&!` to have an `Uninhabited` call ABI `!` must not be `Sized`.

gnzlbg: So we might need to go back to the layout of pointers and references and require `where T: Sized + SomeFormOfInhabited`.

RalfJung: > If so, if &! has an Uninhabited call ABI then ! cannot be Sized, which would mean that any type containing a ! is not Sized..

Not really -- if an `&!` is impossible to construct, it doesn't really matter which call ABI we ascribe to it.

RalfJung: But this does indeed show that treating "uninhabited" as a call ABI might not be the best idea, at least in terms of teaching.

gnzlbg: > it doesn't really matter which call ABI we ascribe to it.

Then we don't have to give it Uninhabited as a call ABI, and we go back to my original question (Are there types for which "The type's validity invariant is satisfiable" and the call ABI is Uninhabited ?). 

---

We say in many places that `&T` and `*T` have the same call ABI when `T: Sized`, we would need to add exceptions for `&!` and `*!` if this is not the case. Also we do need to deal with types for which the call ABI of `&!` might matter, e.g., what's the call ABI of `Option<&!>` ? Ideally we want it to be compatible with `*!`, but I don't see how to make that happen if `&!` has a different call ABI than `*!`.

RalfJung: If we stop considering "uninhabited" a call ABI, what happens is that "layout" gets one more property: "inhabited".

So the first item of my list changes to "The type is considered inhabited in its layout".

RalfJung: > what's the call ABI of Option<&!>

Oh gosh, I have no idea.^^ Probably if we really made `&!` uninhabited that would become a ZST... ouch.

I am not sure what the right framework here is. But people have been bringing up that being able to optimize away methods taking `&!` (because they are dead code) could be very beneficial. That's the main motivation for caring about validity of the memory "behind" the reference in https://github.com/rust-lang/unsafe-code-guidelines/issues/77.

gnzlbg: Ok, so we really do need to work out the layout of uninhabited types before we do anything else here.

I do think that "inhabited", whatever that means, will need to hold here for this guarantee. Whether `&!` is inhabited or not, doesn't change anything here AFAICT.

reference/src/layout/enums.md
```diff
@@ -351,90 +351,76 @@ enum Enum1<T> {
 }
```
RalfJung: Can you edit (further up) the definition of "data carrying" to say "enums where at least one variant ..."? The current "Enums whose variants" is ambiguous IMO.

reference/src/layout/enums.md
RalfJung: I think the way you are stacking adjectives here is juts not very clear. "Data-carrying enums with a single variant" would be better, IMO.

reference/src/layout/enums.md
RalfJung: Popping back up, I don't think we should talk about "(un)inhabited" types in these documents before we got those definitions set. So I am in favor of dropping this subsection.

gnzlbg:
I think all issues have been resolved. Please take a look.

gnzlbg:
I've recorded the proposal for multi-variant enums with one inhabited variant in the tracking issue: https://github.com/rust-lang/unsafe-code-guidelines/issues/79#issuecomment-521518976

reference/src/layout/enums.md
```diff
@@ -378,90 +378,31 @@ enum Enum2<T> {
...
 
-Enums that contain a **single variant** and which do not have an
-explicit `#[repr]` annotation are an important special case. Since
-there is only a single variant, the enum must be instantiated with
-that variant, which means that the enum is in fact equivalent to a
-struct. The question then is to what extent we should **guarantee**
-that the two share an equivalent layout, and also how to define the
-interaction with uninhabited types.
+**Data-carrying** enums with a single variant without a `repr()` annotation have
+the same layout as the variant field. **Fieldless** enums with a single variant
```
RalfJung: This is deliberately still use the vague "same layout", right?

reference/src/layout/enums.md
```diff
@@ -378,90 +378,31 @@ enum Enum2<T> {
...
-    mem::size_of_val(&x), 
-    mem::size_of_val(&y),
-    "types have different sizes",
-);
-println!("types both have size {}", mem::size_of_val(&x));
-# }
-```
+See [Issue #79.](https://github.com/rust-rfcs/unsafe-code-guidelines/issues/79):
+  
+* Layout of multi-variant enums with one inhabited variant.
```
RalfJung:
```suggestion
* Layout of multi-variant enums where only one variant is inhabited.
```

reference/src/layout/enums.md
gnzlbg: Right. Making this more precise should be done as part of #153 (I'd like to merge this first, so that we can see the changes in 153 and discuss the naming there).

gnzlbg:
@RalfJung nit resolved

RalfJung:
@rkruppe was the merge here fine with you? (Your last comment in this PR was quite a while ago.)

rkruppe:
Yes!

Issue #156, opened 2019-07-03
Layout of repr(C) unions has padding


gnzlbg:
In [this thread](https://github.com/rust-lang/rust/issues/60405#issuecomment-506973741) and the comments that follow, the following new information was discovered.

`repr(C)` unions have padding bits of the form:

* trailing padding: if the union size is larger than the size of its largest field, all `union_size - largest_field_size` trailing bits are padding bits.
* interior padding: if the bit `i` of all union fields is a padding bit, the bit `i` of the inion is a padding bit

> Note: because all `repr(C)` union fields are at offset 0, there is no padding _before_ any field.
    
The content of padding bits of `repr(C)` unions is always uninitialized. That is, they are not required to be preserved on copy / move / pass by value, etc. The implementation of the call ABI can exploit this knowledge. 

For example, Rust, clang, and GCC all implement the SysV64 ABI, and when passing a `#[repr(C)] union U { x: (u8, u32) }` around by value, @eddyb [mentioned](https://github.com/rust-lang/rust/issues/60405#issuecomment-508120704) that Rust and GCC pass the bottom 32-bits (where the `u8` is stored) while clang passes the bottom 8-bits. Both implementations are allowed. @comex also [mentioned](https://github.com/rust-lang/rust/issues/60405#issuecomment-507838717) that in some ABIs like RISC-V ELF "appears to require callers to zero- or sign-extend 
arguments in registers in a particular way. In other words, it requires the upper bits (which correspond to padding bytes) to have a specific value, and the callee can assume that they do have that value". That would be incompatible with allowing users to use the padding bits.

That is, `repr(C)` unions are **not** and cannot be just "bags of bits" where one could write to any bit, and that bit value would need to be preserved on copy / move / pass-by-value.

We should document this for `repr(C)` unions in the Unsafe Code Guidelines, so I'm re-opening this issue until that is resolved.

BurntSushi:
I think you meant to ping @comex here. :-) (Not comes.)

RalfJung:
> That is, repr(C) unions are not and cannot be just "bags of bits" where one could write to any bit, and that bit value would need to be preserved on copy / move / pass-by-value.

I do not understand why this is a consequence. Your summary (and thanks for the writeup!) says that "Rust and GCC pass the bottom 32-bits". So it is possible to implement this in a way that preserves all bits. Correct?

[...]

PR #160, opened 2019-07-08
Detail layout of repr(C) unions

gnzlbg:
Documents the behavior described in #156 .

### DIFF
--- a/reference/src/layout/unions.md
+++ b/reference/src/layout/unions.md
@@ -18,16 +18,6 @@ is already entirely determined by their types, and since we intend to allow
 creating references to fields (`&u.f1`), unions do not have any wiggle-room
 there.
 
-### C-compatible layout ("repr C")
-
-For unions tagged `#[repr(C)]`, the compiler will apply the C layout scheme.  Per
-sections [6.5.8.5] and [6.7.2.1.16] of the C11 specification, this means that
-the offset of every field is 0.  Unsafe code can cast a pointer to the union to
-a field type to obtain a pointer to any field, and vice versa.
-
-[6.5.8.5]: http://port70.net/~nsz/c/c11/n1570.html#6.5.8p5
-[6.7.2.1.16]: http://port70.net/~nsz/c/c11/n1570.html#6.7.2.1p16
-
 ### Default layout ("repr rust")
 
 **The default layout of unions is not specified.** As of this writing, we want
@@ -38,3 +28,44 @@ contents are.
 Even if the offsets happen to be all 0, there might still be differences in the
 function call ABI.  If you need to pass unions by-value across an FFI boundary,
 you have to use `#[repr(C)]`.
+
+### Layout of `repr(C)` unions
+
+The layout of `repr(C)` unions follows the C layout scheme. Per sections
+[6.5.8.5] and [6.7.2.1.16] of the C11 specification, this means that the offset
+of every field is 0. Unsafe code can cast a pointer to the union to a field type
+to obtain a pointer to any field, and vice versa. 
+
+[6.5.8.5]: http://port70.net/~nsz/c/c11/n1570.html#6.5.8p5
+[6.7.2.1.16]: http://port70.net/~nsz/c/c11/n1570.html#6.7.2.1p16
+
+Since all fields are at offset 0, this implies that `repr(C)` unions do not have
+padding before or in-between their fields. They can, however, have trailing
+padding (see next example).
+
+Union fields of zero-size participate in the layout computation of the union. For example:
+
+```rust
+# use std::mem::{size_of, align_of};
+#[repr(C)] 
+union U {
+  x: u8,
+  y: [u16; 0],
+}
+# fn main() {
+// The zero-sized type [u16; 0] raises the alignment requirement to 2
+assert_eq!(align_of::<U>(), 2);
+// This introduces trailing padding, raising the union size to 2
+assert_eq!(size_of::<U>(), 2);
+# }
+```
+
+> **NOTE**: U is larger than its largest field, and has therefore 1 byte of
+> trailing padding.
+
+This handling of zero-sized types is equivalent to the handling of zero-sized
+types in struct fields, and matches the behavior of GCC and Clang for unions in
+C when zero-sized types are allowed via their language extensions.
+
+The bit `i` of a `repr(C)` union is a padding bit if the bit `i` of each of its
+fields is a padding bit or trailing padding.
--- a/reference/src/layout/unions.md
+++ b/reference/src/layout/unions.md
@@ -57,8 +57,9 @@ assert_eq!(size_of::<U>(), 2);
 # }
 ``**
 
-**Note**: Fields are overlapped instead of laid out sequentially, so unlike structs there is no "between the fields" that could be filled with padding.
-only have trailing padding.
+> **Note**: Fields are overlapped instead of laid out sequentially, so 
+> unlike structs there is no "between the fields" that could be filled 
+> with padding.
 
 The bit `i` of a `repr(C)` union is a padding bit if the bit `i` of each of its
 fields is a padding bit or trailing padding. That is:
--- a/reference/src/layout/unions.md
+++ b/reference/src/layout/unions.md
@@ -64,7 +64,8 @@ assert_eq!(size_of::<U>(), 2);
 #### Zero-sized fields
 
 If a `#[repr(C)]` union contains a field of zero-size, that field does not
-occupy space in Rust unions (as opposed to, e.g., in C++). For example:
+occupy space in Rust unions, e.g., as opposed to _empty_ types in C++ unions,
+which might occupy space depending on how they are used. For example:
 
 ```rust
 # use std::mem::{size_of, align_of};
--- a/reference/src/layout/unions.md
+++ b/reference/src/layout/unions.md
@@ -64,8 +64,7 @@ assert_eq!(size_of::<U>(), 2);
 #### Zero-sized fields
 
 If a `#[repr(C)]` union contains a field of zero-size, that field does not
-occupy space in Rust unions, e.g., as opposed to _empty_ types in C++ unions,
-which might occupy space depending on how they are used. For example:
+occupy space in Rust unions. For example:
 
 ```rust
 # use std::mem::{size_of, align_of};
--- a/reference/src/layout/unions.md
+++ b/reference/src/layout/unions.md
@@ -61,16 +61,6 @@ assert_eq!(size_of::<U>(), 2);
 > unlike structs there is no "between the fields" that could be filled 
 > with padding.
 
-The bit `i` of a `repr(C)` union is a padding bit if the bit `i` of each of its
-fields is a padding bit or trailing padding. That is:
-
-```rust
-#[repr(C)]
-union U { x: (u8, u16) }
-```
-
-The byte at offset 1 of `U` is a padding byte.
-
 #### Zero-sized fields
 
 If a `#[repr(C)]` union contains a field of zero-size, that field does not
--- a/reference/src/layout/unions.md
+++ b/reference/src/layout/unions.md
@@ -64,7 +64,7 @@ assert_eq!(size_of::<U>(), 2);
 #### Zero-sized fields
 
 If a `#[repr(C)]` union contains a field of zero-size, that field does not
-occupy space in the union. For example:
+occupy space in Rust unions (as opposed to, e.g., in C++). For example:
 
 ```rust
 # use std::mem::{size_of, align_of};
--- a/reference/src/layout/unions.md
+++ b/reference/src/layout/unions.md
@@ -87,7 +87,9 @@ assert_eq!(size_of::<U>(), 1);
 # }
 ```
 
-The field does, however, participate in the layout computation of the union, and can raise its alignment requirement, which in turn can introduce trailing padding. For example:
+The field does, however, participate in the layout computation of the union, and
+can raise its alignment requirement, which in turn can introduce trailing
+padding. For example:
 
 ```rust
 # use std::mem::{size_of, align_of};
@@ -102,7 +104,7 @@ assert_eq!(align_of::<U>(), 2);
 // This introduces trailing padding, raising the union size to 2
 assert_eq!(size_of::<U>(), 2);
 # }
-``**
+```
 
 This handling of zero-sized types is equivalent to the handling of zero-sized
 types in struct fields, and matches the behavior of GCC and Clang for unions in
--- a/reference/src/layout/unions.md
+++ b/reference/src/layout/unions.md
@@ -87,8 +87,7 @@ assert_eq!(size_of::<U>(), 1);
 # }
 ```
 
-The field does, however, participate in the layout computation of the union. For
-example:
+The field does, however, participate in the layout computation of the union, and can raise its alignment requirement, which in turn can introduce trailing padding. For example:
 
 ```rust
 # use std::mem::{size_of, align_of};
@@ -103,8 +102,12 @@ assert_eq!(align_of::<U>(), 2);
 // This introduces trailing padding, raising the union size to 2
 assert_eq!(size_of::<U>(), 2);
 # }
-```
+``**
 
 This handling of zero-sized types is equivalent to the handling of zero-sized
 types in struct fields, and matches the behavior of GCC and Clang for unions in
 C when zero-sized types are allowed via their language extensions.
+
+**C++ compatibility hazard**: C++ does, in general, give a size of 1 to empty
+structs. If an empty struct in C++ is used as an union field, a "naive"
+translation of that code into Rust will not produce a compatible result.
--- a/reference/src/layout/unions.md
+++ b/reference/src/layout/unions.md
@@ -39,11 +39,56 @@ to obtain a pointer to any field, and vice versa.
 [6.5.8.5]: http://port70.net/~nsz/c/c11/n1570.html#6.5.8p5
 [6.7.2.1.16]: http://port70.net/~nsz/c/c11/n1570.html#6.7.2.1p16
 
-Since all fields are at offset 0, this implies that `repr(C)` unions do not have
-padding before or in-between their fields. They can, however, have trailing
-padding (see next example).
+#### Padding
 
-Union fields of zero-size participate in the layout computation of the union. For example:
+Since all fields are at offset 0, `repr(C)` unions do not have padding before
+their fields. They can, however, have trailing padding, to make sure the size is
+a multiple of the alignment:
+
+```rust
+# use std::mem::{size_of, align_of};
+#[repr(C, align(2))]
+union U { x: u8 }
+# fn main() {
+// The repr(align) attribute raises the alignment requirement of U to 2
+assert_eq!(align_of::<U>(), 2);
+// This introduces trailing padding, raising the union size to 2
+assert_eq!(size_of::<U>(), 2);
+# }
+``**
+
+**Note**: there is no room between fields for padding, so `repr(C)` unions can
+only have trailing padding.
+
+The bit `i` of a `repr(C)` union is a padding bit if the bit `i` of each of its
+fields is a padding bit or trailing padding. That is:
+
+```rust
+#[repr(C)]
+union U { x: (u8, u16) }
+```
+
+The byte at offset 1 of `U` is a padding byte.
+
+#### Zero-sized fields
+
+If a `#[repr(C)]` union contains a field of zero-size, that field does not
+occupy space in the union. For example:
+
+```rust
+# use std::mem::{size_of, align_of};
+#[repr(C)] 
+union U {
+  x: u8,
+  y: (),
+}
+# fn main() {
+assert_eq!(size_of::<U>(), 1);
+# }
+```
+
+The field does, however, participate in the layout computation of the union. For
+example:
 
 ```rust
 # use std::mem::{size_of, align_of};
@@ -60,12 +105,6 @@ assert_eq!(size_of::<U>(), 2);
 # }
 ```
 
-> **NOTE**: U is larger than its largest field, and has therefore 1 byte of
-> trailing padding.
-
 This handling of zero-sized types is equivalent to the handling of zero-sized
 types in struct fields, and matches the behavior of GCC and Clang for unions in
 C when zero-sized types are allowed via their language extensions.
-
-The bit `i` of a `repr(C)` union is a padding bit if the bit `i` of each of its
-fields is a padding bit or trailing padding.
--- a/reference/src/layout/unions.md
+++ b/reference/src/layout/unions.md
@@ -55,7 +55,7 @@ assert_eq!(align_of::<U>(), 2);
 // This introduces trailing padding, raising the union size to 2
 assert_eq!(size_of::<U>(), 2);
 # }
-``**
+```
 
 > **Note**: Fields are overlapped instead of laid out sequentially, so 
 > unlike structs there is no "between the fields" that could be filled 
--- a/reference/src/layout/unions.md
+++ b/reference/src/layout/unions.md
@@ -57,7 +57,7 @@ assert_eq!(size_of::<U>(), 2);
 # }
 ``**
 
-**Note**: there is no room between fields for padding, so `repr(C)` unions can
+**Note**: Fields are overlapped instead of laid out sequentially, so unlike structs there is no "between the fields" that could be filled with padding.
 only have trailing padding.
 
 The bit `i` of a `repr(C)` union is a padding bit if the bit `i` of each of its
--- a/reference/src/layout/unions.md
+++ b/reference/src/layout/unions.md
@@ -85,6 +85,8 @@ assert_eq!(size_of::<U>(), 2);
 # }
 ```
 
-**C++ compatibility hazard**: C++ does, in general, give a size of 1 to empty
-structs. If an empty struct in C++ is used as an union field, a "naive"
-translation of that code into Rust will not produce a compatible result.
+**C++ compatibility hazard**: C++ does, in general, give a size of 1 to types
+with no fields. When such types are used as an union field in C++, a "naive"
+translation of that code into Rust will not produce a compatible result. Refer
+to the [struct chapter](structs-and-tuples.md#c-compatible-layout-repr-c) for
+further details.
--- a/reference/src/layout/unions.md
+++ b/reference/src/layout/unions.md
@@ -63,24 +63,12 @@ assert_eq!(size_of::<U>(), 2);
 
 #### Zero-sized fields
 
-If a `#[repr(C)]` union contains a field of zero-size, that field does not
-occupy space in Rust unions. For example:
+`repr(C)` union fields of zero-size are handled in the same way as in struct
+fields, matching the behavior of GCC and Clang for unions in C when zero-sized
+types are allowed via their language extensions.
 
-```rust
-# use std::mem::{size_of, align_of};
-#[repr(C)] 
-union U {
-  x: u8,
-  y: (),
-}
-# fn main() {
-assert_eq!(size_of::<U>(), 1);
-# }
-```
-
-The field does, however, participate in the layout computation of the union, and
-can raise its alignment requirement, which in turn can introduce trailing
-padding. For example:
+That is, these fields occupy zero-size and participate in the layout computation
+of the union as usual:
 
 ```rust
 # use std::mem::{size_of, align_of};
@@ -96,10 +84,6 @@ assert_eq!(align_of::<U>(), 2);
 assert_eq!(size_of::<U>(), 2);
 # }
 ```
-
-This handling of zero-sized types is equivalent to the handling of zero-sized
-types in struct fields, and matches the behavior of GCC and Clang for unions in
-C when zero-sized types are allowed via their language extensions.
 
 **C++ compatibility hazard**: C++ does, in general, give a size of 1 to empty
 structs. If an empty struct in C++ is used as an union field, a "naive"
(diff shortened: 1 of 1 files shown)


reference/src/layout/unions.md
```diff
@@ -38,3 +28,44 @@ contents are.
...
+[6.5.8.5] and [6.7.2.1.16] of the C11 specification, this means that the offset
+of every field is 0. Unsafe code can cast a pointer to the union to a field type
+to obtain a pointer to any field, and vice versa. 
+
+[6.5.8.5]: http://port70.net/~nsz/c/c11/n1570.html#6.5.8p5
+[6.7.2.1.16]: http://port70.net/~nsz/c/c11/n1570.html#6.7.2.1p16
+
+Since all fields are at offset 0, this implies that `repr(C)` unions do not have
+padding before or in-between their fields. They can, however, have trailing
+padding (see next example).
```
RalfJung: Padding isn't really part of the layout, it is part of the "value representation". So I am not sure if this fits here.

reference/src/layout/unions.md
```diff
@@ -38,3 +28,44 @@ contents are.
...
+
+Union fields of zero-size participate in the layout computation of the union. For example:
```
RalfJung: This is just a consequence of prior statements, not a new normative clause, right? It should be marked as that.

gnzlbg: I don't think we have stated anything about repr(C) union fields of zero size. C doesn't have ZSTs, and in the section about struct layout we do specifically call out that when computing the layout of a `repr(C)` struct, if a field has zero-size, it participates in the layout computation in a specific way.

reference/src/layout/unions.md
gnzlbg: I'm not sure either. Instead of using the word "padding" I can just say that a `union` can be larger than its largest field, and let users interpret from there that unions can have trailing padding.

reference/src/layout/unions.md
RalfJung: There's a 3-line note there. That seems appropriate.

reference/src/layout/unions.md
```diff
@@ -38,3 +28,44 @@ contents are.
 Even if the offsets happen to be all 0, there might still be differences in the
 function call ABI.  If you need to pass unions by-value across an FFI boundary,
 you have to use `#[repr(C)]`.
+
+### Layout of `repr(C)` unions
```
RalfJung: Should be "repr C" for consistency.

Or change the other section titles here and in the other documents as well.

reference/src/layout/unions.md
RalfJung: Actually the struct chapter also talks about padding. So I think I am fine mentioning it.

However, the first sentence is weird to me. Union fields are not "next to" each other, so "between" as a notion does not even make sense.

gnzlbg: > However, the first sentence is weird to me. Union fields are not "next to" each other, so "between" as a notion does not even make sense.

Isn't that what the sentence says? Because all fields start at offset 0, there is no space before or in-between the fields that could hold padding.

RalfJung: The text talks about "in-between union fields". I think that makes the text semantically ill-formed.

I'd just remove the "or in-between", so it says

> Since all fields are at offset 0, this implies that `repr(C)` unions do not have
padding their fields.

gnzlbg: Well, repr(Rust) unions do not have to have all fields at offset 0, so they could have padding between fields.

RalfJung: No, they can only have padding *before* their fields. Also the paragraph is (the way I read it) only about `repr(C)` unions.

gnzlbg: > No, they can only have padding before their fields.

Where do we guarantee this ?

reference/src/layout/unions.md
gnzlbg: So the note in the struct layout applies to union layout as well ? That feels to implicit to me, but I can live with that.

reference/src/layout/unions.md
RalfJung: The concept "between" makes no sense, so how would they have padding there...?

"between" makes sense of structs because they put their fields into an *order* and then lay them out *next to each other*. No such thing happens for unions.

Asking for the padding between union fields IMO is like asking for the color of a union field. It's just a nonsense question. For every union field, there can be padding before it (the size of this padding equals the offset of the field) and padding after it (the size of this padding equals the size of the union minus the size of the field), and that's it.

reference/src/layout/unions.md
RalfJung: No sorry that's not what I meant.

What I meant is, we can have a 3-line note here as well. But it seems entirely overkill to spend a section on this.

gnzlbg: Ok, that sounds good.

gnzlbg: @RalfJung the example shows trailing padding, so it cannot be removed. I can remove this one line, and remove the one line after the example, replacing them with a note similar to struct layout, but... then this section ends up being longer, which I'm not sure is what you meant.

reference/src/layout/unions.md
gnzlbg: This snippet guarantees that, for `repr(C)` unions, all union fields have the same offset, and explicitly mentions the corollary that there cannot be any padding between union fields due to this guarantee. I think this is worth mentioning, and am open to better suggestions to explain this that do not use the word "between". If you disagree that this is worth mentioning, please argue why. 

The argument that "this is obvious"/"asking for padding is nonsense" are IMO weak, because unions do have padding, and because to the best of my knowledge we don't guarantee anywhere that all union fields have the same offset.

RalfJung: I strongly feel that talking about padding "between" union fields is very confusing. I literally cannot parse that sentence, it does not have a meaning IMO.

Unions have padding, but to have something "between" two things you need to have two things *next to each other* such that there is a concept of "between". That condition is just not met here.

reference/src/layout/unions.md
RalfJung: > the example shows trailing padding, so it cannot be removed

Wait, are you saying that ZST with alignment are the only way trailing padding can happen in a union? Or why is trailing padding not demonstrated with an example focused on trailing padding (the current example mixes both trailing padding and ZST involvement, it is not a good idea to demonstrate two things in one example)?

RalfJung: Also it would help to have a structure like https://github.com/rust-lang/unsafe-code-guidelines/blob/master/reference/src/layout/structs-and-tuples.md#c-compatible-layout-repr-c, with "paragraph-level" sub headers.

reference/src/layout/unions.md
RalfJung: I do not think anything is omitted by the sentence I proposed above:

> **Padding.** Since all fields are at offset 0, `repr(C)` unions do not have
padding before their fields. They can, however, have trailing
padding, to make sure the size is a multiple of the alignment (see next example).

---

One thing I am wondering: in a union like
```rust
#[repr(C)]
union { f1: u8, f2: u16 }
```
I feel like it makes sense to consider the first *field* as having trailing padding. Basically, all union fields are padded to the union's size. Is "padding" a good term here? Does this concept have another name?

reference/src/layout/unions.md
gnzlbg: > Wait, are you saying that ZST with alignment are the only way trailing padding can happen in a union?

Nope, i'm just saying that the example currently demonstrates it with ZST, and the example is the bulk of this section. Removing / rephrasing the ZSTs bits, does not really make the section any shorter, which seemed to be the concern being raised:

> But it seems entirely overkill to spend a section on this.

RalfJung: "this" = ZST.

I did not even realize that the example is doing both of these things. So, as a first step, can you split the section into two parts: `**Padding.**` and `**Zero-sized fields**.`? The first should then have an example for trailing padding does not use a zero-sized field.

reference/src/layout/unions.md
gnzlbg: > I feel like it makes sense to consider the first field as having trailing padding.
> Basically, all union fields are padded to the union's size. Is "padding" a good term here? Does this concept have another name?

I think this is a good point, and I don't know if this makes sense. The first thing that jumps to mind is that the "trailing padding of a field" isn't really "padding", at least not in the same sense that the compiler isn't required to preserve its contents on move, e.g.,

```rust
#[repr(C)] union U { f1: u8, f2: u16 }
let u = U { f2: 0xff_ff };
let ptr = &u.f1 as *const _;
// Here we aren't really accessing padding:
assert_eq!(*ptr.add(1), 0xff);
```

> I strongly feel that talking about padding "between" union fields is very confusing. I literally cannot parse that sentence, it does not have a meaning IMO.
>
>Unions have padding, but to have something "between" two things you need to have two things next to each other such that there is a concept of "between". That condition is just not met here.

FWIW this is what the text attempts to explain. There is no "between" the fields, and therefore there cannot be any padding there. I am open to suggestions about better ways to put this, but I do think that it is important to call out that there is no "between" `repr(C)` union fields.

reference/src/layout/unions.md
gnzlbg: I'll try doing that.

reference/src/layout/unions.md
RalfJung: > I think this is a good point, and I don't know if this makes sense. The first thing that jumps to mind is that the "trailing padding of a field" isn't really "padding", at least not in the same sense as for struct fields, where the compiler isn't required to preserve its contents on move

From other discussions, however, it looks like even though unions themselves can only have leading and trailing padding, but no interior padding, there can *still* be "gaps" in what is preserved by the union based on padding of the fields.

So what about describing the layout of a union as follows:

* Fields are laid out "on top of each other", which is *the* defining feature of a union compared to a struct where fields are laid out "next to each other". So we determine the size of the union (details unspecified), and then we just have to make all fields have that size by adding appropriate padding before and/or after the field. For `repr(C)`, there is no padding before the field.

    If only I could draw... in my head I have two pictures here: for a struct, if we imagine the fields as being blocks, we have to build something that looks like
    ```
    [  <--> [field 1] <--> [field 2] <--> [field 3] <--> ]
    ```
    where the `<-->` are like springs or spacers or separators -- the sizes (and alignment) of the `[field N]` blocks is fixed, the only freedom we have is to pick an order and pick the size of the gaps. These gaps are the padding.

    In contrast, for a union, this looks more like
    ```
    [ <--> [field 1] <--> ]
    [ <--> [field 2] <--> ]
    [ <--> [field 3] <--> ]
    ```
    with the extra constraint that the three "lines" here must all have the same width. So this time, the only freedom we have is to pick the padding before and after every field. See how there is no "between" that we can even talk about?

* Eventually, once we define what "the padding of a struct" and "the padding of a union" is, we can then say that a byte is a padding byte for a union if it is a padding byte for all fields. This covers both (a) trailing padding that was added to all fields, and (b) interior padding of union members that happens to "align" for all fields.

Does this make sense?

reference/src/layout/unions.md
gnzlbg: I've split the section into two subsections, one talking about padding, and the other one showing that zero-sized fields do not take up any size, but do still participate in the layout computation of the union.

reference/src/layout/unions.md
gnzlbg: That makes sense to me, and I do think that would be a good way to specify things.

reference/src/layout/unions.md
```diff
@@ -39,11 +39,56 @@ to obtain a pointer to any field, and vice versa.
 [6.5.8.5]: http://port70.net/~nsz/c/c11/n1570.html#6.5.8p5
 [6.7.2.1.16]: http://port70.net/~nsz/c/c11/n1570.html#6.7.2.1p16
 
-Since all fields are at offset 0, this implies that `repr(C)` unions do not have
-padding before or in-between their fields. They can, however, have trailing
-padding (see next example).
+#### Padding
```
RalfJung: The precedent with `struct` I think is to use `**Paragraph Headers**`, not `#### Subsection headers`. I think that's better, the others become hard to distinguish when they get deeply nested.

reference/src/layout/unions.md
RalfJung: Feel free to use my "picture(s)". ;)

reference/src/layout/unions.md
```diff
@@ -39,11 +39,56 @@ to obtain a pointer to any field, and vice versa.
...
+  x: u8,
+  y: (),
+}
+# fn main() {
+assert_eq!(size_of::<U>(), 1);
+# }
+```
+
+The field does, however, participate in the layout computation of the union. For
+example:
```
RalfJung: Explain *how* it affects the computation.

reference/src/layout/unions.md
```diff
@@ -60,12 +105,6 @@ assert_eq!(size_of::<U>(), 2);
 # }
 ```
 
-> **NOTE**: U is larger than its largest field, and has therefore 1 byte of
-> trailing padding.
-
 This handling of zero-sized types is equivalent to the handling of zero-sized
 types in struct fields, and matches the behavior of GCC and Clang for unions in
 C when zero-sized types are allowed via their language extensions.
```
RalfJung: For structs we called out a C++ compat hazard, does the same apply here?

reference/src/layout/unions.md
RalfJung: Actually looking at the rendered version, this looks fine.

What remains is a concern about consistency.

reference/src/layout/unions.md
gnzlbg: Yes the same applies here. I'll call that out as well.

reference/src/layout/unions.md
gnzlbg: I prefer to use `###`  for headings on markdown since those allow for easy linking, and semantically express a new heading or sub-section. If the rendered book makes those hard to distinguish, we could tweak its CSS, or report the issue to mdbook upstream. I'd rather open an issue to track this editorial change.

reference/src/layout/unions.md
gnzlbg: I think that doing that is out of scope for this PR. We can open a new PR that uses the pictures for unions and structs. It something that's worth explaining at the top-level of the documents IMO.

RalfJung: Fine for me -- but do you still think my proposed wording doesn't work here if we don't have the picture?

reference/src/layout/unions.md
RalfJung: Looking at https://rust-lang.github.io/unsafe-code-guidelines/layout/structs-and-tuples.html, `#### header` and `**Header**` actually have pretty much the same size.

So, fine for me. We should probably just adjust the other documents as well, eventually.

reference/src/layout/unions.md
```diff
@@ -38,3 +28,83 @@ contents are.
...
+
+The bit `i` of a `repr(C)` union is a padding bit if the bit `i` of each of its
+fields is a padding bit or trailing padding. That is:
+
+```rust
+#[repr(C)]
+union U { x: (u8, u16) }
+```
+
+The byte at offset 1 of `U` is a padding byte.
```
RalfJung: I think the bit starting at "The bit `i`" and ending here should be dropped. This defines what "the padding bytes of a union" are, but we do not even have an agreed-upon definition for "the padding bytes of a type", of which this would have to be an instance.

reference/src/layout/unions.md
```diff
@@ -38,3 +28,83 @@ contents are.
...
+union U { x: u8 }
+# fn main() {
+// The repr(align) attribute raises the alignment requirement of U to 2
+assert_eq!(align_of::<U>(), 2);
+// This introduces trailing padding, raising the union size to 2
+assert_eq!(size_of::<U>(), 2);
+# }
+``**
+
+**Note**: there is no room between fields for padding, so `repr(C)` unions can
```
RalfJung:
```suggestion
**Note**: Fields are overlapped instead of laid out sequentially, so unlike structs there is no "between the fields" that could be filled with padding.
```

reference/src/layout/unions.md
gnzlbg: I think it would work, but we would need to put all of it in the `repr(C)` union section only, instead of being able to use it for unions at the top level, because we don't currently guarantee anywhere (for whatever reason) that all union fields are always at the same offset. 

Most of the wording should apply to all unions anyways, so I think it would make more sense to introduce it at the top level, and only extend that with whatever is necessary in this particular section. 

Before we do that, we probably want to have consensus that we want to guarantee that all union fields for all union types start always at the same offset, since that will simplify some things. And you might want to propose that wording in the validity discussion about unions, since I think this is something that could settle the padding issue there. It's essentially the wording that others have also proposed, it would fix the `repr(transparent)` issue, making `repr(rust)` and `repr(C)` unions more consistent, etc.

RalfJung: Nothing in my picture or my model has a problem with different offsets for different fields.

And I think the only reason why `repr(Rust)` does not fix offset 0 is to be able to have different offsets. For example, we unions also become subject to some amount of layout optimizations, the compiler might want to be clever here:
```rust
union Foo { b: bool, t: (u8, bool) }
```
If `b` gets offset 1 and `t` offset 0, the two `bool` align and the union could have a niche.

reference/src/layout/unions.md
gnzlbg: We finally get to the main change being made here 😆 

In this definition, "padding" is meant in the "C" language sense. AFAICT, we don't have much freedom here. Once we have a better definition of what padding is somewhere else, we should try to reconcile that with this one. But for `repr(C)` unions, interior field padding is not copied on move / copy / .. of the union, and in C one would say that this interior field padding is padding of the union itself.

reference/src/layout/unions.md
gnzlbg: That's the only reason I can imagine as well, but AFAICT those optimizations would require a field of the union to always be valid.

> Nothing in my picture or my model has a problem with different offsets for different fields.

Do we guarantee an upper bound in the size of the union, or can the fields be laid out such that there is actual padding between fields. E.g. in your `Foo` union example, `b` at offset 0, and `t` at offset 3, such that the union does have 3 actually bytes of padding between fields.

reference/src/layout/unions.md
RalfJung: I don't know what "padding in the C language sense" is or means.

My guess is you mean "not preserved on copies". I think that's out of scope for this document -- this is about *layout*, not *value representation*, and it is the latter that determines things like what gets preserved on a copy.

reference/src/layout/unions.md
RalfJung: > Do we guarantee an upper bound in the size of the union

No, not yet.

But there is still no padding between the fields. Your union looks like:
```rust
[ [b] <-- 7 bytes of padding -----------------------------> ]
[ <-- 3 bytes of padding --> [t] <-- 3 bytes of padding --> ]
```
See? No "between".

reference/src/layout/unions.md
gnzlbg: Makes sense. I'll remove this, but we do document the value representation of some types in the layout section (e.g. that a bool is 0 or 1). We need to discuss the value representation of unions at some point.

RalfJung: > but we do document the value representation of some types in the layout section (e.g. that a bool is 0 or 1)

Fair. But for bool it is very simple. ;)

>  We need to discuss the value representation of unions at some point.

Absolutely.

reference/src/layout/unions.md
gnzlbg: Are you proposing that padding before and after a "field" is part of the field itself ? Because as worded before, it did not appear that way. 

So I guess what I'm missing is, what do you think a union field should be ? For a field of type `T`, just the `size_of::<T>()` bits ? or `size_of::<Union>()` bits which includes the bits of `T` ?

reference/src/layout/unions.md
```diff
@@ -38,3 +28,79 @@ contents are.
...
+# use std::mem::{size_of, align_of};
+#[repr(C, align(2))]
+union U { x: u8 }
+# fn main() {
+// The repr(align) attribute raises the alignment requirement of U to 2
+assert_eq!(align_of::<U>(), 2);
+// This introduces trailing padding, raising the union size to 2
+assert_eq!(size_of::<U>(), 2);
+# }
+``**
```
RalfJung: There's another spurious `**` here (and GH's suggestion system doesn't let me suggest ```` ``` ```` ;)

reference/src/layout/unions.md
RalfJung: No not really. I am just thinking of there being padding around every union field, as you can see in my pictures. But I think it might make sense to have something equivalent to the "variant" of an enum, where one already picked a field but did not yet "descend down into it", if that makes sense.

Notice that nowhere in my picture is there a space between two fields. Union fields provide alternative "views" onto memory. To use a union you always first pick a field, and then view the entire union according to "that field's view". If you subscribe to that view, there is no "between", not even in your example. (I sound like a broken record, don't I?).

RalfJung: I guess you could say that padding is "between" two fields if some byte is in the trailing padding of one "variant" and the leading padding of another "variant". But IMO that is a complex concept, and more importantly it is *derived* from the primitive input that actually goes into union layout -- the leading and trailing padding for every "variant".

gnzlbg: > (I sound like a broken record, don't I?).

Not at all.

> But I think it might make sense to have something equivalent to the "variant" of an enum, where one already picked a field but did not yet "descend down into it", if that makes sense.

I think that makes sense. The padding is then in the variant, and not in the field, right ? There is definitely no padding across variants there, and because only one field is active on access, then there is only one field available each time you access the union, so it doesn't make sense to talk about padding between fields, at least not for access.

We then have bytes that aren't part of the value representation of any of the union fields. Whether they are padding before or after a field, interior field padding, etc. does not matter.

gnzlbg: >  but you only ever use one of these views at any given time.

Is this true ? I could pass two `&Union`  to two different threads, and each of them could access a different variant of the union simultaneously. So we might need something more precise here.

RalfJung: > Is this true ? I could pass two &Union to two different threads, and each of them could access a different variant of the union simultaneously. So we might need something more precise here.

I do not think this is an issue. Multiple threads only interact on the level of untyped memory accesses ("go get me the 4 bytes starting at `0xABC00`"). The "view" is used to (a) determine which bytes to ask from memory, and (b) interpret the bytes that came back.

So there is no issue at all with thread A using view 1 and thread B using view 2, concurrently. No party involved has to consider more than one view/"variant" at a time.

reference/src/layout/unions.md
```diff
@@ -38,3 +28,79 @@ contents are.
 Even if the offsets happen to be all 0, there might still be differences in the
 function call ABI.  If you need to pass unions by-value across an FFI boundary,
 you have to use `#[repr(C)]`.
+
+### Layout of "repr C" unions
```
RalfJung:
```suggestion
### C-compatible layout ("repr C")
```
Consistency with the struct chapter.

reference/src/layout/unions.md
```diff
@@ -38,3 +28,79 @@ contents are.
...
+```
+
+> **Note**: Fields are overlapped instead of laid out sequentially, so 
+> unlike structs there is no "between the fields" that could be filled 
+> with padding.
+
+#### Zero-sized fields
+
+If a `#[repr(C)]` union contains a field of zero-size, that field does not
+occupy space in the union. For example:
```
RalfJung: Is it just me, or does this sentence effectively say "a field of size zero has size 0", and then even goes on with an example to demonstrate? That seems... silly. What am I missing?

You seem dead set on having some example about zero-sized fields in this section. Okay, fine. But why *two*, with one being... utterly trivial?

gnzlbg: > What am I missing?

Do we specify anywhere in the document what the size of the field or the union containing the field is?

We consider these types FFI-safe, so from there it can be derived that this has to have some well defined rules. AFAICT we only say "this behaves like C" but C does not have any zero-sized types, so that does not cover this use case. 

C implementations often give these types size 0, such that the union has size 0, but C++ for example gives these fields size 1, such that the union ends up having size 1. 

What does Rust do here for unions and where is that written down ?

RalfJung: There's nothing special about size 0 though in the way to do layout. With something like https://github.com/rust-lang/unsafe-code-guidelines/issues/185, this falls out of a general-principle. We should explain the general principle of `repr(C)` union layout, not list each case (size 0, size 1, size 5, ...) separately.

You already added a remark covering the fact that this differs from what C++ does.

gnzlbg: > We should explain the general principle of repr(C) union layout, not list each case (size 0, size 1, size 5, ...) separately.

Sure, and part of this general principle is that `repr(C)` unions do not error when a zero-sized field is used, but that instead it actually treats them as a zero-sized field.

RalfJung: > Sure, and part of this general principle is that repr(C) unions do not error when a zero-sized field is used, but that instead it actually treats them as a zero-sized field.

Right, so we should describe the general thing and then maybe just say "(this includes fields of size 0)", and that should be all there is specific to that size in the repr(Rust) and repr(C) sections. And then at the bottom of the page, something like: "what we do matches what GCC and clang do for unions in C, but note that in C++, structs with no fields size 1 but also behave a bit like zero-sized types in Rust; hence there are compatibility issues".

This PR heavily implies like zero-sized fields are somehow the exception from some rule *when considering Rust and C FFI*. They are not an exception though, they work like any other size. That makes it very confusing to treat them as an exception. They are an exception only in C++, there is no reason that exception has to taint anything but the compatibility remarks specific to that language.

gnzlbg: Check the latest changes out.

reference/src/layout/unions.md
```diff
@@ -38,3 +28,63 @@ contents are.
...
+> unlike structs there is no "between the fields" that could be filled 
+> with padding.
+
+#### Zero-sized fields
+
+`repr(C)` union fields of zero-size are handled in the same way as in struct
+fields, matching the behavior of GCC and Clang for unions in C when zero-sized
+types are allowed via their language extensions.
+
+That is, these fields occupy zero-size and participate in the layout computation
```
gnzlbg: This still contains the tautological "occupy zero-size" remark, but since the zero-sized field increses the size of the union below, I think it is worth remarking here that the size of the field is indeed zero, and that the size increase is because they participate in the layout computation as usual.

RalfJung: Yeah this is much better. I still think this is overly specific and we should just draw my "picture" but at least I don't cringe any more when reading it. ;)

reference/src/layout/unions.md
```diff
@@ -38,3 +28,63 @@ contents are.
...
+}
+# fn main() {
+// The zero-sized type [u16; 0] raises the alignment requirement to 2
+assert_eq!(align_of::<U>(), 2);
+// This introduces trailing padding, raising the union size to 2
+assert_eq!(size_of::<U>(), 2);
+# }
+```
+
+**C++ compatibility hazard**: C++ does, in general, give a size of 1 to empty
```
RalfJung: Should we maybe say "struct with no field"? "empty" in Rust means something else than in C++.

reference/src/layout/unions.md
```diff
@@ -38,3 +28,63 @@ contents are.
...
+# use std::mem::{size_of, align_of};
+#[repr(C)] 
+union U {
+  x: u8,
+  y: [u16; 0],
+}
+# fn main() {
+// The zero-sized type [u16; 0] raises the alignment requirement to 2
+assert_eq!(align_of::<U>(), 2);
+// This introduces trailing padding, raising the union size to 2
```
RalfJung:
```suggestion
// This in turn introduces trailing padding, raising the union size to 2
```

reference/src/layout/unions.md
comex: Can we add a link somewhere to the section about zero-sized types in `structs-and-tuples.md`?  That section goes into a lot more detail, e.g.
- explicitly stating that empty structs are illegal in C
- mentioning `[[no_unique_address]]`
- differentiating between structs with no fields and structs with fields of zero size
There's no need to repeat all that here, but a link would be helpful.

comex: (Perhaps also mention that Rust unions can't have zero fields.)

gnzlbg: The text now uses "struct with no fields" and refers and links to the structs chapter for details.

reference/src/layout/unions.md
gnzlbg: Yes, I think that once we start adding pictures, and Pad, we can maybe explain this better.

reference/src/layout/unions.md
RalfJung: You resolved this but didn't actually change anything?

RalfJung: Oh wait, GH just showed an old diff for some reason. Sorry.

reference/src/layout/unions.md
RalfJung: Some of your commits seem to have been lost? Like, " Link to struct chapter" just is not in this PR any more.

gnzlbg: That might not have been intended. I merged the "suggestion changes" and that had conflicts so I had to force push :/

RalfJung: Force-push shouldn't be needed after merging...? Weird.

RalfJung:
Other than this final issue about empty structs, I have one more remark. I felt it is eaiser to just open a PR against your PR for this: https://github.com/gnzlbg/unsafe-code-guidelines/pull/1

reference/src/layout/unions.md
gnzlbg: I screw up. I need to learn how to "pull" changes to my branch locally and rebase them without spurious merge commits when my local branch has other changes :/

PR #161, opened 2019-07-08
Guarantee the layout of some unions having a single non-zero-sized field

gnzlbg:
This PR guarantees that unions having a single non-zero-sized field have the same layout as that field if:

* that field has no padding,
* all zero-sized fields have an alignment requirement of 1.

Without the requirement that the field might have no padding, the union would "inherit" the padding of their non-zero-sized field, such that these unions wouldn't be "bags of bits". 

We can relax this requirement later if we decide that unions shouldn't just be bags of bits, but for the time being, requiring no padding allows us to make this guarantee. 

Right now, we don't consider padding to be a part of layout. We should probably do so, since for unions padding doesn't really relate to field offsets - e.g., for repr(C) unions it results from the "overlapping" padding of their fields and trailing padding, see #160. 

An alternative way to word the no padding requirement here would be to guarantee this only for types with a "call ABI" equal to "Scalar" or "Vector", since those types have no padding. cc @eddyb

### DIFF
--- a/reference/src/layout/unions.md
+++ b/reference/src/layout/unions.md
@@ -8,12 +8,12 @@ not to change until an RFC ratifies them.
 
 ### Layout of individual union fields
 
-When laying out an `union`, the compiler has to decide how the fields of the
-union are arranged. Union fields are laid out by the compiler "on top" of each
-other, that is, the bytes of one field might overlap with the bytes of another
-field - as opposed to, e.g., `struct`s, where the fields are laid out "next to"
-each other, such that the bytes of one field never overlap with the bytes of
-another field. This can be visualized as follows:
+The main degree of freedom the compiler has when computing the layout of a union
+is to determine the offset of the fields. Union fields are laid out by the
+compiler "on top" of each other, that is, the bytes of one field might overlap
+with the bytes of another field - as opposed to, e.g., `struct`s, where the
+fields are laid out "next to" each other, such that the bytes of one field never
+overlap with the bytes of another field. This can be visualized as follows:
 
 ```rust,ignore
 [ P P [field0_ty] P P P P ]
--- a/reference/src/layout/unions.md
+++ b/reference/src/layout/unions.md
@@ -57,7 +57,7 @@ you have to use `#[repr(C)]`.
 
 #### Layout of unions with a single non-zero-sized field
 
-The layout of unions with a single single non-[1-ZST]-field" is the same as the
+The layout of unions with a single non-[1-ZST]-field" is the same as the
 layout of that field if it has no [padding] bytes.
 
 For example, here:
--- a/reference/src/layout/unions.md
+++ b/reference/src/layout/unions.md
@@ -17,7 +17,6 @@ overlap. This can be visualized as follows:
 [ <----> [field1_ty] <--> ]
 [ <---> [field2_ty] <---> ]
 ```
-
 **Figure 1** (union-field layout): Each row in the picture shows the layout of
 the union for each of its variants. The `<-...->` and `[ ... ]` denote the
 differently-sized gaps and fields, respectively.
--- a/reference/src/layout/unions.md
+++ b/reference/src/layout/unions.md
@@ -18,10 +18,10 @@ overlap. This can be visualized as follows:
 [ <---> [field2_ty] <---> ]
 ```
 
-> **Figure: union field layout**: Each row in the picture shows the layout of
-> the union for each of its fields, where the square brackets `[]` depict an
-> array of bytes, `<-...->` denotes different amount of [padding], and
-> `[field{i}_ty]` is the bytes of the type of the `i`-th union field.
+**Figure 1** (union-field layout): Each row in the picture shows the layout of
+the union for each of its variants. The `<-...->` and `[ ... ]` denote the
+differently-sized gaps and fields, respectively and the `[field{i}_ty]` are the
+bytes of the type of the `i`-th union field.
 
 The individual fields (`[field{i}_ty_]`) are blocks of fixed size determined by
 the field's [layout]. The only degrees of freedom the compiler has when
@@ -59,11 +59,8 @@ you have to use `#[repr(C)]`.
 
 #### Layout of unions with a single non-zero-sized field
 
-The layout of unions with a single non-zero-sized field is the same as the
-layout of that field if:
-
-* all zero-sized fields are [1-ZST], and
-* the non-zero sized field has no padding bits.
+The layout of unions with a single single non-[1-ZST]-field" is the same as the
+layout of that field if it has no [padding] bytes.
 
 For example, here:
 
--- a/reference/src/layout/unions.md
+++ b/reference/src/layout/unions.md
@@ -12,7 +12,7 @@ A union consists of several variants, one for each field. All variants have the
 same size and start at the same memory address, such that in memory the variants
 overlap. This can be visualized as follows:
 
-```rust,ignore
+```text
 [ <--> [field0_ty] <----> ]
 [ <----> [field1_ty] <--> ]
 [ <---> [field2_ty] <---> ]
@@ -20,16 +20,19 @@ overlap. This can be visualized as follows:
 
 > **Figure: union field layout**: Each row in the picture shows the layout of
 > the union for each of its fields, where the square brackets `[]` depict an
-> array of bytes, `<-...->` denotes different amount of padding, and
+> array of bytes, `<-...->` denotes different amount of [padding], and
 > `[field{i}_ty]` is the bytes of the type of the `i`-th union field.
 
 The individual fields (`[field{i}_ty_]`) are blocks of fixed size determined by
-the field's layout. The only degrees of freedom the compiler has when computing
-the layout of a union are the size of the union, which can be larger than the
-size of its largest field, and the offset of each union field within its
-variant. How these are picked depends on certain constraints, lik for example,
-the alignment requirements of the fields, the `#[repr]` attribute of the
-`union`, etc.
+the field's [layout]. The only degrees of freedom the compiler has when
+computing the layout of a union are the size of the union, which can be larger
+than the size of its largest field, and the offset of each union field within
+its variant. How these are picked depends on certain constraints like, for
+example, the alignment requirements of the fields, the `#[repr]` attribute of
+the `union`, etc.
+
+[padding]: ../glossary.md#padding
+[layout]: ../glossary.md#layout
 
 ### Unions with default layout ("`repr(Rust)`")
 
--- a/reference/src/layout/unions.md
+++ b/reference/src/layout/unions.md
@@ -22,12 +22,12 @@ the union for each of its variants. The `<-...->` and `[ ... ]` denote the
 differently-sized gaps and fields, respectively.
 
 The individual fields (`[field{i}_ty_]`) are blocks of fixed size determined by
-the field's [layout]. The only degrees of freedom the compiler has when
-computing the layout of a union are the size of the union, which can be larger
-than the size of its largest field, and the offset of each union field within
-its variant. How these are picked depends on certain constraints like, for
-example, the alignment requirements of the fields, the `#[repr]` attribute of
-the `union`, etc.
+the field's [layout]. Since we allow creating references to union fields
+(`&u.i`), the only degrees of freedom the compiler has when computing the layout
+of a union are the size of the union, which can be larger than the size of its
+largest field, and the offset of each union field within its variant. How these
+are picked depends on certain constraints like, for example, the alignment
+requirements of the fields, the `#[repr]` attribute of the `union`, etc.
 
 [padding]: ../glossary.md#padding
 [layout]: ../glossary.md#layout
@@ -45,13 +45,13 @@ whether all fields have the same offset, etc.
 As of this writing, we want to keep the option of using non-zero offsets open
 for the future; whether this is useful depends on what exactly the
 compiler-assumed invariants about union contents are. This might become clearer
-after the validity of unions
-([unsafe-code-guidelines/73](https://github.com/rust-lang/unsafe-code-guidelines/issues/73))
-is settled.
+after the validity of unions [#73] is settled.
 
 Even if the offsets happen to be all 0, there might still be differences in the
 function call ABI.  If you need to pass unions by-value across an FFI boundary,
 you have to use `#[repr(C)]`.
+
+[#73]: https://github.com/rust-lang/unsafe-code-guidelines/issues/73
 
 </details>
 
--- a/reference/src/layout/unions.md
+++ b/reference/src/layout/unions.md
@@ -20,15 +20,15 @@ situated, i.e., the compiler picks the gap (often called padding) before and
 after each field. This can be visualized as follows:
 
 ```rust,ignore
-[ P P [field0_ty] P P P P ]
-[ P P P P [field1_ty] P P ]
-[ P P P [field2_ty] P P P ]
+[ <--> [field0_ty] <----> ]
+[ <----> [field1_ty] <--> ]
+[ <---> [field2_ty] <---> ]
 ```
 
 > **Figure: union field layout**: Each row in the picture shows the layout of
 > the union for each of its fields, where the square brackets `[]` depict an
-> array of bytes. Here, `P` is a byte of type `Pad` and `[field{i}_ty]` is the
-> bytes of the type of the `i`-th union field.
+> array of bytes, `<-...->` denotes different amount of padding, and
+> `[field{i}_ty]` is the bytes of the type of the `i`-th union field.
 
 The individual fields (`[field{i}_ty_]`) are blocks of fixed size determined by
 the field's layout. The compiler picks the offset of the fields with respect to
--- a/reference/src/layout/unions.md
+++ b/reference/src/layout/unions.md
@@ -8,12 +8,16 @@ not to change until an RFC ratifies them.
 
 ### Layout of individual union fields
 
-The main degree of freedom the compiler has when computing the layout of a union
-is to determine the offset of the fields. Union fields are laid out by the
-compiler "on top" of each other, that is, the bytes of one field might overlap
-with the bytes of another field - as opposed to, e.g., `struct`s, where the
-fields are laid out "next to" each other, such that the bytes of one field never
-overlap with the bytes of another field. This can be visualized as follows:
+A union consists of several variants, one for each field. These variants are
+laid out "on top" of each other, so they must all have the same size. That is,
+the byte of each variant overlaps with the byte located at the same offset in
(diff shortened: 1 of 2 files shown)


RalfJung:
> all zero-sized fields have an alignment requirement of 1.

This is another place where we should use the "ZST1" terminology.

reference/src/layout/unions.md
```diff
@@ -6,17 +6,98 @@ not to change until an RFC ratifies them.
...
+you have to use `#[repr(C)]`.
+
+</details>
+
+#### Layout of unions with a single non-zero-sized field
+
+The layout of unions with a single non-zero-sized field is the same as the
+layout of that field if:
+
+* that field has no padding bits, and
```
RalfJung: We have not defined padding (what it is, how it behaves, where it occurs in which types) anywhere, and IMO it is not part of layout. So I feel like we are getting a little ahead of ourselves here.

Padding is IMO part of the value representation of a type; no interpretation of layout (size, align, niche, call ABI, field offsets) really includes "padding". For structs it naturally emerges as the "gaps between" fields, for enums I have no idea and for unions I am confused. ;)

RalfJung:
> Without the requirement that the field might have no padding, the union would "inherit" the padding of their non-zero-sized field, such that these unions wouldn't be "bags of bits".

We can say they have the same size+align, but niches and callABI might be different. Then we can avoid talking about padding, which I would prefer.

reference/src/layout/unions.md
gnzlbg: So we touched this on the meeting today. And it is not clear whether we have the definitions required to word this yet. We do not have a definition for padding, it is unclear whether a definition for padding is needed, but we could try to word this without using the "term" padding with the definitions that we have, and if we can't, we might be able to identify whatever definitions would be required to word something like this (and block this on working on those).

reference/src/layout/unions.md
```diff
@@ -6,17 +6,98 @@ not to change until an RFC ratifies them.
 
 [#13]: https://github.com/rust-rfcs/unsafe-code-guidelines/issues/13
 
-The only degree of freedom the compiler has when computing the layout of a union
-like
+### Layout of individual union fields
 
-```rust,ignore
-union U { f1: T1, f2: T2 }
+The layout of each union field is determined by its type.
```
RalfJung: I think this is basically https://github.com/rust-lang/unsafe-code-guidelines/issues/185 for unions.

RalfJung: By this I mean: I think the text here should use that picture. The rationale already explains why that picture is accurate.

"The layout of each union field is determined by its type" sounds like a tautology. The layout of all fields is already given when we compute the layout of a union.

gnzlbg:
So i've pushed a couple of commits that add pictures and make some progress. 

The issues with talking about "padding" in the guarantee about Rust unions with a single non-zero sized field are unfortunate. I think that the next step for this PR  would be to also write down what could we guarantee if the non-zero-sized field were to have padding bits (e.g. only size and alignment) and _why_ (i.e. write down the rationale) - this should help refresh everybody's memory about what the constraints are.

RalfJung:
Can we wait with this PR until after landing the other union layout one? They look like they will heavily conflict anyway.

gnzlbg:
Yes I'd prefer to merge the other one first and then rebase this one and see if it still makes sense in its current form.

RalfJung:
In fact I'd prefer to stage this: first just add "pictures" without saying more than we did before. That's basically an editorial change. Then try to say more.

The "picture" part should be roughly synced with https://github.com/rust-lang/unsafe-code-guidelines/pull/186.

reference/src/layout/unions.md
```diff
@@ -6,17 +6,109 @@ not to change until an RFC ratifies them.
 
 [#13]: https://github.com/rust-rfcs/unsafe-code-guidelines/issues/13
 
-The only degree of freedom the compiler has when computing the layout of a union
```
RalfJung: I like this "the only degree of freedom" formulation, could you keep it?

reference/src/layout/unions.md
```diff
@@ -6,17 +6,109 @@ not to change until an RFC ratifies them.
...
+other, that is, the bytes of one field might overlap with the bytes of another
+field - as opposed to, e.g., `struct`s, where the fields are laid out "next to"
+each other, such that the bytes of one field never overlap with the bytes of
+another field. This can be visualized as follows:
 
 ```rust,ignore
-union U { f1: T1, f2: T2 }
+[ P P [field0_ty] P P P P ]
+[ P P P P [field1_ty] P P ]
+[ P P P [field2_ty] P P P ]
```
RalfJung: We'll have to come to an agreement whether to use `P P P P P` or `<------->`. I prefer the arrows, they look like the "spacers" that come up when doing visual layout of components in a UI.

reference/src/layout/unions.md
RalfJung: However, the old sentence is not entirely correct.

I would write something like

"A union consists of several *variants*, one for each field. These variants are laid out "on top" of each other, so they must all have the same size. The only degree of freedom the compiler has when computing the layout of a union is to pick where in its variant each field is situated, i.e., the compiler picks the gap (often called padding) before and after each field."

reference/src/layout/unions.md
gnzlbg: Are the inner `[]` required ?

I think it would be simpler to just treat these are:

```rust
[ P P field0_ty P P P P ]
[ P P P P field1_ty P P ]
[ P P P field2_ty P P P ]
```

> I prefer the arrows, they look like the "spacers" that come up when doing visual layout of components in a UI.

That makes sense. The advantage I see about using `Pad` is that then the `[ ]` span a list of space separated types, and for each its layout is computed following the usual rules.

gnzlbg: If we were to use `()` and commas as separators, the rows would look like tuples:

```rust
( P, P, field0_ty, P, P, P, P )
( P, P, P, P, field1_ty, P, P )
( P, P, P, field2_ty, P, P, P )
```

I don't know if this makes it easier or harder to understand these figures. Either way we should have good captions / glossary for them.

reference/src/layout/unions.md
gnzlbg: Does the compiler only have one degree of freedom? I think it can, at least, choose: 

* offset of fields,
* amount of trailing padding (e.g. we don't guarantee an upper bound for this),

within certain constraints.

reference/src/layout/unions.md
RalfJung: > Are the inner [] required ?

I think it looks odd without them. The field should appear as a "block" somehow.

> If we were to use () and commas as separators, the rows would look like tuples:

With this it looks like every element of the tuple is equally sized. 
Really I want a picture of some wooden blocks with springs between them. I don't think commas make any sense there.

reference/src/layout/unions.md
RalfJung: Well maybe call it "degrees of freedom" then. What you are saying is all already contained in my proposed wording.

reference/src/layout/unions.md
```diff
@@ -8,12 +8,16 @@ not to change until an RFC ratifies them.
...
-with the bytes of another field - as opposed to, e.g., `struct`s, where the
-fields are laid out "next to" each other, such that the bytes of one field never
-overlap with the bytes of another field. This can be visualized as follows:
+A union consists of several variants, one for each field. These variants are
+laid out "on top" of each other, so they must all have the same size. That is,
+the byte of each variant overlaps with the byte located at the same offset in
+all other variants, and the bytes of one field might overlap with the bytes of
+another field - as opposed to, e.g., `struct`s, where the fields are laid out
+"next to" each other, such that the bytes of one field never overlap with the
+bytes of another field. The main degree of freedom the compiler has when
```
RalfJung: Wow, this is the most convoluted explanation of "overlap" I have ever seen. Why do you deem this necessary?

Also, if we say "the main degree of freedom" that leaves the question "what are the others". Are there any others?

reference/src/layout/unions.md
gnzlbg: Changed this to use <--->

reference/src/layout/unions.md
gnzlbg: I've summarized this further

gnzlbg:
So I've rebased and solved the open issues.

reference/src/layout/unions.md
```diff
@@ -6,29 +6,109 @@ not to change until an RFC ratifies them.
...
+> **Figure: union field layout**: Each row in the picture shows the layout of
+> the union for each of its fields, where the square brackets `[]` depict an
+> array of bytes, `<-...->` denotes different amount of padding, and
+> `[field{i}_ty]` is the bytes of the type of the `i`-th union field.
+
+The individual fields (`[field{i}_ty_]`) are blocks of fixed size determined by
+the field's layout. The only degrees of freedom the compiler has when computing
+the layout of a union are the size of the union, which can be larger than the
+size of its largest field, and the offset of each union field within its
+variant. How these are picked depends on certain constraints, lik for example,
```
RalfJung:
```suggestion
variant. How these are picked depends on certain constraints, like for example,
```

reference/src/layout/unions.md
```diff
@@ -6,29 +6,109 @@ not to change until an RFC ratifies them.
...
 ```
 
-is to determine the offset of the fields. The layout of these fields themselves
-is already entirely determined by their types, and since we intend to allow
-creating references to fields (`&u.f1`), unions do not have any wiggle-room
-there.
+> **Figure: union field layout**: Each row in the picture shows the layout of
+> the union for each of its fields, where the square brackets `[]` depict an
+> array of bytes, `<-...->` denotes different amount of padding, and
+> `[field{i}_ty]` is the bytes of the type of the `i`-th union field.
```
RalfJung: The figure caption here has different style than what you suggested in https://github.com/rust-lang/unsafe-code-guidelines/pull/186. I think I prefer the other style, could you adjust this here?

RalfJung:
```suggestion
> `[field{i}_ty]` are the bytes of the type of the `i`-th union field.
```

reference/src/layout/unions.md
```diff
@@ -6,29 +6,109 @@ not to change until an RFC ratifies them.
...
+[ <----> [field1_ty] <--> ]
+[ <---> [field2_ty] <---> ]
 ```
 
-is to determine the offset of the fields. The layout of these fields themselves
-is already entirely determined by their types, and since we intend to allow
-creating references to fields (`&u.f1`), unions do not have any wiggle-room
-there.
+> **Figure: union field layout**: Each row in the picture shows the layout of
+> the union for each of its fields, where the square brackets `[]` depict an
```
RalfJung: What are the square brackets here referring to? The `struct` figure has no similar explanation. I'd remove that part I think.

reference/src/layout/unions.md
```diff
@@ -6,29 +6,109 @@ not to change until an RFC ratifies them.
...
 
 Even if the offsets happen to be all 0, there might still be differences in the
 function call ABI.  If you need to pass unions by-value across an FFI boundary,
 you have to use `#[repr(C)]`.
 
+</details>
+
+#### Layout of unions with a single non-zero-sized field
+
+The layout of unions with a single non-zero-sized field is the same as the
```
RalfJung: What about saying "single non-1-ZST-field"? Then the first condition becomes redundant and this rule even becomes more general.

reference/src/layout/unions.md
```diff
@@ -6,29 +6,109 @@ not to change until an RFC ratifies them.
...
+layout of that field if:
+
+* all zero-sized fields are [1-ZST], and
+* the non-zero sized field has no padding bits.
```
RalfJung: Padding occurs byte-wise, right? So I think here and below this should be "padding bytes". Also please add a link to the glossary.

reference/src/layout/unions.md
```diff
@@ -18,10 +18,10 @@ overlap. This can be visualized as follows:
 [ <---> [field2_ty] <---> ]
 ```
```
RalfJung: Not sure if the empty line here makes a difference, but I think removing it makes the source look better by gluing the caption closer to the figure.

reference/src/layout/unions.md
```diff
@@ -6,29 +6,108 @@ not to change until an RFC ratifies them.
...
+The individual fields (`[field{i}_ty_]`) are blocks of fixed size determined by
+the field's [layout]. The only degrees of freedom the compiler has when
+computing the layout of a union are the size of the union, which can be larger
+than the size of its largest field, and the offset of each union field within
+its variant. How these are picked depends on certain constraints like, for
+example, the alignment requirements of the fields, the `#[repr]` attribute of
+the `union`, etc.
+
+[padding]: ../glossary.md#padding
+[layout]: ../glossary.md#layout
```
RalfJung: Shouldn't this be `glossary.html` to work in the rendered book?

reference/src/layout/unions.md
```diff
@@ -6,29 +6,108 @@ not to change until an RFC ratifies them.
...
 
 Even if the offsets happen to be all 0, there might still be differences in the
 function call ABI.  If you need to pass unions by-value across an FFI boundary,
 you have to use `#[repr(C)]`.
 
+</details>
+
+#### Layout of unions with a single non-zero-sized field
+
+The layout of unions with a single single non-[1-ZST]-field" is the same as the
```
RalfJung:
```suggestion
The layout of unions with a single non-[1-ZST]-field" is the same as the
```

reference/src/layout/unions.md
```diff
@@ -6,29 +6,108 @@ not to change until an RFC ratifies them.
...
+# assert_eq!(size_of::<U1>(), align_of::<Zst2>());
+# assert_eq!(align_of::<U1>(), align_of::<Zst2>());
+assert_eq!(align_of::<Zst2>(), 16);
+# }
+```
+
+the layout of `U1` is **unspecified** because:
+
+* `Zst2` is not a [1-ZST], and
+* `SomeOtherStruct` has an unspecified layout and could contain padding bits.
```
RalfJung:
```suggestion
* `SomeOtherStruct` has an unspecified layout and could contain padding bytes.
```

reference/src/layout/unions.md
gnzlbg: I don't think so. The links should be to the markdown files, and then whether you generate html, or latex, or something else will depend on the markdown backend used.

RalfJung: Does the book compiler fix those links? Certainly in the past we have used HTML link, e.g. [here](https://github.com/rust-lang/unsafe-code-guidelines/blame/2bda4cbb469df96044a2aae47840cb34cee5edc8/reference/src/layout/enums.md#L379).

RalfJung: OTOH, links actually do seem to get adjusted by HTML generation. So I agree with you, your links are right -- and the others are wrong.

gnzlbg: > Does the book compiler fix those links? 

Yes, it does.

> Certainly in the past we have used HTML link, e.g. here.

Damn, we shouldn't do that. I'll fix that in a different PR.

gnzlbg: I've opened: https://github.com/rust-lang-nursery/mdBook/issues/1019

reference/src/layout/unions.md
```diff
@@ -6,29 +6,107 @@ not to change until an RFC ratifies them.
...
+the union for each of its variants. The `<-...->` and `[ ... ]` denote the
+differently-sized gaps and fields, respectively.
+
+The individual fields (`[field{i}_ty_]`) are blocks of fixed size determined by
+the field's [layout]. The only degrees of freedom the compiler has when
+computing the layout of a union are the size of the union, which can be larger
+than the size of its largest field, and the offset of each union field within
+its variant. How these are picked depends on certain constraints like, for
+example, the alignment requirements of the fields, the `#[repr]` attribute of
+the `union`, etc.
```
RalfJung: For structs, my PR says something like

> The layout of these fields
themselves is already entirely determined by their types, and since we intend to
allow creating references to fields (`&s.f1`), structs do not have any
wiggle-room there.

Might be worth also adding a sentence here justifying why that is the only freedom the compiler has.

reference/src/layout/unions.md
```diff
@@ -6,29 +6,107 @@ not to change until an RFC ratifies them.
...
-to keep the option of using non-zero offsets open for the future; whether this
-is useful depends on what exactly the compiler-assumed invariants about union
-contents are.
+<details><summary><b>Rationale</b></summary>
+
+As of this writing, we want to keep the option of using non-zero offsets open
+for the future; whether this is useful depends on what exactly the
+compiler-assumed invariants about union contents are. This might become clearer
+after the validity of unions
+([unsafe-code-guidelines/73](https://github.com/rust-lang/unsafe-code-guidelines/issues/73))
```
RalfJung: Please factor the link as a markdown reference (`[unsafe-code-guidelines/73][union-validity]` or so).

reference/src/layout/unions.md
gnzlbg: I've squeezed that in a bit.

reference/src/layout/unions.md
```diff
@@ -6,29 +6,107 @@ not to change until an RFC ratifies them.
...
-**The default layout of unions is not specified.** As of this writing, we want
-to keep the option of using non-zero offsets open for the future; whether this
-is useful depends on what exactly the compiler-assumed invariants about union
-contents are.
+<details><summary><b>Rationale</b></summary>
+
+As of this writing, we want to keep the option of using non-zero offsets open
+for the future; whether this is useful depends on what exactly the
+compiler-assumed invariants about union contents are. This might become clearer
+after the validity of unions [#73] is settled.
```
RalfJung:
```suggestion
after the [validity of unions][#73] is settled.
```

reference/src/layout/unions.md
```diff
@@ -6,29 +6,107 @@ not to change until an RFC ratifies them.
...
-is to determine the offset of the fields. The layout of these fields themselves
-is already entirely determined by their types, and since we intend to allow
-creating references to fields (`&u.f1`), unions do not have any wiggle-room
-there.
+Except for the guarantees provided below for some specific cases, the default
+layout of Rust unions is, _in general_, **unspecified**.
 
-### Default layout ("repr rust")
+That is, there are no _general_ guarantees about the offset of the fields,
+whether all fields have the same offset, etc.
```
rkruppe: I'd like to see the call ABI aspect also be mentioned here, not just in the rationale.

gnzlbg: Like this?

```suggestion
whether all fields have the same offset, what the call ABI of the union is, etc.
```

rkruppe: LGTM

PR #163, opened 2019-07-08
Guarantee that some structs are zero sized

gnzlbg:
This addresses part of #37, by guaranteeing that structs have zero size if they don't contain any non-zero-sized field.

### DIFF
--- a/reference/src/layout/structs-and-tuples.md
+++ b/reference/src/layout/structs-and-tuples.md
@@ -124,10 +124,8 @@ x: u16, y: T }` where `T = u32` are not guaranteed to be identical.
 
 #### Zero-sized structs
 
-Structs with default layout (`repr(Rust)`), layout with increased alignment
-(`repr(align(N))`), packed layout (`repr(packed(N))`), or C-compatible layout
-(`repr(C)`) are zero-sized, if they contain no fields of non-zero size. That is,
-either the type has no fields, or all of its fields have zero size.
+For `repr(Rust)`, `repr(packed(N))`, `repr(align(N))`, and `repr(C)`
+structs: if all fields of a struct have size 0, then the struct has size 0.
 
 For example, all these types are zero-sized:
 
--- a/reference/src/layout/structs-and-tuples.md
+++ b/reference/src/layout/structs-and-tuples.md
@@ -128,6 +128,20 @@ Structs with default layout are zero-sized, if they contain no fields of
 non-zero size. That is, either the type has no fields, or all of its fields have
 zero size.
 
+For example, all these types are zero-sized:
+
+```rust
+# use std::mem::size_of;
+struct Zst0;
+struct Zst1(Zst0);
+struct Zst2(Zst1, Zst0);
+# fn main() {
+# assert_eq!(size_of::<Zst0>(), 0);
+# assert_eq!(size_of::<Zst1>(), 0);
+# assert_eq!(size_of::<Zst2>(), 0);
+# }
+```
+
 #### Unresolved questions
 
 During the course of the discussion in [#11] and [#12], various
--- a/reference/src/layout/structs-and-tuples.md
+++ b/reference/src/layout/structs-and-tuples.md
@@ -124,16 +124,17 @@ x: u16, y: T }` where `T = u32` are not guaranteed to be identical.
 
 #### Zero-sized structs
 
-Structs with default layout or default layout with increased alignment are
-zero-sized, if they contain no fields of non-zero size. That is, either the type
-has no fields, or all of its fields have zero size.
+Structs with default layout (`repr(Rust)`), layout with increased alignment
+(`repr(align(N))`), packed layout (`repr(packed(N))`), or C-compatible layout
+(`repr(C)`) are zero-sized, if they contain no fields of non-zero size. That is,
+either the type has no fields, or all of its fields have zero size.
 
 For example, all these types are zero-sized:
 
 ```rust
 # use std::mem::size_of;
 #[repr(align(32))] struct Zst0;
-struct Zst1(Zst0);
+#[repr(C)] struct Zst1(Zst0);
 struct Zst2(Zst1, Zst0);
 # fn main() {
 # assert_eq!(size_of::<Zst0>(), 0);
--- a/reference/src/layout/structs-and-tuples.md
+++ b/reference/src/layout/structs-and-tuples.md
@@ -124,15 +124,15 @@ x: u16, y: T }` where `T = u32` are not guaranteed to be identical.
 
 #### Zero-sized structs
 
-Structs with default layout are zero-sized, if they contain no fields of
-non-zero size. That is, either the type has no fields, or all of its fields have
-zero size.
+Structs with default layout or default layout with increased alignment are
+zero-sized, if they contain no fields of non-zero size. That is, either the type
+has no fields, or all of its fields have zero size.
 
 For example, all these types are zero-sized:
 
 ```rust
 # use std::mem::size_of;
-struct Zst0;
+#[repr(align(32))] struct Zst0;
 struct Zst1(Zst0);
 struct Zst2(Zst1, Zst0);
 # fn main() {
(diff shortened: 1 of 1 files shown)


reference/src/layout/structs-and-tuples.md
```diff
@@ -122,6 +122,26 @@ compiler will not reorder it, to allow for the possibility of
...
+# use std::mem::size_of;
+struct Zst0;
+struct Zst1(Zst0);
+struct Zst2(Zst1, Zst0);
+# fn main() {
+# assert_eq!(size_of::<Zst0>(), 0);
+# assert_eq!(size_of::<Zst1>(), 0);
+# assert_eq!(size_of::<Zst2>(), 0);
+# }
+```
```
RalfJung: I am feeling like we are increasingly listing weird special cases here and will soon be lost in the woods of them, where really these all fall out from the same pattern. :/

This overlaps with https://github.com/rust-lang/unsafe-code-guidelines/pull/164. I think we should try to find a more general statement that subsumes both.

OTOH, this here also applies if the ZST have nontrivial alignment, right? If yes, there should be an example of that.

RalfJung:
Why are all these `size == 0`important enough to list them here, wasting precious space and brain resources for every reader?

RalfJung:
One way we could avoid all of that, at the cost of a bigger commitment, is by saying that `repr(Rust)` is the same as `repr(C)` after the compiler did some field reordering. That would save us a *lot* of text. Maybe we can make it a bit more relaxed to give ourselves some more freedom for future changes?

reference/src/layout/structs-and-tuples.md
```diff
@@ -122,6 +122,26 @@ compiler will not reorder it, to allow for the possibility of
 unsizing. E.g., `struct Foo { x: u16, y: u32 }` and `struct Foo<T> {
 x: u16, y: T }` where `T = u32` are not guaranteed to be identical.
 
+#### Zero-sized structs
+
+Structs with default layout are zero-sized, if they contain no fields of
```
RalfJung: This should explicitly say `repr(Rust)`

reference/src/layout/structs-and-tuples.md
gnzlbg: > where really these all fall out from the same pattern. :/

I'm open to suggestions. @rkruppe had the idea that we should document, e.g., all zero-sized related guarantees in some "central" place, so that one doesn't need to go through all layout documents looking for them. I agreed that this might make sense, once we have collected them all. At that point, the pattern might be obvious.

gnzlbg: > OTOH, this here also applies if the ZST have nontrivial alignment, right? If yes, there should be an example of that.

Yes, this applies for all alignments. I"ll modify the example to show that.

gnzlbg:
> Why are all these size == 0important enough to list them here, wasting precious space and brain resources for every reader?

Please, do suggest a better way to guarantee all these details, that does not have those downsides.

gnzlbg:
>  is by saying that repr(Rust) is the same as repr(C) after the compiler did some field reordering. 

It is unclear to me what the consequences of that are. This should deserve its own issues, where that can be explored. AFAICT we do guarantee that ZSTs in `repr(C)` structs have zero size, but we also do mention that they can introduce padding. It is unclear to me whether we do currently guarantee that a repr(C) struct containing only ZSTs (of possibly varying alignment) is zero-sized.

gnzlbg:
> In particular, while clang do guarantee that zero-sized fields in C structs have zero size (but maybe add padding), I am not sure whether they give structs containing only ZSTs a size of zero.

They do in C (https://gcc.godbolt.org/z/Tcu49v), they don't do it in C++ (https://gcc.godbolt.org/z/tn9Mod), but that difference already exist.

RalfJung:
> Please, do suggest a better way to guarantee all these details, that does not have those downsides.

A separate file would be a good start IMO.

gnzlbg:
Shall we then remove all mentions of zero-sized types from all layout files and move them to a separate file?

RalfJung:
Hm, that also seems odd somehow as that one file will talk about enums and structus and unions, right?

Maybe having a clearly separated section on "questions around ZST" in each of these files is better. I am not sure.

gnzlbg:
@RalfJung I agree with you in that we need to find a better solution, but I believe it will be easier to refactor the guarantees after we agree on what those should be first. It might well be that, due to the refactor, we end up guaranteeing a little bit more here and there due to making things more general.

I think that a section on each file is probably necessary to work out all of the details, but we also have a top-level layout document at the beginning of this chapter that is currently empty, where we could try to provide general rules that always apply.

Lokathor:
1) Place an entry for "Zero Sized Type (ZST)" in the Glossary. The definition can be very simple but particularly you want to introduce the acronym to people as part of the glossary entry.
2) Place the info for how a struct/union/enum ends up being zero sized as a section on the pages for each category of type.

gnzlbg:
It's unclear to me whether there is anything else to do with this PR.

The issue that needs resolving is whether this is something we want to guarantee or not, and it is unclear from the comments whether that is the case. If there is no consensus, we should close this PR and document in the issue that there was no consensus about guaranteeing this. 

Trying to coalesce these guarantees for easier consumption somewhere is IMO pointless if the guarantees themselves don't have consensus.

reference/src/layout/structs-and-tuples.md
```diff
@@ -122,6 +122,27 @@ compiler will not reorder it, to allow for the possibility of
 unsizing. E.g., `struct Foo { x: u16, y: u32 }` and `struct Foo<T> {
 x: u16, y: T }` where `T = u32` are not guaranteed to be identical.
 
+#### Zero-sized structs
+
+Structs with default layout (`repr(Rust)`), layout with increased alignment
+(`repr(align(N))`), packed layout (`repr(packed(N))`), or C-compatible layout
+(`repr(C)`) are zero-sized, if they contain no fields of non-zero size. That is,
```
RalfJung: Is this an "if and only if"?

gnzlbg: I think so. If a struct contains a field of non-zero size, I don't see how it could end up being zero-sized.

rkruppe: I can't think of a counter-example, but it seems like only one direction is much commonly thought about and discussed, so I'd be a bit scared to commit to the other direction as well. Besides, it's usually more important to know when something definitely is a ZST (e.g., so you can ignore it for layout related questions), not when something definitely isn't a ZST.

RalfJung: Fair, but but then we should try to word this in a way that it is clear that only one direction is defined. In particular with the way that conditional and conclusion are reversed here, I think it is easy to interpret this as bidirectional.

English is such a bad language for precise specifications...

rkruppe:
r=me as well modulo resolving Ralf's nit (one way or another), this was completely uncontroversial (unlike e.g., enums) and is clearly useful.

reference/src/layout/structs-and-tuples.md
RalfJung: So, to make a concrete proposal (also removing a pointless double-negation):

> If all fields of a struct have size 0, then the struct has size 0.

Your sentence lists a few representations explicitly, but is there any repr for which this does *not* make sense?

comex: Well, that statement is trivially true if there struct has no fields.  C++ allows structs with no fields but gives them a size of 1, and in theory there could be some kind of `#[repr(C++)]` that accounts for this.

I'm not sure how plausible `#[repr(C++)]` is, but another hypothetical use case could be to fix the [ABI issue with passing small C++ classes by vaue](https://github.com/rust-lang/rust-bindgen/issues/778).

gnzlbg: > but is there any repr for which this does not make sense?

If we ever add a `repr()` for which this does not make sense, then this wouldn't hold for  `repr(transparent)` either.

The current wording mentions all `repr`s that currently exist, except for `repr(transparent)` , for which this just follows from its rules.

RalfJung: If you really want to list the `repr`, I'd propose prepending something like:

> For `repr(Rust)`, `repr(packed(N))`, `repr(align(N))` and `repr(C)`:

(Btw, why is it `packed` but not `aligned`? Seems inconsistent...)

gnzlbg: I've reworded the guarantee with this feedback. Let me know if its ok now.

RalfJung:
Thanks! This sounds a bit clumsy, which I realize is my fault. But I think at least the meaning is clear. We can always improve the wording later.

PR #164, opened 2019-07-08
Guarantee the layout of structs with a single non-zero-sized field

### LINKED ISSUE #34: Layout of single-field structs
PR #31 identified this as an area for continued discussion. Specifically, if you have a struct with single field (`struct Foo { x: T }`), should we guarantee that the memory layout of `Foo` is identical to the memory layout of `T` (note that ABI details around function calls may still draw a distinction, which is why `#[repr(transparent)]` is needed). What about zero-sized types like `PhantomData`?

@rkruppe [wrote](https://github.com/rust-rfcs/unsafe-code-guidelines/pull/31#issuecomment-426730654):

> Long ago I proposed that we might want to guarantee (some subset of) newtype unpacking for repr(Rust) structs. @nikomatsakis carried this over into #11 as discussion point but it received no further discussion. I like to think that means it's uncontroversial 😄 I've also never heard of any reason why one might not want that to be true.
> 
> To make a specific proposal, let's restrict it to structs [1] that contain a single field having the same _memory layout_ as the type of the sole field. So `struct Foo<T>(T);` and `struct Foo<T> { x: T }` would be laid out like `T` in memory, though possibly still passed differently in function calls.
> 
> [1] The same guarantee for `(T,)` are already covered by the special case of homogeneous tuples being laid out like arrays that is already in this PR.

gnzlbg:
This PR guarantees the layout of structs with a single non-zero-sized field to be the same layout as that of that field if the alignment requirements of all other fields is 1.

Closes #34 .

### DIFF
--- a/reference/src/layout/structs-and-tuples.md
+++ b/reference/src/layout/structs-and-tuples.md
@@ -141,10 +141,10 @@ struct Zst2(Zst1, Zst0);
 # }
 ```
 
-#### Default layout of structs with a single non-zero-sized field
+#### Default layout of structs where only a single field is not a 1-ZST
 
-The default layout of structs with a single non-zero-sized field is the same as
-the layout of that field if all other fields are [1-ZST]
+The default layout of structs where only a single field is not a 1-ZST is the
+same as the layout of that non-1-ZST field.
 
 For example, the layout of:
 
--- a/reference/src/layout/structs-and-tuples.md
+++ b/reference/src/layout/structs-and-tuples.md
@@ -146,8 +146,10 @@ struct S3(Zst1); // same layout as Zst1
 
 #### Zero-sized structs
 
-For `repr(Rust)`, `repr(packed(N))`, `repr(align(N))`, and `repr(C)`
-structs: if all fields of a struct have size 0, then the struct has size 0.
+For `repr(Rust)`, `repr(packed(N))`, `repr(align(N))`, and `repr(C)` structs: if
+all fields of a struct have size 0, then the struct has size 0. In particular, a
+struct with no fields is a ZST, and if it has no repr attribute it is moreover a
+1-ZST as it also has no alignment requirements.
 
 For example, all these types are zero-sized:
 
--- a/reference/src/layout/structs-and-tuples.md
+++ b/reference/src/layout/structs-and-tuples.md
@@ -122,6 +122,7 @@ compiler will not reorder it, to allow for the possibility of
 unsizing. E.g., `struct Foo { x: u16, y: u32 }` and `struct Foo<T> {
 x: u16, y: T }` where `T = u32` are not guaranteed to be identical.
 
+<<<<<<< HEAD
 #### Zero-sized structs
 
 For `repr(Rust)`, `repr(packed(N))`, `repr(align(N))`, and `repr(C)`
@@ -141,6 +142,28 @@ struct Zst2(Zst1, Zst0);
 # }
 ```
 
+=======
+#### Default layout of structs with a single non-zero-sized field
+
+The default layout of structs with a single non-zero-sized field is the same as
+the layout of that field if the alignment requirement of all other fields is 1.
+
+For example, the layout of:
+
+```rust
+struct SomeStruct(i32, ());
+```
+
+is the same as the layout of `i32`, but the layout of:
+
+```rust
+#[repr(align(16))] struct Zst;
+struct SomeOtherStruct(i32, Zst);
+```
+
+is **unspecified**, since there is a zero-sized field in `SomeOtherStruct` with
+alignment greater than 1.
+
 #### Unresolved questions
 
 During the course of the discussion in [#11] and [#12], various
@@ -150,14 +173,14 @@ issue has been opened for further discussion on the repository. This
 section documents the questions and gives a few light details, but the
 reader is referred to the issues for further discussion.
 
-**Single-field structs ([#34]).** If you have a struct with single field
-(`struct Foo { x: T }`), should we guarantee that the memory layout of
-`Foo` is identical to the memory layout of `T` (note that ABI details
-around function calls may still draw a distinction, which is why
-`#[repr(transparent)]` is needed). What about zero-sized types like
-`PhantomData`?
+**Zero-sized structs ([#37]).** If you have a struct which --
+transitively -- contains no data of non-zero size, then the size of
+that struct will be zero as well. These zero-sized structs appear
+frequently as exceptions in other layout considerations (e.g.,
+single-field structs). An example of such a struct is
+`std::marker::PhantomData`.
 
-[#34]: https://github.com/rust-rfcs/unsafe-code-guidelines/issues/34
+[#37]: https://github.com/rust-rfcs/unsafe-code-guidelines/issues/37
 
 **Homogeneous structs ([#36]).** If you have homogeneous structs, where all
 the `N` fields are of a single type `T`, can we guarantee a mapping to
--- a/reference/src/layout/structs-and-tuples.md
+++ b/reference/src/layout/structs-and-tuples.md
@@ -124,10 +124,6 @@ compiler will not reorder it, to allow for the possibility of
 unsizing. E.g., `struct Foo { x: u16, y: u32 }` and `struct Foo<T> {
 x: u16, y: T }` where `T = u32` are not guaranteed to be identical.
 
-#### Structs with no fields
-
-Structs with default layout and no fields are [1-ZST].
-
 #### Structs with 1-ZST fields
 
 For the purposes of struct layout [1-ZST] fields are ignored.
--- a/reference/src/layout/structs-and-tuples.md
+++ b/reference/src/layout/structs-and-tuples.md
@@ -122,7 +122,6 @@ compiler will not reorder it, to allow for the possibility of
 unsizing. E.g., `struct Foo { x: u16, y: u32 }` and `struct Foo<T> {
 x: u16, y: T }` where `T = u32` are not guaranteed to be identical.
 
-<<<<<<< HEAD
 #### Zero-sized structs
 
 For `repr(Rust)`, `repr(packed(N))`, `repr(align(N))`, and `repr(C)`
@@ -142,11 +141,10 @@ struct Zst2(Zst1, Zst0);
 # }
 ```
 
-=======
 #### Default layout of structs with a single non-zero-sized field
 
 The default layout of structs with a single non-zero-sized field is the same as
-the layout of that field if the alignment requirement of all other fields is 1.
+the layout of that field if all other fields are [1-ZST]
 
 For example, the layout of:
 
@@ -160,9 +158,7 @@ is the same as the layout of `i32`, but the layout of:
 #[repr(align(16))] struct Zst;
 struct SomeOtherStruct(i32, Zst);
 ```
-
-is **unspecified**, since there is a zero-sized field in `SomeOtherStruct` with
-alignment greater than 1.
+is **unspecified**, since `Zst` is not a [1-ZST].
 
 #### Unresolved questions
 
@@ -423,3 +419,5 @@ proposal (and -- further -- it does not match our existing behavior):
   thread](https://github.com/rust-rfcs/unsafe-code-guidelines/pull/31#discussion_r224955817)).
 - Many people would prefer the name ordering to be chosen for
   "readability" and not optimal layout.
+
+[1-ZST]: ../glossary.md#zero-sized-type--zst
--- a/reference/src/layout/structs-and-tuples.md
+++ b/reference/src/layout/structs-and-tuples.md
@@ -141,24 +141,21 @@ struct Zst2(Zst1, Zst0);
 # }
 ```
 
-#### Default layout of structs where only a single field is not a 1-ZST
+#### Structs with 1-ZST fields
 
-The default layout of structs where only a single field is not a 1-ZST is the
-same as the layout of that non-1-ZST field.
+For the purposes of struct layout [1-ZST] fields are ignored.
 
-For example, the layout of:
-
-```rust
-struct SomeStruct(i32, ());
-```
-
-is the same as the layout of `i32`, but the layout of:
+For example, 
 
 ```rust
-#[repr(align(16))] struct Zst;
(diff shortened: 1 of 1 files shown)


reference/src/layout/structs-and-tuples.md
```diff
@@ -122,6 +122,27 @@ compiler will not reorder it, to allow for the possibility of
 unsizing. E.g., `struct Foo { x: u16, y: u32 }` and `struct Foo<T> {
 x: u16, y: T }` where `T = u32` are not guaranteed to be identical.
 
+#### Default layout of structs with a single non-zero-sized field
+
+The default layout of structs with a single non-zero-sized field is the same as
+the layout of that field if the alignment requirement of all other fields is 1.
```
RalfJung: This sentence puts part of the conditions on other fields at the beginning and parts at the end.

IMO we should, somewhere, define a term for a "zero-sized-one-aligned type". Working title ZSOAT. Then we can just say things like "structs with a single non-ZSOAT field".

gnzlbg: The glossary would be the right place to define that term. We should probably define ZST as zero-sized-type there as well. ZSOAT is also quite a mouthful, maybe just `ZST1` ?

Lokathor: With a numeral it makes it easy to talk about higher aligned zero sized types as well, `ZST4` for example

Lokathor:
Can we make this guarantee stronger? It seems like as long as all ZST values have an alignment less than or equal to the Non-ZST value that layout should be unaffected.

eg: `*mut T` and `*mut T, PhantomData<&mut T>` should fairly probably be definitely the same.

rkruppe:
`PhantomData<T>` has alignment 1 regardless of `T` so it's already covered.

reference/src/layout/structs-and-tuples.md
RalfJung: ZSOAT was meant as a proposal that is so bad that we clearly cannot use it. :)

`ZST1`/`ZST4` sounds good to me! The general `ZSTN` looks a bit funny though, `ZST-N` would be clearer, so maybe it should be `ZST-1` then?

RalfJung: The obvious alternative is `1-ZST`. "One-aligned zero-sized type" works better than "zero-sized type, 1-aligned".

gnzlbg: > The obvious alternative is 1-ZST. "One-aligned zero-sized type" works better than "zero-sized type, 1-aligned".

I like `1-ZST`.

gnzlbg: It is unclear whether this shall that be done in a different PR or here. Adding an abbreviation for these types feels like an orthogonal problem to me, and requires going through the whole UCGs to use the abbreviation appropriately.

gnzlbg:
> Can we make this guarantee stronger?

We could. This was not discussed in the issue, but I suppose that we can make it stronger as a follow up.

reference/src/layout/structs-and-tuples.md
RalfJung: I'd rather add the abbreviation before we land all the PRs you opened that talk about N-ZST. Then we'll have strictly fewer places to fix later.

gnzlbg:
@rkruppe @RalfJung what do you think about strengthening the guarantee to:

> The default layout of structs with a single non-zero-sized field is the same as
> the layout of that field if the alignment requirement of all other fields is less than or equal to
> the alignment requirement of the non-zero-sized field.

?

rkruppe:
I don't think we should innovate here (has this wording *ever* popped up *anywhere* before?), especially as @Lokathor's motivation for proposing it (PhantomData) doesn't need this stronger guarantee after all. The only additional things this strengthening covers are empty arrays and user-declared ZSTs with an explicit `repr(align(N))` neither of which seems important to me. It also deviates from the precedent set by `repr(transparent)` -- in a sense we're doing this to downgrade `repr(transparent)` from a very important opt-in to a guardrail and arguably shouldn't go further.

RalfJung:
Unblocking as the 1-ZST PR landed. Or was this blocked on something else?

gnzlbg:
So I've updated this to use the "1-ZST" terminology and noticed two things:

* the text is a bit harder to read (maybe there is a better way to use the "1-ZST" terminology?) - I also noticed this in #161 .

* we could guarantee the layout of

```rust
struct S([0; u8], [0; u8], [0; u16]);
```

to be the same as that of `[0; u 16]` by, instead of guaranteeing this for structs with one non-zero-sized field, guaranteeing those for structs that have one field that is not a 1-ZST. 

I am not sure whether that is worth doing, and tend to agree with @rkruppe that the simple guarantee is enough. Just food for thought. What isn't clear to me is how we could generalize such a guarantee in the spirit of what @Lokathor proposed [here](https://github.com/rust-lang/unsafe-code-guidelines/pull/164#issuecomment-511134453), but without mentioning that one type is a non-zero-sized type.

RalfJung:
> structs that have one field that is not a 1-ZST.

I think this is what we should do. It is most consistent. The current definition leaves a weird "gap".

reference/src/layout/structs-and-tuples.md
```diff
@@ -122,6 +122,7 @@ compiler will not reorder it, to allow for the possibility of
 unsizing. E.g., `struct Foo { x: u16, y: u32 }` and `struct Foo<T> {
 x: u16, y: T }` where `T = u32` are not guaranteed to be identical.
 
+<<<<<<< HEAD
```
RalfJung: Something went very wrong here. Please resolve the conflicts.

gnzlbg: I git pull from my github branch after merging some suggestions into my local branch which had new commits and modifications and everything broke :/

reference/src/layout/structs-and-tuples.md
```diff
@@ -141,6 +142,27 @@ struct Zst2(Zst1, Zst0);
 # }
 ```
 
+=======
+#### Default layout of structs with a single non-zero-sized field
+
+The default layout of structs with a single non-zero-sized field is the same as
+the layout of that field if all other fields are [1-ZST]
```
RalfJung:
```suggestion
the layout of that field if all other fields are [1-ZST].
```

RalfJung:
> the text is a bit harder to read (maybe there is a better way to use the "1-ZST" terminology?) - I also noticed this in #161 .

I think this might be due to the mismatch mentioned/"gap" in my previous post. Or do you think this is hard to read?

"The default layout of structs where only a single field is not a 1-ZST, is the same as the layout of said field".

reference/src/layout/structs-and-tuples.md
RalfJung: Yeah, one has to remember to pull after applying suggestions (and to push before). I made that mistake, too...

gnzlbg:
Done.

reference/src/layout/structs-and-tuples.md
```diff
@@ -150,14 +169,14 @@ issue has been opened for further discussion on the repository. This
...
-(`struct Foo { x: T }`), should we guarantee that the memory layout of
-`Foo` is identical to the memory layout of `T` (note that ABI details
-around function calls may still draw a distinction, which is why
-`#[repr(transparent)]` is needed). What about zero-sized types like
-`PhantomData`?
+**Zero-sized structs ([#37]).** If you have a struct which --
+transitively -- contains no data of non-zero size, then the size of
+that struct will be zero as well. These zero-sized structs appear
+frequently as exceptions in other layout considerations (e.g.,
+single-field structs). An example of such a struct is
```
RalfJung: Should this link to the new section you added?

RalfJung:
In fact, I personally would just go for the stronger statement that for struct layout, 1-ZST are ignored. I don't see any reason to guarantee/propose some instances of that without guaranteeing/proposing the general, simpler to explain and more useful, principle.

Both the current content of this PR and "structs consisting only of 1-ZST are 1-ZST2 then basically just fall out of that principle, and the additional orthogonal statemenets

* structs with no fields (and no `repr` annotation) are 1-ZST
* structs with 1 field (and no `repr` annotation) have the same layout as that field

reference/src/layout/structs-and-tuples.md
```diff
@@ -141,6 +141,22 @@ struct Zst2(Zst1, Zst0);
...
+
+```rust
+type Zst1 = ();
+struct S1(i32, Zst1);
+
+type Zst2 = [u16; 0];
+struct S2(Zst2, Zst1);
+```
+
+the layout of `S1` is the same as that of `i32` and the layout of `S2` as that of `Zst2`.
```
RalfJung: This is saying more than just "1-ZST fields are ignored".

You are also assuming here that single-field structs have layout identical to their only field. That is a separate statement, and should IMO get a separate section.

rkruppe:
That seems to go beyond what got consensus back when struct layout was discussed. For example, it would mean that all instantiations of this type (i.e., regardless of the unit) have the same layout:

```rust
struct Vec2<Unit> {
    x: f32,
    y: f32,
    _unit: PhantomData<Unit>
}
```

but as was discussed extensively, there is some desire to keep open the possibilities of randomized and profile-driven type layout decision, and none of the very small restrictions on that (single-field structs, single-variant enums, etc.) that were proposed without anyone objecting cover this.

Don't get me wrong, I am personally also very much in favor of layout ignoring 1-ZSTs universally (and many other things besides). But I don't think we should just "sneak in" something which was explicitly discussed and didn't get consensus.

reference/src/layout/structs-and-tuples.md
```diff
@@ -141,6 +141,22 @@ struct Zst2(Zst1, Zst0);
 # }
 ```
 
+#### Structs with 1-ZST fields
+
+For the purposes of struct layout [1-ZST] fields are ignored.
```
RalfJung: I just noticed there's a "Zero-sized structs" section just above. This has strong overlap now.

We *could* collapse that other section to just be about "structs with no fields". That nicely combines with this section for structs consisting only of 1-ZST. However, that would reduce what we say (e.g. about structs consisting only of ZST but some having non-trivial alignment requirements). Do we care? It would also simplify the documentation which is nice.

RalfJung:
>  For example, I think it would mean that all instantiations of this type (i.e., regardless of the unit) have the same layout:

I don't see how it would say that. Those are still all separate instances of a generic type. The 1-ZST is ignored, but the type parameter is not.

rkruppe:
I realized the same thing after posting and just finished editing it in 😅

RalfJung:
@rkruppe in another word, we don't guarantee that the following two have the same layout:
```rust
struct Vec2_1 {
    x: f32,
    y: f32,
}
struct Vec2_2 {
    x: f32,
    y: f32,
}
```
Likewise, we don't guarantee that all instance of these have the same layout:
```rust
struct Vec2Plus<T> {
    x: f32,
    y: f32,
    plus: *mut T,
}
```
No ZST are involved in any of this. Why would the ZST in your example make any difference?

The fact that `Vec2<A>` and `Vec2<B>` don't have to have the same layout is an instance of the same principle. We have nominal typing, and every different choice of parameters is a different "name".

reference/src/layout/structs-and-tuples.md
gnzlbg: I realized the overlap as well, but thought about:

```rust
type Zst2 = [u16; 0];
struct S(Zst2);
```

which has zero-size but is not covered by the "we ignore 1-ZST"s clause. 

I decided to keep them as separate subsections on purpose, at least until someone figures out a good way to combine them.

RalfJung: I'd personally prefer to cut the other section until we find a nice way to expand what we got. ;)

But I can see how others might have a different opinion here.

But at least, I think it is worth adding a note like "in particular, a struct with *no* fields is a ZST, and if it has no `repr` attribute it is moreover a 1-ZST as it also has no alignment requirements".

gnzlbg: I'd rather not loose anything that we have already guaranteed. Each guarantee takes too long to just throw them away for an editorial change. 

I'm all in from restructuring the text as much as time allows once its in, but those should be non-functional changes.

reference/src/layout/structs-and-tuples.md
```diff
@@ -122,10 +124,32 @@ compiler will not reorder it, to allow for the possibility of
 unsizing. E.g., `struct Foo { x: u16, y: u32 }` and `struct Foo<T> {
 x: u16, y: T }` where `T = u32` are not guaranteed to be identical.
 
+#### Structs with no fields
+
+Structs with default layout and no fields are [1-ZST].
```
RalfJung: Isn't this strictly redundant, subsumed by "Zero-sized structs"?

reference/src/layout/structs-and-tuples.md
RalfJung: Fair.

reference/src/layout/structs-and-tuples.md
gnzlbg: Indeed.

reference/src/layout/structs-and-tuples.md
RalfJung: (This comment is still open.)

RalfJung:
Opened a PR for my remaining comments: https://github.com/gnzlbg/unsafe-code-guidelines/pull/2

rkruppe:
I'm perpetually confused by what any given person means by "layout", didn't we at some point agree to stop using that term to prevent confusion? Assuming that in this diff "layout" = size, align, field offsets and padding, possibly niches but definitely not call ABI, the content seems fine. But I'm unsure whether, editorially speaking, we should already be avoiding "layout" in new paragraphs or if that's gonna be a separate clean-up PR that goes over all the documents.

RalfJung:
That would be https://github.com/rust-lang/unsafe-code-guidelines/pull/153. I think the plan (like in [the other PR](https://github.com/rust-lang/unsafe-code-guidelines/pull/159#discussion_r314648207)) was to land this PR with the vague wording so that we have some examples to work with when finalizing the terminology. It's much easier to do that when you have some examples.

gnzlbg:
What we mean by layout here will be precised in #153 .

Issue #168, opened 2019-07-17
Shared and Unique slices and ZST


Lokathor:
At what is currently the bottom of https://github.com/rust-lang/unsafe-code-guidelines/issues/93 there were some interesting questions raised, but it was offtopic to that issue so here's a new one.

**Big Question:** How exactly do ZST interact with slices?

There's obviously a lot of sub-questions involved here.

* The particular bit of discussion in the other issue is that a user wanted to concat two shared slices.
* The normal way to do this is to see if one slice starts at one byte past the end of the other, and then combine the lengths if so. (EDIT: important note for readers, you can't combine two slices that are part of separate allocations even if they're adjacent in memory, that's UB)
* However, the length of a shared slice of ZST is _almost_ arbitrary (exception: uninhabited types cannot have a shared slice length >0), so you don't need to even check for the pointer being exactly one byte past the end, you can skip straight to adding up the lengths and use either input's address as the output's address (if the inputs are of an uninhabited type both lengths will be 0 and 0+0==0, or someone did UB before you and it doesn't matter what you do).

So the immediate questions to ask ourselves are:

1) Do we all agree that this is valid reasoning?
2) Does this logic hold with unique slices?
3) Any other questions people want to raise about how slices and ZST interact?

RalfJung:
The way I understood it, the key question we debated was whether this is a safe function:
```rust
fn arbitrary_blowup<'a, T>(slice: &'a [T], length: usize) -> &'a [T] {
    assert!(core::mem::size_of::<T>() == 0);
    assert!(!slice.is_empty());
    unsafe {
        core::slice::from_raw_parts(slice.as_ptr(), length);
    }
}
```
By "safe" I mean "can be called by arbitrary safe code without causing UB".

I originally thought it was not, but now I am convinced it is.

@gnzlbg asks
> I’m not sure I follow, arbitrary blowup gets a reference to a slice, and
increases the size of the slice, returning a reference to it. AFAICT this
creates a reference to values that do not exist.

The thing is, `&slice[0]` and `&slice[1]` return values that are *identical in any aspect* when this is a ZST slice. The index you are passing just does not matter at all, except for the bounds check. So how could there possibly be unsafety from growing a non-empty slice to something bigger?

Lokathor:
Does the index also not matter for `&mut [ZST]`?

If it does matter, how does that work with `split_at_mut`, since wouldn't both sides of the split end up with the same base address anyway?

[...]

PR #170, opened 2019-07-19
Initial Glossary entry for Aliasing

Lokathor:
This is basically the same definition as the Rustonomicon has, but now we're re-stating it here with (I hope) improved clarity.

Of note, it's now written down somewhere visible that ZSTs never alias.

Also I made the whole Glossary alphabetical.

Note: This is not intended to be new guarantees at all, I don't think I wrote anything that can be taken that way, it's just a definition so that people can speak with common terms.

### DIFF
--- a/reference/src/glossary.md
+++ b/reference/src/glossary.md
@@ -14,7 +14,7 @@ If data immediately pointed to by a `*const T` or `&*const T` is mutated, that's
 *Interior mutability* refers to the ability to perform interior mutation without causing UB.
 All interior mutation in Rust has to happen inside an [`UnsafeCell`](https://doc.rust-lang.org/core/cell/struct.UnsafeCell.html), so all data structures that have interior mutability must (directly or indirectly) use `UnsafeCell` for this purpose.
 
-#### Validity and safety invariant
+#### Validity Invariant
 
 The *validity invariant* is an invariant that all data must uphold any time it is accessed or copied in a typed manner.
 This invariant is known to the compiler and exploited by optimizations such as improved enum layout or eliding in-bounds checks.
@@ -30,6 +30,8 @@ fn main() { unsafe {
   let t: T = std::mem::transmute(TERM);
 } }
 ```
+
+#### Safety Invariant
 
 The *safety* invariant is an invariant that safe code may assume all data to uphold.
 This invariant is used to justify which operations safe code can perform.
--- a/reference/src/glossary.md
+++ b/reference/src/glossary.md
@@ -23,8 +23,8 @@ fn main() {
 }
 ```
 
-In this case, both `r` and `s` alias each other, since they both point to the
-memory of `u`.
+In this case, both `r` and `s` alias each other, since they both point to all of
+the bytes of `u`.
 
 However, `head` and `tail` do not alias each other: `head` points to the first
 byte of `u` and `tail` points to the other seven bytes of `u` after it.
--- a/reference/src/glossary.md
+++ b/reference/src/glossary.md
@@ -3,7 +3,7 @@
 #### Aliasing
 
 (Please note: a full aliasing model for Rust has not yet been constructed, but
-at the moment we can give the following guidelines.)
+at the moment we can give the following definition.)
 
 *Aliasing* is any time one pointer or reference points to a "span" of memory
 that overlaps with the span of another pointer or reference. A span of memory is
--- a/reference/src/glossary.md
+++ b/reference/src/glossary.md
@@ -3,7 +3,9 @@
 #### Aliasing
 
 (Please note: a full aliasing model for Rust has not yet been constructed, but
-at the moment we can give the following definition.)
+at the moment we can give the following definition. The most developed potential
+aliasing model so far is known as "Stacked Borrows", and can be found
+[here](https://github.com/Lokathor/unsafe-code-guidelines/blob/lokathor/wip/stacked-borrows.md).)
 
 *Aliasing* is any time one pointer or reference points to a "span" of memory
 that overlaps with the span of another pointer or reference. A span of memory is
--- a/reference/src/glossary.md
+++ b/reference/src/glossary.md
@@ -5,7 +5,7 @@
 (Please note: a full aliasing model for Rust has not yet been constructed, but
 at the moment we can give the following definition. The most developed potential
 aliasing model so far is known as "Stacked Borrows", and can be found
-[here](https://github.com/Lokathor/unsafe-code-guidelines/blob/lokathor/wip/stacked-borrows.md).)
+[here](https://github.com/rust-lang/unsafe-code-guidelines/blob/master/wip/stacked-borrows.md).)
 
 *Aliasing* is any time one pointer or reference points to a "span" of memory
 that overlaps with the span of another pointer or reference. A span of memory is
--- a/reference/src/glossary.md
+++ b/reference/src/glossary.md
@@ -52,15 +52,22 @@ If data immediately pointed to by a `*const T` or `&*const T` is mutated, that's
 *Interior mutability* refers to the ability to perform interior mutation without causing UB.
 All interior mutation in Rust has to happen inside an [`UnsafeCell`](https://doc.rust-lang.org/core/cell/struct.UnsafeCell.html), so all data structures that have interior mutability must (directly or indirectly) use `UnsafeCell` for this purpose.
 
-#### Layout
+#### Validity and safety invariant
 
-The *layout* of a type defines its size and alignment as well as the offsets of its subobjects (e.g. fields of structs/unions/enum/... or elements of arrays).
-Moreover, the layout of a type records its *function call ABI* (or just *ABI* for short): how the type is passed *by value* across a function boundary.
+The *validity invariant* is an invariant that all data must uphold any time it is accessed or copied in a typed manner.
+This invariant is known to the compiler and exploited by optimizations such as improved enum layout or eliding in-bounds checks.
 
-Note: Originally, *layout* and *representation* were treated as synonyms, and Rust language features like the `#[repr]` attribute reflect this. 
-In this document, *layout* and *representation* are not synonyms.
+In terms of MIR statements, "accessed or copied" means whenever an assignment statement is executed.
+That statement has a type (LHS and RHS must have the same type), and the data being assigned must be valid at that type.
+Moreover, arguments passed to a function must be valid at the type given in the callee signature, and the return value of a function must be valid at the type given in the caller signature.
+OPEN QUESTION: Are there more cases where data must be valid?
 
-#### Safety Invariant
+In terms of code, some data computed by `TERM` is valid at type `T` if and only if the following program does not have UB:
+```rust,ignore
+fn main() { unsafe {
+  let t: T = std::mem::transmute(TERM);
+} }
+```
 
 The *safety* invariant is an invariant that safe code may assume all data to uphold.
 This invariant is used to justify which operations safe code can perform.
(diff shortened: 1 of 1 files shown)


RalfJung:
> Also I made the whole Glossary alphabetical.

Could you make that a separate PR, for ease of reviewing?

reference/src/glossary.md
```diff
@@ -1,5 +1,43 @@
...
 
+#### Aliasing
+
+(Please note: a full aliasing model for Rust has not yet been constructed, but
+at the moment we can give the following definition.)
+
+*Aliasing* is any time one pointer or reference points to a "span" of memory
+that overlaps with the span of another pointer or reference. A span of memory is
+similar to how a slice works: there's a base byte address as well as a length in
+bytes.
```
RalfJung: Thanks a lot! I like it.

There is one possible source of confusion here though: LLVM's `noalias` actually uses a very different definition. That one is about the *accesses* that happen through a pointer and whether they *conflict* in the sense that the ordering is relevant. So two pointers that point to immutable memory never alias each other because two read accesses never "conflict".

But we probably want the "standard" notion of aliasing here for Rust, it works better with our usual statement that "shared references may alias". But it might be worth having a note somewhere pointing out that LLVM uses a different definition?

Lokathor:
Alright

* cleanup deferred for now
* Noted that `noalias` is an LLVM IR implementation detail that has a different idea of what aliasing is, but that it's a detail of LLVM-Rust, not really a properly of the Rust Abstract Machine.

reference/src/glossary.md
```diff
@@ -1,5 +1,48 @@
...
+    };
+    let (head, tail) = s.split_first().unwrap();
+}
+```
+
+In this case, both `r` and `s` alias each other, since they both point to all of
+the bytes of `u`.
+
+However, `head` and `tail` do not alias each other: `head` points to the first
+byte of `u` and `tail` points to the other seven bytes of `u` after it.
```
gnzlbg: I think it is worth stating the relationship between `head`/`tail` and `s` here, e.g., do `head` and `tail` alias `s` ?

For example, given:

```rust
    let mut x = 32_i32;
    let y = &mut x;
    let z = y as *mut _;
    unsafe {
        *z = 3;
    }
    *y = 42;
```

When does `z` alias `y` and whether that is UB ? I suppose that without an aliasing model, it might not be possible to tell what's exactly UB, but it would be worth it to write down what precisely is unspecified because of the lack of an aliasing model and showing a small example.

Lokathor: This particular example seems similar to a thing @RalfJung has explained before... I _think_ that `z` counts as a further borrow on `y`, so... something.

I mean it's clear that `y` and `z` point to the same memory span. It's not clear what that means in terms of aliasing/ownership rules.

RalfJung: > I think it is worth stating the relationship between head/tail and s here, e.g., do head and tail alias s ?

They do, and that is worth stating.

However, the point of this definition is to define when aliasing *happens*, not when it is *allowed*.

Lokathor: added

RalfJung: Thanks!

@gnzlbg do you agree with this? I don't think we should even try to define "allowed vs disallowed aliasing" in the glossary at this point.

At best we could link to `wip/stacked-borrows.md`.

gnzlbg: > I don't think we should even try to define "allowed vs disallowed aliasing" in the glossary at this point.

Ok, sounds good. I guess what confused me is that at the beginning an "aliasing model" is mentioned, and it wasn't fully clear to me how far this aliasing definition should go.

I think linking to the stacked-borrows wip document as an example of one potential aliasing model would help.

Lokathor: alright, can do after work

reference/src/glossary.md
```diff
@@ -1,5 +1,49 @@
...
+the bytes of `u`.
+
+However, `head` and `tail` do not alias each other: `head` points to the first
+byte of `u` and `tail` points to the other seven bytes of `u` after it. Also,
+both `head` and `tail` alias `s`.
+
+* The span length of `&T`, `&mut T`, `*const T`, or `*mut T` when `T` is
+  [`Sized`](https://doc.rust-lang.org/core/marker/trait.Sized.html) is
+  `size_of<T>()`.
+* When `T` is not `Sized` the span length is `size_of_val(t)`.
```
danielhenrymantilla: - _cannot find `t` in this scope_ :P

  - also, **how can we get the size of the pointee of a raw pointer?** (`mem::size_of_val` takes a `&T`).

      - this point may deserve its own discussion

Lokathor: yeah, it's own discussion, because that's a very tricky one, that T-libs has to handle.

RalfJung: T-lang, rather. But for now I think we can just assume the thing is "forced" to reference type. `&*t` would be extra-explicit I guess.

Lokathor: pointers are never automatically forced to references (very unsafe), so you'd have to use the `&*t` form in an unsafe block

RalfJung: I know, but this is pseudo-Rust anyway. ;)

gnzlbg: One could use type ascription here maybe, to denote that the type of `t` is `T`, e.g.,  `size_of_val(t: T)`

Lokathor: So with `size_of_val` it takes a `&T`, but you wouldn't write `size_of_val(&t)` when `t` is an unsized type since `t` is _already_ in reference form.

reference/src/glossary.md
```diff
@@ -1,5 +1,49 @@
...
+```
+
+In this case, both `r` and `s` alias each other, since they both point to all of
+the bytes of `u`.
+
+However, `head` and `tail` do not alias each other: `head` points to the first
+byte of `u` and `tail` points to the other seven bytes of `u` after it. Also,
+both `head` and `tail` alias `s`.
+
+* The span length of `&T`, `&mut T`, `*const T`, or `*mut T` when `T` is
```
RalfJung: What about adding a header, to bind `t`? "The span length of a pointer or reference `t` is defined as follows:"

Lokathor: yeah lemme try something like that

reference/src/glossary.md
Lokathor: If you mean https://github.com/Lokathor/unsafe-code-guidelines/blob/lokathor/wip/stacked-borrows.md the start of that says that it's not endorsed by the UCG WG. We could make it endorsed, and link to it, but that should be a separate PR if so.

reference/src/glossary.md
Lokathor: Alright, I think I improved that part quite a bit.

reference/src/glossary.md
RalfJung: The idea was to link to it as an *example* of what an aliasing model *could* look like, not as *the* aliasing model.

Lokathor: Alright, check now

reference/src/glossary.md
```diff
@@ -1,5 +1,57 @@
...
+```
+
+In this case, both `r` and `s` alias each other, since they both point to all of
+the bytes of `u`.
+
+However, `head` and `tail` do not alias each other: `head` points to the first
+byte of `u` and `tail` points to the other seven bytes of `u` after it. Both `head`
+and `tail` alias `s`, any overlap is sufficient to count as an alias.
+
+* The span of a pointer or reference is the size of the value being pointed to or referenced.
```
RalfJung: I still maintain that this enumeration (in general, any enumeration) should have a header.

reference/src/glossary.md
```diff
@@ -1,5 +1,57 @@
...
+* For some type `T` that is [`Sized`](https://doc.rust-lang.org/core/marker/trait.Sized.html)
```
```suggestion
* For a type `T` that is [`Sized`](https://doc.rust-lang.org/core/marker/trait.Sized.html)
```

reference/src/glossary.md
```diff
@@ -1,5 +1,57 @@
...
+  The span length of a pointer or reference to `T` is found with `size_of::<T>()`.
+* When `T` is not `Sized` the story is a little tricker:
+  * If you have a reference `r` you can use `size_of_val(r)` to determine the
+    span of the reference.
+  * If you have a pointer `p` you must unsafely convert that to a reference before
+    you can use `size_of_val`. There is not currently a safe way to determine the
+    span of a pointer to an unsized type.
+* The Data Layout chapter also has more information on the sizes of different types.
```
RalfJung: What is the purpose of this note? It surely shouldn't be part of the *definition*.

reference/src/glossary.md
```diff
@@ -1,5 +1,57 @@
 ## Glossary
 
+#### Aliasing
+
+(Please note: a full aliasing model for Rust has not yet been constructed, but
+at the moment we can give the following definition. The most developed potential
+aliasing model so far is known as "Stacked Borrows", and can be found
+[here](https://github.com/Lokathor/unsafe-code-guidelines/blob/lokathor/wip/stacked-borrows.md).)
```
RalfJung: Please don't link to your fork. Link to the official repository's master branch.

Lokathor: oh whoops. didn't mean that one

reference/src/glossary.md
Lokathor: Well, it is impossible to safely use `size_of_val` with `*const [T]` and `*mut [T]` if you don't know the span size that must be valid and not aliasing but you _can_, if you're feeling spicy, transmute your fat pointer into a layout-equivalent `DynObject` or `Slice<T>` and then read the `usize` field, for example.

It's worth a mention that people should go check that chapter I think.

reference/src/glossary.md
Lokathor: I'm confused, you mean like some sort of title for the bullet points?

RalfJung: Yes:
```
This title describes what the enumeration enumerates:
* enumeration item 1
* enumeration item 2
```

reference/src/glossary.md
RalfJung: But that's covered already by "unsafely convert that to a reference".

Mentioning going to the chapter seems fine (then it should be a link though). But it's not part of the definition of a "span", so it should be *after* the enumeration, not *in* the enumeration.

Lokathor: Okay, between this point and the other unresolved point I now understand that you're looking at this list _very_ differently from how I intended it. But we can go with that.

RalfJung:
Ah whatever, I think I will just do an editing pass and also define "Tag" while I am at it. ;)  Thanks a lot @Lokathor !

oberien:
I think I'm a bit late to this party, but how does the definition here interact with slices of ZSTs (#168)?

> A span of memory is similar to how a slice works: there's a base byte address as well as a length in
bytes.

> One interesting side effect of these rules is that references and pointers to Zero Sized Types _never_ alias each other, because their span length is always 0 bytes.

According to these rules, slices of ZSTs always have a span byte length 0 and thus _never_ alias. But I'd argue that `a` and `b` in the following code *do* alias the same allocation, even though the allocation has a length of 0 bytes:

```rust
let foo = [(); 1337];
let a = &foo[..];
let b = &foo[..7];
```

RalfJung:
> According to these rules, slices of ZSTs always have a span byte length 0 and thus never alias.

That is correct, and actually explicitly stated further down in this PR.

> But I'd argue that a and b in the following code do alias the same allocation, even though the allocation has a length of 0 bytes:

I'd argue you are wrong, then. ;) Also see https://github.com/rust-lang/rust-memory-model/issues/44.
What is the benefit of adding a special case for ZST this way?

oberien:
I think my misunderstanding here is that mutable references are highly underdocumented. The only "real" documentation of them I know of is the following part of the nomicon:

> There are two kinds of reference:
>
> * Shared reference: &
> * Mutable reference: &mut
>
> Which obey the following rules:
>
> * A reference cannot outlive its referent
> * A mutable reference cannot be aliased
>
> That's it. That's the whole model references follow.

Reading through the linked issue, I understand that those requirements are only *necessary* and not *sufficient* definitions. I somewhat interpreted them as *sufficient* definitions and thus interpreted the noaliasing rule of ZSTs such that you can have multiple mutable references to the same ZST at the same time, which obviously is just wrong. I think I just need to split the meanings of "aliasing" and "references" in my brain.

RalfJung:
You can think of there being arbitrarily many ZST "between" any two memory addresses, and references can point to those. So even if two ZST references point to the same location, they don't "overlap".

That pragmatic response is that aliasing is about figuring out if memory operations through one reference can be observed through the other reference, or if they can be freely reordered. But no memory access actually ever happens with a ZST reference.

PR #171, opened 2019-07-21
[function pointers example] Marked the function pointer and exported function as `unsafe`

danielhenrymantilla:
Although the example was targeted at C FFI, in which case `unsafe` has no value, one should never forget that these functions can also be called from Rust side. **Having both the ffi function and the callback be unsafe is more realistic of there being a raw pointer involved** (else the only way to provide such a safe callback from Rust would be if it never dereferenced the `thunk` parameter).

### DIFF
--- a/reference/src/layout/function-pointers.md
+++ b/reference/src/layout/function-pointers.md
@@ -97,9 +97,9 @@ pub extern "C" fn cons(node: Option<Box<Cons>>, data: c_int) -> Box<Cons> {
 }
 
 #[no_mangle]
-pub extern "C" fn iterate(
+pub unsafe extern "C" fn iterate(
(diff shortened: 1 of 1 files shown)


Lokathor:
Given how the C abi affects your ability to properly unwind and how it's currently being worked on and all, having a safe C abi function is certainly questionable at best. As I understand it, even calling Rust->Rust using C abi gives you all that unwind trouble.

Good change

gnzlbg:
Do we need this example at all ? It took me a while to go through the whole page on function types, just to be able to understand their layout and validity invariant, and to me it did not feel like the example added much.

RalfJung:
Okay, merging then.

PR #173, opened 2019-07-24
Document the abbreviation of zero-sized types.

### LINKED ISSUE #172: Abbreviation for zero-sized types with particular alignment
We often need to talk about zero-sized types (ZSTs) with a particular alignment requirement, e.g., 1, or in general, N:

```rust
#[repr(align(N))] struct Zst;
```

It was suggested in https://github.com/rust-lang/unsafe-code-guidelines/pull/164#discussion_r303198338 that we should have an abbreviation for this defined in the glossary, where we also define the zero-sized type (ZST) abbreviation, to avoid writing "a zero-sized one-aligned type" and similar constructs all over the place.

A proposal was to use the `N-ZST` schema for an "N-aligned zero-sized type", e.g., `1-ZST` would stand for a "one-aligned zero-sized type".

gnzlbg:
Closes #172 .

### DIFF
--- a/reference/src/glossary.md
+++ b/reference/src/glossary.md
@@ -80,7 +80,7 @@ This document also uses the "1-ZST" abbreviation, which stands for "one-aligned
 zero-sized type", to refer to zero-sized types with an alignment requirement of 1. 
 
 For example, `()` is a "1-ZST" but `[u16; 0]` is not because it has an alignment
-requirement of two.
+requirement of 2.
 
 ### TODO
 
--- a/reference/src/glossary.md
+++ b/reference/src/glossary.md
@@ -73,6 +73,13 @@ niches. For example, the "all bits uninitialized" is an invalid bit-pattern for
 `&mut T`, but this bit-pattern cannot be used by layout optimizations, and is not a
 niche.
 
+#### Zero-sized type / ZST
+
+Types with zero size are called zero-sized types, which is abbreviated as "ZST".
+When the alignment of the type is relevant, the schema "N-ZST" which stands for
+"N-aligned zero-sized type" is used. For example: "1-ZST" stands for
+"one-aligned zero-sized type", that is, a type with a size of zero and an
+alignment requirement of one.
 
 ### TODO
 
--- a/reference/src/glossary.md
+++ b/reference/src/glossary.md
@@ -76,10 +76,9 @@ niche.
 #### Zero-sized type / ZST
 
 Types with zero size are called zero-sized types, which is abbreviated as "ZST".
-When the alignment of the type is relevant, the schema "N-ZST" which stands for
-"N-aligned zero-sized type" is used. For example: "1-ZST" stands for
-"one-aligned zero-sized type", that is, a type with a size of zero and an
-alignment requirement of one.
+This document also uses the "1-ZST" abbreviation, which stands for "one-aligned
+zero-sized type", to refer to zero-sized types with an alignment requirement of
+1.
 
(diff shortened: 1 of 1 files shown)


reference/src/glossary.md
```diff
@@ -73,6 +73,13 @@ niches. For example, the "all bits uninitialized" is an invalid bit-pattern for
 `&mut T`, but this bit-pattern cannot be used by layout optimizations, and is not a
 niche.
 
+#### Zero-sized type / ZST
+
+Types with zero size are called zero-sized types, which is abbreviated as "ZST".
+When the alignment of the type is relevant, the schema "N-ZST" which stands for
+"N-aligned zero-sized type" is used. For example: "1-ZST" stands for
+"one-aligned zero-sized type", that is, a type with a size of zero and an
+alignment requirement of one.
```
RalfJung: I think this should say "at most" somewhere, i.e., a 2-ZST has alignment requirement *at most* 2.

Lokathor: I would expect only _exactly_ two to count as `2-ZST`.

RalfJung: lol. Well good that we have this discussion then. ;)

I didn't think about which one makes mo sense, to be fair. Just went with what seemed "obvious" -- most of the time you don't care if types are actually "more relaxed" about alignment than what you think they are.

rkruppe: If we're going to lose time nitpicking this, then I would prefer to drop the generalization for now (for n=1 there's no difference) and only introduce it if we ever need the notion of an N-ZST for N > 1.

Lokathor: If we need to express simple requirements we could use something like >1ZST for "ZST with alignment greater than 1"

but there is indeed no current need for this particular extension of the "1ZST" name, so we could punt it to later.

RalfJung: Agreed. We can restrict this to 1-ZST for now.

@gnzlbg r=me with that change.

gnzlbg: Done.

reference/src/glossary.md
```diff
@@ -73,6 +73,12 @@ niches. For example, the "all bits uninitialized" is an invalid bit-pattern for
 `&mut T`, but this bit-pattern cannot be used by layout optimizations, and is not a
 niche.
 
+#### Zero-sized type / ZST
+
+Types with zero size are called zero-sized types, which is abbreviated as "ZST".
+This document also uses the "1-ZST" abbreviation, which stands for "one-aligned
+zero-sized type", to refer to zero-sized types with an alignment requirement of
+1.
```
RalfJung: Might be worth mentioning (because this is not terribly widely known) that `[u32; 0]` is an example of a ZST that is *not* a 1-ZST.

reference/src/glossary.md
```diff
@@ -73,6 +73,14 @@ niches. For example, the "all bits uninitialized" is an invalid bit-pattern for
...
 niche.
 
+#### Zero-sized type / ZST
+
+Types with zero size are called zero-sized types, which is abbreviated as "ZST".
+This document also uses the "1-ZST" abbreviation, which stands for "one-aligned
+zero-sized type", to refer to zero-sized types with an alignment requirement of 1. 
+
+For example, `()` is a "1-ZST" but `[u16; 0]` is not because it has an alignment
+requirement of two.
```
```suggestion
requirement of 2.
```
Or it should be "one" in the previous paragraph.

RalfJung:
Thanks!

PR #175, opened 2019-07-25
WIP documents for memory interface and value domain

RalfJung:
This also resolves https://github.com/rust-lang/unsafe-code-guidelines/pull/40. I concede to the majority opinion that the 4 terms in question are "value", "place", "value expression" and "place expression".

EDIT: Oh, I should also mention: just like the other WIP document, this is basically my personal proposal / opinion; I was hoping for a review to increase clarity but I do not think we should strive for consensus on the opinions expressed in those documents.

### DIFF
--- a/wip/memory-interface.md
+++ b/wip/memory-interface.md
@@ -4,21 +4,22 @@
 
 The purpose of this document is to describe the interface between a Rust program and memory.
 This interface is a key part of the Rust Abstract Machine: it lets us separate concerns by splitting the Machine (i.e., its specification) into two pieces, connected by this well-defined interface:
-* The *expression/statement semantics* of Rust boils down to explaining which "memroy events" (calls to the memory interface) happen in which order.
+* The *expression/statement semantics* of Rust boils down to explaining which "memroy events" (calls to the memory interface) happen in which order. This part of the specification is *pure* in the sense that it has no "state": everything that needs to be remembered from one expression evaluation to the next is communicated through memory.
 * The Rust *memory model* explains which interactions with the memory are legal (the others are UB), and which values can be returned by reads.
 
-The interface is also opinionated in several ways; this is not intended to be able to support *any imaginable* memory model, but rather start the process of reducing the design space of what we consider a "reasonable" memory model for Rust.
-For example, it explicitly acknowledges that pointers are not just integers and that uninitialized memory is special (both are true for C and C++ as well but you have to read the standard very careful, and consult non-normative defect report responses, to see this).
+The interface shown below is also opinionated in several ways.
+It is not intended to be able to support *any imaginable* memory model, but rather start the process of reducing the design space of what we consider a "reasonable" memory model for Rust.
+For example, it explicitly acknowledges that pointers are not just integers and that uninitialized memory is special (both are true for C and C++ as well but you have to read the standard very careful, and consult [non-normative defect report responses](http://www.open-std.org/jtc1/sc22/wg14/www/docs/dr_260.htm), to see this).
 Another key property of the interface presented below is that it is *untyped*.
-This encodes the fact that in Rust, *operations are typed, but memory is not*---a key difference to C and C++ with their type-based strict aliasing rules.
-At the same time, the memory model provides a *side-effect free* way to turn pointers into "raw bytes", which is *not* [the direction C++ is moving towards](http://www.open-std.org/jtc1/sc22/wg14/www/docs/n2364.pdf), so we might have to revisit this choice later.
+This implies that in Rust, *operations are typed, but memory is not*---a key difference to C and C++ with their type-based strict aliasing rules.
+At the same time, the memory model provides a *side-effect free* way to turn pointers into "raw bytes", which is *not* [the direction C++ is moving towards](http://www.open-std.org/jtc1/sc22/wg14/www/docs/n2364.pdf), and we might have to revisit this choice later if it turns out to not be workable.
 
 ## Pointers
 
 One key question a memory model has to answer is *what is a pointer*.
 It might seem like the answer is just "an integer of appropriate size", but [that is not the case][pointers-complicated].
 This becomes even more prominent with aliasing models such as [Stacked Borrows].
-So we will leave this question open, and treat `Pointer` as an "associated type" of the memory interface
+So the interface will leave it up to the concrete instance to answer this question, and carry `Pointer` as an associated type.
 
 ## Bytes
 
@@ -58,7 +59,7 @@ impl<Pointer: PtrToInt> Byte<Pointer> {
             Byte::Raw(int) => Some(int),
             Byte::Uninit => None,
             Byte::PtrFragment { ptr, idx } =>
-                ptr.get_byte(idx),
+                Some(ptr.get_byte(idx)),
         }
     }
 }
@@ -96,9 +97,10 @@ trait Memory {
     fn read(&mut self, ptr: Self::Pointer, len: u64) -> Result<Vec<Byte<Self::Pointer>>, Error>;
 
     /// Offset the given pointer.
-    fn offset(&mut self, ptr: Self::Pointer, offset: u64, mode: OffsetMode) -> Result<Self::Pointer, Error>;
+    fn offset(&mut self, ptr: Self::Pointer, offset: u64, mode: OffsetMode)
+        -> Result<Self::Pointer, Error>;
 
-    /// Cast the given integer to a pointer.
+    /// Cast the given integer to a pointer.  (The other direction is handled by `PtrToInt` below.)
     fn int_to_ptr(&mut self, int: u64) -> Result<Self::Pointer, Error>;
 }
 
@@ -128,8 +130,6 @@ This is a very basic memory interface that is incomplete in at least the followi
 * To represent concurrency, many operations need to take a "thread ID" and `read` and `write` need to take an [`Ordering`].
 * To represent [Stacked Borrows], there needs to be a "retag" operation, and that one will in fact be "lightly typed" (it cares about `UnsafeCell`).
 * Maybe we want operations that can compare pointers without casting them to integers.
-
-But I think it can still be useful to provide some basic terminology and grounds for further discussion.
 
 [pointers-complicated]: https://www.ralfj.de/blog/2018/07/24/pointers-and-bytes.html
 [uninit]: https://www.ralfj.de/blog/2019/07/14/uninit.html
--- a/wip/memory-interface.md
+++ b/wip/memory-interface.md
@@ -4,8 +4,10 @@
 
 The purpose of this document is to describe the interface between a Rust program and memory.
 This interface is a key part of the Rust Abstract Machine: it lets us separate concerns by splitting the Machine (i.e., its specification) into two pieces, connected by this well-defined interface:
-* The *expression/statement semantics* of Rust boils down to explaining which "memory events" (calls to the memory interface) happen in which order - expressed as calls to the methods of this interface, and reactions to its return values. This part of the specification is *pure* in the sense that it has no "state": everything that needs to be remembered from one expression evaluation to the next is communicated through memory.
-* The Rust *memory model* explains which interactions with the memory are legal (the others are UB), and which values can be returned by reads. A memory model is defined by implementing the memory interface.
+* The *expression/statement semantics* of Rust boils down to explaining which "memory events" (calls to the memory interface) happen in which order - expressed as calls to the methods of this interface, and reactions to its return values.
+  This part of the specification is *pure* in the sense that it has no "state": everything that needs to be remembered from one expression evaluation to the next is communicated through memory.
+* The Rust *memory model* explains which interactions with the memory are legal (the others are UB), and which values can be returned by reads.
+  A memory model is defined by implementing the memory interface.
 
 The interface shown below is also opinionated in several ways.
 It is not intended to be able to support *any imaginable* memory model, but rather start the process of reducing the design space of what we consider a "reasonable" memory model for Rust.
--- a/wip/memory-interface.md
+++ b/wip/memory-interface.md
@@ -9,9 +9,9 @@ This interface is a key part of the Rust Abstract Machine: it lets us separate c
 
 The interface shown below is also opinionated in several ways.
 It is not intended to be able to support *any imaginable* memory model, but rather start the process of reducing the design space of what we consider a "reasonable" memory model for Rust.
-For example, it explicitly acknowledges that pointers are not just integers and that uninitialized memory is special (both are true for C and C++ as well but you have to read the standard very careful, and consult [non-normative defect report responses](http://www.open-std.org/jtc1/sc22/wg14/www/docs/dr_260.htm), to see this).
+For example, it explicitly acknowledges that pointers are not just integers and that uninitialized memory is special (both are true for C and C++ as well but you have to read the standard very careful, and consult [defect report responses](http://www.open-std.org/jtc1/sc22/wg14/www/docs/dr_260.htm), to see this).
 Another key property of the interface presented below is that it is *untyped*.
-This implies that in Rust, *operations are typed, but memory is not*---a key difference to C and C++ with their type-based strict aliasing rules.
+This implies that in Rust, *operations are typed, but memory is not* - a key difference to C and C++ with their type-based strict aliasing rules.
 At the same time, the memory model provides a *side-effect free* way to turn pointers into "raw bytes", which is *not* [the direction C++ is moving towards](http://www.open-std.org/jtc1/sc22/wg14/www/docs/n2364.pdf), and we might have to revisit this choice later if it turns out to not be workable.
 
 ## Pointers
@@ -19,7 +19,7 @@ At the same time, the memory model provides a *side-effect free* way to turn poi
 One key question a memory model has to answer is *what is a pointer*.
 It might seem like the answer is just "an integer of appropriate size", but [that is not the case][pointers-complicated].
 This becomes even more prominent with aliasing models such as [Stacked Borrows].
-So the interface will leave it up to the concrete instance to answer this question, and carry `Pointer` as an associated type.
+So we leave it up to the memory model to answer this question, and make `Pointer` an associated type.
 Practically speaking, `Pointer` will be some representation of an "address", plus [provenance] information.
 
 [provenance]: https://github.com/rust-lang/unsafe-code-guidelines/blob/master/reference/src/glossary.md#pointer-provenance
@@ -27,7 +27,7 @@ Practically speaking, `Pointer` will be some representation of an "address", plu
 ## Bytes
 
 The unit of communication between the memory model and the rest of the program is a *byte*.
-Again the question of "what is a byte" is not as trivial as it might seem; beyond `u8` values we have to represent `Pointer`s and [uninitialized memory][uninit].
+Again, the question of "what is a byte" is not as trivial as it might seem; beyond `u8` values we have to represent `Pointer`s and [uninitialized memory][uninit].
 We define the `Byte` type (in terms of an arbitrary `Pointer` type) as follows:
 
 ```rust
@@ -39,6 +39,7 @@ enum Byte<Pointer> {
     /// One byte of a pointer.
     PtrFragment {
         /// The pointer of which this is a byte.
+        /// That is, the byte is a fragment of this pointer.
         ptr: Pointer,
         /// Which byte of the pointer this is.
         /// `idx` will always be in `0..PTR_SIZE`.
@@ -47,10 +48,15 @@ enum Byte<Pointer> {
 }
 ```
 
-The purpose of `PtrFragment` is to be able to have a byte-wise representation of a `Pointer`.
+The purpose of `PtrFragment` is to enable a byte-wise representation of a `Pointer`.
 On a 32-bit system, the sequence of 4 bytes representing `ptr: Pointer` is:
 ```
-[PtrFragment { ptr, idx: 0 }, PtrFragment { ptr, idx: 1 }, PtrFragment { ptr, idx: 2 }, PtrFragment { ptr, idx: 3 }]
+[
+    PtrFragment { ptr, idx: 0 },
+    PtrFragment { ptr, idx: 1 },
+    PtrFragment { ptr, idx: 2 },
+    PtrFragment { ptr, idx: 3 },
+]
 ```
 
 Based on the `PtrToInt` trait (see below), we can turn every initialized `Byte` into an integer in `0..256`:
@@ -73,13 +79,18 @@ impl<Pointer: PtrToInt> Byte<Pointer> {
 The Rust memory interface is described by the following (not-yet-complete) trait definition:
 
 ```rust
+/// All operations are fallible, so they return `Result`.  If they fail, that
+/// means the program caused UB. What exactly the `UndefinedBehavior` type is
+/// does not matter here.
+type Result<T=()> = std::result::Result<T, UndefinedBehavior>;
+
 /// *Note*: All memory operations can be non-deterministic, which means that
 /// executing the same operation on the same memory can have different results.
 /// We also let all operations potentially mutate memory. For example, reads
 /// actually do change the current state when considering concurrency or
 /// Stacked Borrows.
-/// And finally, all operations are fallible (they return `Result`); if they
-/// fail, that means the program caused UB.
+/// This is pseudo-Rust, so we just use fully owned types everywhere for
+/// symmetry and simplicity.
 trait Memory {
     /// The type of pointer values.
     type Pointer: Copy + PtrToInt;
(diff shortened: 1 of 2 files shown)


reference/src/glossary.md
```diff
@@ -73,10 +73,44 @@ niches. For example, the "all bits uninitialized" is an invalid bit-pattern for
...
+* and loading a [value](#value) of the same type from it (through the place-to-value coercion).
+
+
+#### Value
+
+A *value* (called "value of the expression" or "rvalue" in C and "prvalue" in C++) is what gets stored in a [place](#place), and also the result of computing a [*value expression*][place-value-expr].
+A value has a type.
+For example, a value of type `u8` is a mathematical integer in the range `0..256`.
+Values can be (according to their type) turned into a list of bytes, which is called a [representation](#representation) of the value.
+Values are ephemeral; they arise during the computation of an instruction but are only ever persisted in memory through their representation.
```
Lokathor: Does "ephemeral" fit here? Is there anything to a value other than its byte representation which exists during a computation and then doesn't get persisted in memory?

wip/value-domain.md
```diff
@@ -0,0 +1,128 @@
...
+        idx: u64,
+        data: Box<Self>,
+    },
+    /// A "bag of raw bytes".
+    RawBag(Vec<Byte<Pointer>>),
+    /* ... */
+}
+```
+
+As Rust grows, we might expand this definition. That is okay; all previously defined representation relations are still well-defined when the domain grows, the newly added values will just not be valid for old types as one would expect.
```
Lokathor: So, is it _complete for current rust_ and just not future-proof? Or is it incomplete _even in current rust_ and it's just WIP and don't worry about it?

reference/src/glossary.md
gnzlbg: Pointer provenance maybe ? AFAICT that gets persisted in memory but it is not part of their representation, although we haven't clarified this.

wip/memory-interface.md
```diff
@@ -0,0 +1,137 @@
...
+* The *expression/statement semantics* of Rust boils down to explaining which "memroy events" (calls to the memory interface) happen in which order. This part of the specification is *pure* in the sense that it has no "state": everything that needs to be remembered from one expression evaluation to the next is communicated through memory.
+* The Rust *memory model* explains which interactions with the memory are legal (the others are UB), and which values can be returned by reads.
+
+The interface shown below is also opinionated in several ways.
+It is not intended to be able to support *any imaginable* memory model, but rather start the process of reducing the design space of what we consider a "reasonable" memory model for Rust.
+For example, it explicitly acknowledges that pointers are not just integers and that uninitialized memory is special (both are true for C and C++ as well but you have to read the standard very careful, and consult [non-normative defect report responses](http://www.open-std.org/jtc1/sc22/wg14/www/docs/dr_260.htm), to see this).
```
gnzlbg: I thought defect reports were normative, even if the conclusion of the report is "nothing has to be done".

wip/memory-interface.md
```diff
@@ -0,0 +1,137 @@
...
+Another key property of the interface presented below is that it is *untyped*.
+This implies that in Rust, *operations are typed, but memory is not*---a key difference to C and C++ with their type-based strict aliasing rules.
+At the same time, the memory model provides a *side-effect free* way to turn pointers into "raw bytes", which is *not* [the direction C++ is moving towards](http://www.open-std.org/jtc1/sc22/wg14/www/docs/n2364.pdf), and we might have to revisit this choice later if it turns out to not be workable.
+
+## Pointers
+
+One key question a memory model has to answer is *what is a pointer*.
+It might seem like the answer is just "an integer of appropriate size", but [that is not the case][pointers-complicated].
+This becomes even more prominent with aliasing models such as [Stacked Borrows].
+So the interface will leave it up to the concrete instance to answer this question, and carry `Pointer` as an associated type.
```
gnzlbg: These reads like it expect multiple different Pointer types. It might be helpful to maybe mention this here somehow.

wip/memory-interface.md
```diff
@@ -0,0 +1,137 @@
...
+/// fail, that means the program caused UB.
+trait Memory {
+    /// The type of pointer values.
+    type Pointer: Copy + PtrToInt;
+
+    /// The size of pointer values.
+    const PTR_SIZE: u64;
+
+    /// Create a new allocation.
+    fn allocate(&mut self, size: u64, align: u64) -> Result<Self::Pointer, Error>;
```
gnzlbg: Is there a reason this doesn't use `std::alloc::Layout` and prefers `size` and `align` ?

wip/memory-interface.md
```diff
@@ -0,0 +1,137 @@
...
+
+    /// Remove an allocation.
+    fn deallocate(&mut self, ptr: Self::Pointer, size: u64, align: u64) -> Result<(), Error>;
+
+    /// Write some bytes to memory.
+    fn write(&mut self, ptr: Self::Pointer, bytes: Vec<Byte<Self::Pointer>>) -> Result<(), Error>;
```
gnzlbg: Why does bytes need to own the bytes here ?

wip/value-domain.md
```diff
@@ -0,0 +1,128 @@
...
+The Rust value domain is described by the following (incomplete) type definition (in terms of an arbitrary `Pointer` type instantiated by [the memory interface][memory-interface]):
+
+```rust
+enum Value<Pointer> {
+    /// A mathematical integer.
+    Int(BigInt),
+    /// A Boolean value.
+    Bool(bool),
+    /// A pointer.
+    Ptr(Pointer),
```
gnzlbg: Are references pointers here ?

wip/value-domain.md
```diff
@@ -0,0 +1,128 @@
...
+    Uninit,
+    /// An n-tuple.
+    Tuple(Vec<Self>),
+    /// A variant of a sum type.
+    Variant {
+        idx: u64,
+        data: Box<Self>,
+    },
+    /// A "bag of raw bytes".
+    RawBag(Vec<Byte<Pointer>>),
```
gnzlbg: Why is this needed ? An array should be able to cover this.

wip/value-domain.md
```diff
@@ -0,0 +1,128 @@
...
+    /// A mathematical integer.
+    Int(BigInt),
+    /// A Boolean value.
+    Bool(bool),
+    /// A pointer.
+    Ptr(Pointer),
+    /// An uninitialized value.
+    Uninit,
+    /// An n-tuple.
+    Tuple(Vec<Self>),
```
gnzlbg: Does this covers tuple, tuple structs, structs, and arrays ?

reference/src/glossary.md
```diff
@@ -73,10 +73,44 @@ niches. For example, the "all bits uninitialized" is an invalid bit-pattern for
...
 
 ### TODO
 
 * *tag*
-* *rvalue*
-* *lvalue*
-* *representation*
+
+[value-domain]: https://github.com/rust-lang/unsafe-code-guidelines/tree/master/wip/value-domain.md
+[place-value-expr]: https://doc.rust-lang.org/reference/expressions.html#place-expressions-and-value-expressions
```
gnzlbg: I think these should be repeated here, for completeness.

wip/value-domain.md
```diff
@@ -0,0 +1,128 @@
...
+## Values
+
+The Rust value domain is described by the following (incomplete) type definition (in terms of an arbitrary `Pointer` type instantiated by [the memory interface][memory-interface]):
+
+```rust
+enum Value<Pointer> {
+    /// A mathematical integer.
+    Int(BigInt),
+    /// A Boolean value.
+    Bool(bool),
```
Centril: Ostensibly it would be more precise to define `Int` and `Bool` using a sigma type with `({x: Nat | x ∈ start..end}, start: Nat, end: Nat)` or using [`Fin : (n : Nat) -> Type`](https://github.com/idris-lang/Idris-dev/blob/master/libs/base/Data/Fin.idr#L11) but well... can't use Rust for notation anymore. =P

wip/value-domain.md
Centril: Also what about our friend `#[rustc_layout_scalar_valid_range_start(N)]`?

wip/value-domain.md
```diff
@@ -0,0 +1,128 @@
...
+    /// A pointer.
+    Ptr(Pointer),
+    /// An uninitialized value.
+    Uninit,
+    /// An n-tuple.
+    Tuple(Vec<Self>),
+    /// A variant of a sum type.
+    Variant {
+        idx: u64,
```
Centril: Shouldn't this be abstracted away? `discriminant_value` returns this today but that might not necessarily hold tomorrow.

wip/value-domain.md
```diff
@@ -0,0 +1,128 @@
...
+A value `Tuple([t, u])` is represented by a list of bytes `rt ++ pad1 ++ ru ++ pad2` (using `++` for list concatenation) if:
+
+* `t` is represented by `rt` at type `T`.
+* `u` is represented by `ru` at type `U`.
+* The length of `rt ++ pad1` is equal to the offset of the `U` field.
+* The length of the entire list `rt ++ pad1 ++ ru ++ pad2` is equal to the size of `Pair<T, U>`.
+
+This relation specifies that values of type `Pair` are always 2-tuples (aka, pairs).
+It also says that the actual content of the padding bytes is entirely irrelevant, we only care to have the right number of them to "pad" `ru` to the right place and to "pad" the entire list to have the right length.
+So, for example when considering `Pair<u8, u16>`, the value `Tuple[42, 119]` is represented on a little-endian target by `[Raw(42), byte, Raw(119), Raw(0)]` for *any* `byte: Byte`.
```
Centril: Speaking of... is endianess an associated type of "`Machine`"?

wip/value-domain.md
RalfJung: It's not complete for anything.

RalfJung:
Wow, so much feedback so quick! Thanks a lot. I hope I can go over it tonight.

One thing I should have mentioned: the part in the glossary is hopefully uncontroversial, we should agree on it for merging. The "WIP" part is, just like the other WIP document, basically my personal proposal / opinion; I was hoping for a review to increase clarity but I do not think we should strive for consensus on the opinions expressed in those documents.

reference/src/glossary.md
RalfJung: The provenance of a pointer is part of `Pointer`.

wip/memory-interface.md
RalfJung: From what I was told, they are normative for the standard they were reported against, but when reading C18 I don't have to read all defect reports of all prior versions. That would be ludicrous.

C18 contains the exact same wording as C11 for indeterminate values.

I can still remove the "non-normative" part of the sentence, it doesn't really add anything.

wip/memory-interface.md
RalfJung: Fewer dependencies.

wip/memory-interface.md
RalfJung: This is more pseudo-code than code. Using a different type for the argument of `read` and the result of `write` seemed silly.

wip/value-domain.md
RalfJung: Yes, this is about pointer *values* which are the values of both the reference and raw pointer *types*. I will clarify.

wip/value-domain.md
RalfJung: I don't follow. No other variant of `Value` even lets you have a `Byte`.

reference/src/glossary.md
RalfJung: What should be repeated?

reference/src/glossary.md
```diff
@@ -73,10 +73,44 @@ niches. For example, the "all bits uninitialized" is an invalid bit-pattern for
 `&mut T`, but this bit-pattern cannot be used by layout optimizations, and is not a
 niche.
 
+#### Place
+
+A *place* (called "lvalue" in C and "glvalue" in C++) is the result of computing a [*place expression*][place-value-expr].
+A place definitely stores some location in memory (a "pointer"), and might contain more information such as size or alignment (the details will have to be determined as the Rust Abstract Machine gets specified more precisely).
```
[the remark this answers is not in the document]
RalfJung: I don't understand the distinction.

wip/memory-interface.md
```diff
@@ -0,0 +1,137 @@
...
+
+```rust
+enum Byte<Pointer> {
+    /// The "normal" case: a (frozen, initialized) integer in `0..256`.
+    Raw(u8),
+    /// An uninitialized byte.
+    Uninit,
+    /// One byte of a pointer.
+    PtrFragment {
+        /// The pointer of which this is a byte.
```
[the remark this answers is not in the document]
RalfJung: Suggestions? I cannot come up with another phrasing that seems any less complex.

wip/memory-interface.md
```diff
@@ -0,0 +1,137 @@
...
+}
+```
+
+The purpose of `PtrFragment` is to be able to have a byte-wise representation of a `Pointer`.
+On a 32-bit system, the sequence of 4 bytes representing `ptr: Pointer` is:
+```
+[PtrFragment { ptr, idx: 0 }, PtrFragment { ptr, idx: 1 }, PtrFragment { ptr, idx: 2 }, PtrFragment { ptr, idx: 3 }]
+```
+
+Based on the `PtrToInt` trait (see below), we can turn every initialized `Byte` into an integer in `0..256`:
```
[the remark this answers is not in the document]
RalfJung: That's why I said *initialized* `Byte`.

wip/memory-interface.md
```diff
@@ -0,0 +1,137 @@
...
+/// actually do change the current state when considering concurrency or
+/// Stacked Borrows.
+/// And finally, all operations are fallible (they return `Result`); if they
+/// fail, that means the program caused UB.
+trait Memory {
+    /// The type of pointer values.
+    type Pointer: Copy + PtrToInt;
+
+    /// The size of pointer values.
+    const PTR_SIZE: u64;
```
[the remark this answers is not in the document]
RalfJung: Maybe? I think putting it in memory for now works just as well.

reference/src/glossary.md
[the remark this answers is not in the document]
Centril: It seems to me "memory location" refers to the cells holding the data itself and "stores some location" is a pointer to those cells? -- a question of indirection that is?

wip/memory-interface.md
```diff
@@ -0,0 +1,137 @@
...
+    }
+}
+```
+
+## Memory interface
+
+The Rust memory interface is described by the following (not-yet-complete) trait definition:
+
+```rust
+/// *Note*: All memory operations can be non-deterministic, which means that
```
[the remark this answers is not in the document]
RalfJung: This is about the semantics of run-time Rust. Why are you asking about `const fn`?

wip/memory-interface.md
```diff
@@ -0,0 +1,137 @@
...
+/// executing the same operation on the same memory can have different results.
+/// We also let all operations potentially mutate memory. For example, reads
+/// actually do change the current state when considering concurrency or
+/// Stacked Borrows.
+/// And finally, all operations are fallible (they return `Result`); if they
+/// fail, that means the program caused UB.
+trait Memory {
```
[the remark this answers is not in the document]
RalfJung: Agda is way too much like Haskell to be readable. ;)

wip/memory-interface.md
```diff
@@ -0,0 +1,137 @@
...
+The Rust memory interface is described by the following (not-yet-complete) trait definition:
+
+```rust
+/// *Note*: All memory operations can be non-deterministic, which means that
+/// executing the same operation on the same memory can have different results.
+/// We also let all operations potentially mutate memory. For example, reads
+/// actually do change the current state when considering concurrency or
+/// Stacked Borrows.
+/// And finally, all operations are fallible (they return `Result`); if they
+/// fail, that means the program caused UB.
```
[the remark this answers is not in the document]
RalfJung: That just adds noise though.  So no I think that's not a good idea.

wip/value-domain.md
RalfJung: I also don't think I agree. If this was Coq I'd use `Z`, not some `Fin` type, for `Int`.

wip/memory-interface.md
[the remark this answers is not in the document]
Centril: Ah woops; read too fast!

wip/value-domain.md
RalfJung: Maybe.

reference/src/glossary.md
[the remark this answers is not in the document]
RalfJung: In my terminology, "location" is what we call the value of ML's `ref` types.

It *identifies* a cell that holds data, but it is not itself that cell.
A `Pointer` is basically location + provenance.

wip/memory-interface.md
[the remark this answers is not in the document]
Centril: Well a `const fn` could execute at run-time as well and we could ostensibly tell LLVM about the deterministic aspects of it?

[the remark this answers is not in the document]
RalfJung: Maybe, but what does that comment have to do with a basic doc comment about the *interface* of run-time memory?

One day when we have not just an interface but also an implementation of it (i.e., a memory model), and we are writing a document where we formally argue why our compilation to LLVM is correct... *then* I think your question becomes relevant. But I don't see how its answer affects the document we are discussing here.

reference/src/glossary.md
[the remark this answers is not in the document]
Centril: Ah; can we clarify this distinction (I don't have suggestions... :D)?

wip/memory-interface.md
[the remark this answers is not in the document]
Centril: Maybe just add:
```suggestion
        /// The pointer of which this is a byte.
        /// That is, the byte is a fragment of this pointer.
```
?

wip/memory-interface.md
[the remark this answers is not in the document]
Centril: Alternatively, maybe write `-> Result<Stuff, UndefinedBehavior>`?

wip/memory-interface.md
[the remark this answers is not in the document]
Centril: > Maybe, but what does that comment have to do with a basic doc comment about the _interface_ of run-time memory?

Fair enough; With emphasis on *"can"* I suppose that within a `const fn` context any non-deterministic an implementation would return `Error` then.

wip/value-domain.md
Centril: Ah I see. That makes sense, the range would be more appropriate for the value relation for a type instead. Tho... in that case, why do we need `Bool` at all and why is that not subsumed by `Int`?

wip/memory-interface.md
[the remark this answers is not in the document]
RalfJung: That, or the non-determinism must not be observable, or something like that.

The details of what actually *is* non-deterministic have to be filled in my the memory model anyway.

wip/value-domain.md
RalfJung: Are you a set theorist or what? ;) "boolean" and "integer" are two distinct mathematical concepts.

Centril: Sure, but why is this distinction made here rather than when relating values to `bool`?

wip/memory-interface.md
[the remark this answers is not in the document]
Centril: My thinking was that there are going to be more configurations like this so accumulating then in one place could be useful?

wip/value-domain.md
RalfJung: Because the entire point of `Value` is to capture the mathematical concepts that are represented by the data we store in memory. If there is no concept of a "boolean" in `Value`, then the language literally does not have the concept of a "boolean" anywhere.

reference/src/glossary.md
gnzlbg: I think this is made clearer below, where it is shown that bytes can belong to a pointer, and therefore provenance is part of the representation.

wip/memory-interface.md
gnzlbg: > From what I was told, they are normative for the standard they were reported against, but when reading C18 I don't have to read all defect reports of all prior versions. That would be ludicrous.

Sounds correct to me. The changes proposed in defect reports are automatically merged into the standard draft, so they become part of the next standard version automatically. That particular report is a bit weird, in that its conclusion is that the defect isn't a defect because the standard is clear and therefore nothing must be done. If people have to fill a defect report, and reference it everywhere, the standard is obviously not clear..

reference/src/glossary.md
gnzlbg: I meant the place-expression and value-expression part of the reference.

wip/value-domain.md
gnzlbg: Oh, I thought that BigInt would let you have a byte. So if aI cast a pointer to an `usize`, is that usize a `BigInt` or a `RawBag` ?

reference/src/glossary.md
RalfJung: I see. Seems weird to me to duplicate parts of the reference?

reference/src/glossary.md
```diff
@@ -73,10 +73,44 @@ niches. For example, the "all bits uninitialized" is an invalid bit-pattern for
...
+A place has a type.
+Places are ephemeral; they arise during the computation of an instruction but cannot be persisted in memory.
```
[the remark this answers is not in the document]
RalfJung: Not sure which one "above" you mean (depending on the page these are displayed in a different order), but... statics are a good point. I had not thought about them.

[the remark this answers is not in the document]
Centril: Argh... by "above" I mean "below" =P

(https://github.com/rust-lang/unsafe-code-guidelines/pull/175#discussion_r307252771)

reference/src/glossary.md
RalfJung: The difference between a value and its representation is comparable to the difference between run-time data and serialized data. The run-time data is also ephemeral in some sense; only serialized data is persisted across machine reboots, for example.

reference/src/glossary.md
```diff
@@ -73,10 +73,44 @@ niches. For example, the "all bits uninitialized" is an invalid bit-pattern for
...
+
+#### Value
+
+A *value* (called "value of the expression" or "rvalue" in C and "prvalue" in C++) is what gets stored in a [place](#place), and also the result of computing a [*value expression*][place-value-expr].
+A value has a type.
+For example, a value of type `u8` is a mathematical integer in the range `0..256`.
+Values can be (according to their type) turned into a list of bytes, which is called a [representation](#representation) of the value.
+Values are ephemeral; they arise during the computation of an instruction but are only ever persisted in memory through their representation.
+
+#### Representation (relation)
```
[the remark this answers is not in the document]
RalfJung: I was afraid going too formal would drive away too many readers.

wip/memory-interface.md
RalfJung: Agreed, I think the C committee's position here is harmful. But alas, that is not something we are going to fix here and now.

wip/memory-interface.md
RalfJung: @gnzlbg  I do not understand what you mean. This works like any other associated type: the memory model defines which type this is *actually* going to be.

wip/memory-interface.md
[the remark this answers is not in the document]
RalfJung: Yeah I was going to do this once we have more than a single item.

wip/value-domain.md
RalfJung: I meant `Byte` as in what the memory-interface defines. `BigInt` is a mathematical integer, as the comment indicates. So when you cast a ptr to an int, you get a mathematical integer and that is represented via the `Int` variant.

`RawBag` is for unions.

wip/memory-interface.md
gnzlbg: If it were an associated type, it would be a `type Pointer = ...;` field. But making it an input type parameter gives the impression that one can instantiate this with different Pointer types, for the same memory model.

RalfJung: Oh. The intention is to instantiate this with `Memory::Pointer` but we cannot do that yet because we need to define `Byte` to even define the trait.

I will try to clarify.

wip/value-domain.md
RalfJung: Good question.

I don't think that will change anything about `Value`, but it is worth giving an example representation relation for it.

wip/value-domain.md
RalfJung: I added a note saying `RawBag` is for unions, and hope this answers your question.

RalfJung:
Okay, I think I responded to all comments and edited the document accordingly. Thanks a ton for all your feedback!

wip/memory-interface.md
```diff
@@ -0,0 +1,156 @@
...
+[provenance]: https://github.com/rust-lang/unsafe-code-guidelines/blob/master/reference/src/glossary.md#pointer-provenance
+
+## Bytes
+
+The unit of communication between the memory model and the rest of the program is a *byte*.
+Again, the question of "what is a byte" is not as trivial as it might seem; beyond `u8` values we have to represent `Pointer`s and [uninitialized memory][uninit].
+We define the `Byte` type as follows, where `Pointer` will later be instantiated with the `Memory::Pointer` associated type.
+
+```rust
+enum Byte<Pointer> {
```
RalfJung: Maybe `Byte` is not a good name for this, as it can be viewed as re-defining the term. I am basically starting from "a byte is what gets stored at a memory address", but other people might think "a byte is 8 bits" or "a byte is an integer in `0..256`". and then this type here makes no sense. It turns out that matching all these definitions at once is just not possible, though.

RalfJung:
It's been two weeks, and this is just a WIP document. Any objections to merging?

Centril:
Let's do it!

PR #180, opened 2019-07-25
glossary: define place, value, representation

RalfJung:
As suggested during the meeting, moving this from https://github.com/rust-lang/unsafe-code-guidelines/pull/175 into its own PR.

I think I resolved the feedback I got, except for what I will now add again as comments here.

Cc @gnzlbg @Lokathor @rkruppe @Centril

### DIFF
--- a/reference/src/glossary.md
+++ b/reference/src/glossary.md
@@ -181,12 +181,12 @@ requirement of 2.
 
 A *place* (called "lvalue" in C and "glvalue" in C++) is the result of computing a [*place expression*][place-value-expr].
 A place is basically a pointer (pointing to some location in memory, potentially carrying [provenance](#pointer-provenance)), but might contain more information such as size or alignment (the details will have to be determined as the Rust Abstract Machine gets specified more precisely).
-A place has a type.
-Places cannot be "stored" in memory, only [values](#value) can.
+A place has a type, indicating the type of [values](#value) that it stores.
+Places cannot be "stored" in memory, only values can.
 
 The key operations on a place are:
 * storing a [value](#value) of the same type in it (when it is used on the left-hand side of an assignment),
-* turning it into a [pointer value](#value) (when it is used inside `&expr`),
+* turning it into a [pointer value](#value) (when it is used inside `&expr`), which is also the only way to "store" a place,
 * and loading a [value](#value) of the same type from it (through the place-to-value coercion).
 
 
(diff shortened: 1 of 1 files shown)


reference/src/glossary.md
```diff
@@ -129,10 +129,44 @@ niches. For example, the "all bits uninitialized" is an invalid bit-pattern for
...
 
 ### TODO
 
 * *tag*
-* *rvalue*
-* *lvalue*
-* *representation*
+
+[value-domain]: https://github.com/rust-lang/unsafe-code-guidelines/tree/master/wip/value-domain.md
+[place-value-expr]: https://doc.rust-lang.org/reference/expressions.html#place-expressions-and-value-expressions
```
RalfJung: @gnzlbg suggested to copy the definition of "place expression" and "value expression" from the reference, but IMO linking to them makes much more sense.

reference/src/glossary.md
```diff
@@ -129,10 +129,44 @@ niches. For example, the "all bits uninitialized" is an invalid bit-pattern for
...
+A *value* (called "value of the expression" or "rvalue" in C and "prvalue" in C++) is what gets stored in a [place](#place), and also the result of computing a [*value expression*][place-value-expr].
+A value has a type, and it denotes the abstract mathematical concept that is represented by data in our programs.
+For example, a value of type `u8` is a mathematical integer in the range `0..256`.
+Values can be (according to their type) turned into a list of bytes, which is called a [representation](#representation) of the value.
+Values are ephemeral; they arise during the computation of an instruction but are only ever persisted in memory through their representation.
+(This is comparable to how run-time data in a program is ephemeral and is only ever persisted in serialized form.)
+
+#### Representation (relation)
+
+A *representation* of a [value](#value) is a list of bytes that is used to store or "represent" that value in memory.
```
gnzlbg: By "byte" you probably mean here something more than just 8-bits. 

The other PR defines that term, so it does not need to be done here, but maybe add a (TODO: define byte) here somewhere that gets fixed at some point ?

Lokathor:
Looks good. I'm going to unsubscribe from this but feel free to ping me once this glossary entry is added and I'll do that other PR where all the entries are put in alphabetical order and wrapped to 80 columns.

reference/src/glossary.md
RalfJung: I think this definition makes sense without precisely defining what a "byte" is -- even if the reader thinks "byte == `u8`", there is useful information here. And precisely defining `Byte` will probably be a non-trivial achievement as part of a language spec effort.

gnzlbg: Maybe just add `byte` to the TODO list below ?

With this PR which mentions provenance and the PR defining provenance "byte" makes one wonder where is provenance in the representation.

reference/src/glossary.md
gnzlbg: I suggested doing that for completeness but the link is ok. If we end up needing to modify those definitions slightly we can do so in the reference proper.

reference/src/glossary.md
RalfJung: I don't like that TODO list, IMO issues are better. I will open an issue.

RalfJung:
@ubsan I think you also have opinions on this.

PR #187, opened 2019-08-02
Stacked Borrows: update for shallow retagging

RalfJung:
Matches https://github.com/rust-lang/miri/pull/872


RalfJung:
The PR landed in Miri, so I am going to also land this here.

Issue #95, opened 2019-03-07
🚲🏠 Renaming validity and safety?


Centril:
It has been noted on several occasions, e.g. https://github.com/rust-lang/rust/issues/53491#issuecomment-470496252 and https://github.com/rust-lang/rust/issues/53491#issuecomment-470504576 that "valid" is prone to confusion with "safe" due to the way the former sounds. In particular, it raises the question "Valid with respect to *what*"?

Thus it is probably a good idea to rename one or the other into something else.

@RalfJung suggested *"__initialization__ invariant"* (because any initialized `T` must satisfy such invariants) as a replacement of "validity invariant" while I suggested *"__machine__ invariant"* (because it signifies the invariants the abstract machine requires of a `T`). If we rename validity to either of those then "safety" might not need renaming; but if it does, "type system invariant" may be one candidate.

RalfJung:
One reason why I suggested "initialziation invariant" is because it connects with `MaybeUninit`.

`MaybeNoMachine` doesn't really roll very well...

RalfJung:
Also note that the validity invariant only applies to data that the compiler considers initialized -- this has been a point of confusion several times. Having something with "initialization" in its name should help clarify that.

RalfJung:
In the `MaybeUninit` docs I am talking about the "initialization invariant". But I have since regretted that choice, because e.g. a `MaybeUninit<u32>` is actually "initialized" in uninitialized memory -- and we might end up actually allowing the same for `u32`.

Ixrec:
On one or two past threads [I've suggested that we use "sound" to mean "has defined behavior" and "unsound" to mean "is UB"](https://internals.rust-lang.org/t/unsafe-code-annotations/9239/11?u=ixrec), and I still think that's a good idea. So by analogy, I'd suggest renaming the "validity invariant" to the "**soundness invariant**". At least to me, that makes it immediately obvious what it's supposed to mean: if you violate this, then your code is broken, full stop.

Regardless of which particular word we choose, I do think that it's strongly desirable that "X-ness invariant" corresponds to "if your code is X, it does not have UB". I would not be surprised if "sound" and "unsound" have different connotations to other people and we have to come up with something else, _but_ if we end up agreeing on (for example) "validity invariant" for this, I believe we should also agree to start using "valid" with the very specific meaning of "does not have UB".

[...]

PR #190, opened 2019-08-05
define (un)sound and UB

RalfJung:
Cc https://github.com/rust-lang/unsafe-code-guidelines/issues/95

### DIFF
--- a/reference/src/glossary.md
+++ b/reference/src/glossary.md
@@ -146,6 +146,11 @@ Moreover, such unsafe code must not return a non-UTF-8 string to the "outside" o
 To summarize: *Data must always be valid, but it only must be safe in safe code.*
 For some more information, see [this blog post](https://www.ralfj.de/blog/2018/08/22/two-kinds-of-invariants.html).
 
+#### Soundness (of code / of a library)
+
+We say that a library (can be an individual function) is *sound* if it is impossible for safe code to cause Undefined Behavior using its public API.
+Conversely, the library is *unsound* if safe code *can* cause Undefined Behavior.
+
 #### Layout
 
 The *layout* of a type defines its size and alignment as well as the offsets of its subobjects (e.g. fields of structs/unions/enum/... or elements of arrays).
--- a/reference/src/glossary.md
+++ b/reference/src/glossary.md
@@ -129,7 +129,7 @@ fn main() { unsafe {
 The *safety* invariant is an invariant that safe code may assume all data to uphold.
 This invariant is used to justify which operations safe code can perform.
 The safety invariant can be temporarily violated by unsafe code, but must always be upheld when interfacing with unknown safe code.
-It is not relevant when arguing whether some *program* has UB, but it is relevant when arguing whether some code safely encapsulates its unsafety -- IOW, it is relevant when arguing whether some *library* is [sound][soundness].
+It is not relevant when arguing whether some *program* has UB, but it is relevant when arguing whether some code safely encapsulates its unsafety -- in other words, it is relevant when arguing whether some *library* is [sound][soundness].
 
 In terms of code, some data computed by `TERM` (possibly constructed from some `arguments` that can be *assumed* to satisfy the safety invariant) is valid at type `T` if and only if the following library function can be safely exposed to arbitrary (safe) code as part of the public library interface:
 ```rust,ignore
@@ -155,7 +155,7 @@ In return, the compiler promises to compile the code in a way that the final pro
 If it turns out the program *does* have undefined behavior, the contract is void, and the program produced by the compiler is essentially garbage (in particular, it is not bound by any specification; the program does not even have to be well-formed executable code).
 
 In Rust, the [Nomicon](https://doc.rust-lang.org/nomicon/what-unsafe-does.html) and the [Reference](https://doc.rust-lang.org/reference/behavior-considered-undefined.html) both have a list of behavior that the language considers undefined.
-Rust promises that safe code cannot cause Undefined Behavior---it takes the burden of this contract on itself.
+Rust promises that safe code cannot cause Undefined Behavior---the compiler and authors of unsafe code takes the burden of this contract on themselves.
 For unsafe code, however, the burden is still on the programmer.
 
 Also see: [Soundness][soundness].
--- a/reference/src/glossary.md
+++ b/reference/src/glossary.md
@@ -148,8 +148,8 @@ For some more information, see [this blog post](https://www.ralfj.de/blog/2018/0
 
 #### Soundness (of code / of a library)
 
-We say that a library (can be an individual function) is *sound* if it is impossible for safe code to cause Undefined Behavior using its public API.
-Conversely, the library is *unsound* if safe code *can* cause Undefined Behavior.
+We say that a library (or an individual function) is *sound* if it is impossible for safe code to cause Undefined Behavior using its public API.
+Conversely, the library/function is *unsound* if safe code *can* cause Undefined Behavior.
 
 #### Layout
 
--- a/reference/src/glossary.md
+++ b/reference/src/glossary.md
@@ -129,7 +129,7 @@ fn main() { unsafe {
 The *safety* invariant is an invariant that safe code may assume all data to uphold.
 This invariant is used to justify which operations safe code can perform.
 The safety invariant can be temporarily violated by unsafe code, but must always be upheld when interfacing with unknown safe code.
-It is not relevant when arguing whether some *program* has UB, but it is relevant when arguing whether some code safely encapsulates its unsafety -- IOW, it is relevant when arguing whether some *library* is [sound][soundness]
+It is not relevant when arguing whether some *program* has UB, but it is relevant when arguing whether some code safely encapsulates its unsafety -- IOW, it is relevant when arguing whether some *library* is [sound][soundness].
 
 In terms of code, some data computed by `TERM` (possibly constructed from some `arguments` that can be *assumed* to satisfy the safety invariant) is valid at type `T` if and only if the following library function can be safely exposed to arbitrary (safe) code as part of the public library interface:
 ```rust,ignore
--- a/reference/src/glossary.md
+++ b/reference/src/glossary.md
@@ -146,10 +146,28 @@ Moreover, such unsafe code must not return a non-UTF-8 string to the "outside" o
 To summarize: *Data must always be valid, but it only must be safe in safe code.*
 For some more information, see [this blog post](https://www.ralfj.de/blog/2018/08/22/two-kinds-of-invariants.html).
 
+#### Undefined Behavior
+[ub]: #undefined-behavior
+
+*Undefined Behavior* is a concept of the contract between the Rust programmer and the compiler:
+The programmer promises that the code exhibits no undefined behavior.
+In return, the compiler promises to compile the code in a way that the final program does on the real hardware what the source program does according to the Rust Abstract Machine.
+If it turns out the program *does* have undefined behavior, the contract is void, and the program produced by the compiler is essentially garbage (in particular, it is not bound by any specification; the program does not even have to be well-formed executable code).
+
+In Rust, the [Nomicon](https://doc.rust-lang.org/nomicon/what-unsafe-does.html) and the [Reference](https://doc.rust-lang.org/reference/behavior-considered-undefined.html) both have a list of behavior that the language considers undefined.
+Rust promises that safe code cannot cause Undefined Behavior---it takes the burden of this contract on itself.
+For unsafe code, however, the burden is still on the programmer.
+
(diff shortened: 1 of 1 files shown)


RalfJung:
There's already [blog posts out there](https://iandouglasscott.com/2019/07/26/rust-safe-and-unsafe-as-theorems-and-axioms/) that use this term otherwise -- but they also [quickly get comments correcting that](https://www.reddit.com/r/rust/comments/ci9mak/rust_safe_and_unsafe_as_theorems_and_axioms/ev68dbr/), so I think the term isn't "burnt" yet. We just have to make sure we push this as consistent terminology in enough places.

Are there chapters of the Reference/Nomicon that would fit for this?

RalfJung:
Cc https://github.com/rust-lang-nursery/reference/pull/647

reference/src/glossary.md
```diff
@@ -146,6 +146,11 @@ Moreover, such unsafe code must not return a non-UTF-8 string to the "outside" o
 To summarize: *Data must always be valid, but it only must be safe in safe code.*
 For some more information, see [this blog post](https://www.ralfj.de/blog/2018/08/22/two-kinds-of-invariants.html).
 
+#### Soundness (of code / of a library)
+
+We say that a library (can be an individual function) is *sound* if it is impossible for safe code to cause Undefined Behavior using its public API.
```
Lokathor: The part in parens feels awkward. How about "can be" changing to "or"

RalfJung: Like this?

Lokathor: Yes, excellent.

reference/src/glossary.md
```diff
@@ -146,6 +146,11 @@ Moreover, such unsafe code must not return a non-UTF-8 string to the "outside" o
 To summarize: *Data must always be valid, but it only must be safe in safe code.*
 For some more information, see [this blog post](https://www.ralfj.de/blog/2018/08/22/two-kinds-of-invariants.html).
 
+#### Soundness (of code / of a library)
```
Centril: Should this be structured like so?:

### Soundness

#### Soundness of a type system

#### Soundness of code / a library

### Completeness

Lokathor: Soundness of Code is something that every single Rust programmer who ever writes `unsafe {` must know about.

Soundness of Type System... I don't think I even understand this concept fully, and I probably don't ever _need_ to understand it because I don't edit the type system of Rust. Same thing for Completeness. I honestly don't know what that word refers to, but I probably never need to know unless I'm writing an RFC about Rust's type system itself.

I don't think that Soundness of Type System / Completeness have no place here, but they're definitely the secondary usages.

Also, a mention that it is planned to sort all Glossary entries alphabetically once the churn dies down a bit, so we should not design any entry with the assumption that you'll see any other entry right after it.

RalfJung: @Centril  I'd rather not turn this into a copy of [nlab](https://ncatlab.org/nlab/show/HomePage).

Centril: @RalfJung The dense category theory you find in nlab seems quite far apart from a few words *in plain English* about what soundness in general means. I would at least say something like:

> Assuming that undefined behavior does not arise from misuse of `unsafe` in safe code, the typesystem of Rust is sound if no safe code is accepted by it that can cause undefined behavior.

@Lokathor I think it is clear that the purpose of this repo is more than just unsafe code guidelines at this point. (E.g. discussing the RAM more generally.)

RalfJung: That's not soundness in general though, that's soundness of a particular type system. It is also standard terminology and not terribly relevant to most Rust programmers (unless they want to change the Rust type system).

Also RAM is a really bad acronym I feel. It's already taken.^^

Centril: > That's not soundness in general though, that's soundness of a particular type system.

You wanted it to be more Rust-specific, yes? :)

> It is also standard terminology

Sure, but so is really the understanding in the snippet re. usage of `unsafe` -- it's not unique to Rust and applies to any "believe_me" construct you may find in e.g. Agda ([postulates](https://agda.readthedocs.io/en/v2.5.4/language/postulates.html)), Idris ([`believe_me`](https://github.com/idris-lang/Idris-dev/blob/master/libs/prelude/Builtins.idr#L183)), Coq ([axioms](https://github.com/coq/coq/wiki/CoqAndAxioms)), Haskell ([`unsafeCoerce :: a -> b`](https://hackage.haskell.org/package/base-4.12.0.0/docs/Unsafe-Coerce.html#v:unsafeCoerce)), etc.

> and not terribly relevant to most Rust programmers (unless they want to change the Rust type system).

Or if you want to change the R-AM in ways incompatible with the type system rendering it unsound. -- As for what is relevant to "most Rust programmers", I feel there's going to be plenty of stuff in the reference / UCG docs that most Rust devs don't have to care about (thankfully, as you can write large libraries without a single line of `unsafe`).

> Also RAM is a really bad acronym I feel. It's already taken.^^

R-AM? ;)

RalfJung: > Sure, but so is really the understanding in the snippet re. usage of unsafe -- it's not unique to Rust and applies to any "believe_me" construct you may find in e.g. Agda (postulates), Idris (believe_me), Coq (axioms), Haskell (unsafeCoerce :: a -> b), etc.

Except that even researches very well-versed in these languages are regularly confused by Rust's `unsafe`. No other language really takes it that serious.

A common reaction to Stacked Borrows is "wait, if the problem is only unsafe code, then what's the problem -- just ignore unsafe code, it that code breaks stuff it's their fault".

If we define soundness of a type system, we have to define *tons* of standard PL terminology. It will make the Rust-specific terminology just drown. If you want to start a second "PL glossary for the working Rust programmer", fine, but I think this glossary here should have a a focus on Rust-specific *concepts* (as opposed to instances of well-known generic concepts).

RalfJung: > As for what is relevant to "most Rust programmers", I feel there's going to be plenty of stuff in the reference / UCG docs that most Rust devs don't have to care about (thankfully, as you can write large libraries without a single line of unsafe).

Fair. I meant "most unsafe Rust programmers". The ones for whom this WG is working.

gnzlbg: @Centril 

We use the word "sound" in a couple of places when talking about the safety invariant that safe Rust wrappers over unsafe Rust code should maintain. The definition provided clarifies those uses.

> Should this be structured like so?:

What problems would doing this solve?

RalfJung: @gnzlbg 
> safety invariant 

Good point! I added a link to this new term from the safety invariant description.

Centril: > Except that even researchers very well-versed in these languages are regularly confused by Rust's `unsafe`. No other language really takes it that serious.
> 
> For example, a common reaction to Stacked Borrows is "wait, if the problem is only unsafe code, then what's the problem -- just ignore unsafe code, if that code breaks stuff it's their fault". That's e.g. how Haskell treats `unsafePerformIO`. Similarly, OCaml also makes no effort to even spell out when and how `Obj.magic` is correct to use.

I think this is primarily a consequence of `unsafePerformIO` being relatively rare in comparison to `unsafe` in Rust. That seems expected. However, as one sees from [the Coq page on `Axiom`s](https://github.com/coq/coq/wiki/CoqAndAxioms) it does seem to be taken seriously when the soundness of proofs depends on it.

> If we define soundness of a type system, we have to define _tons_ of standard PL terminology. It will make the Rust-specific terminology just drown in the "noise". If you want to start a second "PL glossary for the working Rust programmer", fine, but I think this glossary here should have a a focus on Rust-specific _concepts_ (as opposed to instances of well-known generic concepts).

I am content to just have something like that one sentence I wrote about what soundness for Rust's type system means; I think it gives an intuition for soundness overall without going into "tons of standard PL terminology". I do not see how that would follow -- it feels like a strawman?

> What problems would doing this solve?

It would help to give an intuition re. ["Unsafe as a Human-Assisted Type System"](https://matklad.github.io/2019/07/25/unsafe-as-a-type-system.html) grounding `unsafe` as a mechanism for "User-land ways to extend the type system".

RalfJung: > I think this is primarily a consequence of unsafePerformIO being relatively rare in comparison to unsafe in Rust. That seems expected. However, as one sees from the Coq page on Axioms it does seem to be taken seriously when the soundness of proofs depends on it.

Proof assistants are very serious about this, that's true, but they also typically don't let axioms cause UB -- if computation happens, they only get stuck.

I have heard contradicting things about `unsafePerformIO` usage in Haskell; about half the Haskell people I meet say its use is not uncommon and you'll find it in production-grade libraries, others say it should never be used in production code.

`Obj.magic` in OCaml... well, it's used by Coq, for example. ;)

RalfJung: @Centril I added some context. I found that hard to do without defining "Undefined Behavior" so I did that, too.

Centril: > I have heard contradicting things about `unsafePerformIO` usage in Haskell; about half the Haskell people I meet say its use is not uncommon and you'll find it in production-grade libraries, others say it should never be used in production code.

As far as I know `unsafePerformIO` and friends are actually used in the base library so it's hard to not transitively depend on it... I venture *a guess* that the more eggheaded users of Haskell would dislike the footguns more than some production users. ^^ Some links [for](https://www.schoolofhaskell.com/user/snoyberg/general-haskell/advanced/evaluation-order-and-state-tokens#unsafeperformio--unsafedupableperformio--inlineperformio--aka-he-who-shall-not-be-named-) the [curious](https://stackoverflow.com/questions/10529284/is-there-ever-a-good-reason-to-use-unsafeperformio?rq=1).

> @Centril I added some context. I found that hard to do without defining "Undefined Behavior" so I did that, too.

This turned out even better than what I wanted. ❤️

reference/src/glossary.md
```diff
@@ -129,7 +129,7 @@ fn main() { unsafe {
 The *safety* invariant is an invariant that safe code may assume all data to uphold.
 This invariant is used to justify which operations safe code can perform.
 The safety invariant can be temporarily violated by unsafe code, but must always be upheld when interfacing with unknown safe code.
-It is not relevant when arguing whether some *program* has UB, but it is relevant when arguing whether some code safely encapsulates its unsafety -- IOW, it is relevant when arguing whether some *library* can be used by safe code to *cause* UB.
+It is not relevant when arguing whether some *program* has UB, but it is relevant when arguing whether some code safely encapsulates its unsafety -- IOW, it is relevant when arguing whether some *library* is [sound][soundness]
```
Centril:
```suggestion
It is not relevant when arguing whether some *program* has UB, but it is relevant when arguing whether some code safely encapsulates its unsafety -- IOW, it is relevant when arguing whether some *library* is [sound][soundness].
```

reference/src/glossary.md
```diff
@@ -146,6 +146,30 @@ Moreover, such unsafe code must not return a non-UTF-8 string to the "outside" o
 To summarize: *Data must always be valid, but it only must be safe in safe code.*
 For some more information, see [this blog post](https://www.ralfj.de/blog/2018/08/22/two-kinds-of-invariants.html).
 
+#### Undefined Behavior
+[ub]: #undefined-behavior
+
+*Undefined Behavior* is a concept of the contract between the Rust programmer and the compiler:
+The programmer promises that the code exhibits no undefined behavior.
+In return, the compiler promises to compile the code in a way that the final program does on the real hardware what the source program does according to the Rust Abstract Machine.
+If it turns out the program *does* have undefined behavior, the contract is void, and the program produced by the compiler is essentially garbage (in particular, it is not bound by any specification; the program does not even have to be well-formed executable code).
```
bjorn3: Where are the nasal demons? :smile:

gnzlbg: Should we go as fast as saying that: 

> If it turns out the program *does* have undefined behavior, the contract is void, and the program is not a Rust program.

? 

That kind of clarifies that we are in the business of defining the behavior of Rust programs, and since those aren't Rust programs, we don't really need to say anything about them. Asking "what's their behavior" is as useful as asking "what's the behavior of trying to compile a C++ `.cpp` file with rustc.

bjorn3: I meant that a reference to the nasal demons joke is missing :)

gnzlbg: (I know - sorry, my reply wasn't addressed at you, but at the snippet of text)

RalfJung: I think "what is a Rust program" is getting into philosophy. Like, it is still syntactically a Rust program.

I think it is easier to try and carve-out the subset of "well-behaved" Rust programs than to exclude some things from the set of "Rust programs" all together.

RalfJung: > I meant that a reference to the nasal demons joke is missing :)

Yeah, that was quite deliberate. ;)

Issue #191, opened 2019-08-06
Glossary formatting


RalfJung:
@Lokathor suggested to sort the glossary alphabetically, and I agree that makes sense. We have some PRs in flight that we should land first though, to avoid conflicts.

I am proposing we also change the headers from level 4 to level 3 (`#### Term` to `### Term`). Level 4 has the same font size as normal text, so it looks the same as normal bold text, which can be confusing.

Open Glossary PRs:
* [ ] https://github.com/rust-lang/unsafe-code-guidelines/pull/153
* [x] https://github.com/rust-lang/unsafe-code-guidelines/pull/190
* [x] https://github.com/rust-lang/unsafe-code-guidelines/pull/192
* [x] https://github.com/rust-lang/unsafe-code-guidelines/pull/161

ehsanmok:
I will do that once they are merged.

RalfJung:
@Lokathor also announced they would do it; please coordinate. :)

Lokathor:
Oh I've got enough trouble wrangling `libm` to behave :3 feel free to let anyone else do the glossary thing.

RalfJung:
All right, @ehsanmok it's yours.

ehsanmok:
@RalfJung any update on this?

RalfJung:
Oh, this got lost it seems...

https://github.com/rust-lang/unsafe-code-guidelines/pull/153 is still open but overall there was not much PR activity recently, so sorting and formatting it now would be fine I think.

RalfJung:
(I might make a PR to just do these edits)

PR #192, opened 2019-08-06
Glossary: de-bold paragraph headers

RalfJung:
Once https://github.com/rust-lang/unsafe-code-guidelines/issues/191 lands we can make them bold again, but currently this looks confusing.

### DIFF
--- a/reference/src/glossary.md
+++ b/reference/src/glossary.md
@@ -65,7 +65,7 @@ The exact form of provenance in Rust is unclear.
 It is also unclear whether provenance applies to more than just pointers, i.e., one could imagine integers having provenance as well (so that pointer provenance can be preserved when pointers are cast to an integer and back).
 In the following, we give some examples if what provenance *could* look like.
 
-**Using provenance to track originating allocation.**
+*Using provenance to track originating allocation.*
 For example, we have to distinguish pointers to the same location if they originated from different allocations.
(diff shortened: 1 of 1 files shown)


RalfJung:
You mean fifth-level headings? Are those a thing?

I also think the equivalent of LaTeX `\paragraph` makes more sense here.

gnzlbg:
> You mean fifth-level headings? Are those a thing?

Yes that's what I meant. I don't think Markdown has any other way to encode these (unless one uses HTML labels?).

> Anyway I think the equivalent of LaTeX \paragraph makes more sense here.

I don't know of any such thing in Markdown or mdbook.

RalfJung:
> I don't know of any such thing in Markdown or mdbook.

I am emulating them with the italics here.

Lokathor:
Instead of starting with two # and then jumping to four #, just start with one and then use two or three for entries and then plus one level for sections within entries.

If a particular entry needs more than one level of sub-section it's complex enough to either get broken up or just made far more terse.

The glossary _doesn't_ have to be exacting if the full definition is very long. Instead, we simply give a short description (enough to get the idea) and a link to the full story in a proper part of the book.

RalfJung:
@Lokathor yes we want to change the global level, see https://github.com/rust-lang/unsafe-code-guidelines/issues/191. But that is blocked, so I wanted to land this in the interim.

RalfJung:
So, any objections for this small tweak to make the glossary more readable *now* without having to re-format everything?

Issue #183, opened 2019-07-31
Define: "padding bytes"


gnzlbg:
We need a technical term for a byte that is not part of the "value representation" of a type, and we need to at least mention how these bytes currently behave, and provide some guarantees. 

For example, a definition that would work today could be:

> A "foo" byte is a byte that is not part of the value representation of a type; "foo" bytes can change without any external program action.
>
> For example, the compiler inserts "foo" bytes in structs to pad struct fields to their alignment, or in unions to make the size of the union a multiple of its alignment.

The "foo bytes can change without any external program action" semantics is what we have today. We could guarantee more, but AFAICT right now we do not. I'd prefer if we could have some basic definition that reflects what happens today first, and punt any discussion about guaranteeing more about these bytes to a follow up discussion. 

Question: can these "foo" bytes occur due to something that is not padding? If not, then we can call them padding bytes. I can imagine a general mechanism to tell the compiler that some byte of a type is a "foo" byte, similar to what we do for the NonNull and similar types today. That could be useful, for example, for mapping a `[u8; N]` to some C layout type, where you know the offsets of the bytes in the array that are padding bytes in C.

RalfJung:
> A "foo" byte is a byte that is not part of the value representation of a type

Let me try to formalize this further: A byte at index `i` is a "foo" byte for type `T` if, for all value `v` and lists of bytes `b` such that `v` and `b` are related at `T`, changing `b` at index `i` to any other byte yields a `b'` such `v` and `b'` are related.

> "foo" bytes can change without any external program action.

I really very strongly dislike magical clauses like this -- primarily because they are quite the opposite of operational. We should give an operational specification for an Abstract Machine that concretely says what happens on every program step. "can change" is not a term that can be used in this context.

And I also think it is unnecessary. In fact, my definition of a ["typed copy"](https://github.com/rust-lang/unsafe-code-guidelines/pull/175) already takes care of padding bytes:

> We can implement `TypedMemory` for any `Memory` as follows:
> * For `typed_read`, read `ty.size()` many bytes from memory, and then determine which value this list of bytes represents. If it does not represent any value, we have UB.
> * For `typed_write`, pick any representation of `val` for `ty`, and call `Memory::write`. If no representation exists, we have UB.
> 
> In particular, this defines the semantics of a "typed copy": when an assignment is executed in Rust (at some type `T`), or when an argument of type `T` is passed to a function or a return value of type `T` returned from a function, this is called a "typed copy".
[...]

Issue #174, opened 2019-07-24
Padding Bytes Guarantees?


jswrenn:
The context of this issue is determining when it is safe for a struct to implement `FromZeros` (_à la_ [`zerocopy::FromBytes`](https://docs.rs/zerocopy/0.2.6/zerocopy/trait.FromBytes.html)), a marker trait implemented for a type `T` _iff_ any sequence of initialized zeroed bytes of length `size_of::<T>()` is a valid instance of `T`. (For such types, `mem::zeroed` is safe!)

Per the [reference](https://doc.rust-lang.org/reference/type-layout.html#representations):
> The representation of a type can change the padding between fields, but does not change the layout of the fields themselves.

Therefore: for a struct `T` to be `FromZeros`, the fields of `T` must also be `FromZeros`.

_I'm trying to determine whether this requirement alone is sufficient._

My understanding is that the padding bytes between fields is expressly undefined (i.e., it can be anything, including uninitialized). **Will `Rust` ever rely on padding having a _particular_ value?**

If so, for a struct `T` to be `FromZeros`, _it must have no padding_, as a `0u8` might not correspond to a valid padding byte for `T`. This poses a very severe limitation on what structs could be `FromZeros` (and, by extension, [`FromBytes`](https://docs.rs/zerocopy/0.2.6/zerocopy/trait.FromBytes.html)).

Lokathor:
I also use a similar unsafe marker to make safe zeroing <https://docs.rs/lokacore/0.1.0/lokacore/trait.Zeroable.html>, so I suppose this applies to me as well.

RalfJung:
Indeed, this should be properly guaranteed somewhere, at some point.

My current understanding is that, *for structs*, your condition is sufficient: in our current implementation, padding bytes can and do change as the struct gets copied, so rustc cannot rely on padding having a particular value.

jswrenn:
> in our current implementation

Part of why I'm cautious is that I can totally imagine optimizations that are enabled by assuming a particular value. For instance, if you assume all padding of a type `T` has a particular value, it can speed up equality checks between two instances of `T` by loading them _in their entirety_ into registers and checking for equality (rather than checking field-by-field).

Making this guarantee will rule out performing that optimization. (That said, I think the benefits of that optimization pale in comparison to the sort of optimizations enabled by  [`zerocopy::FromBytes`](https://docs.rs/zerocopy/0.2.6/zerocopy/trait.FromBytes.html)!)

[...]

PR #195, opened 2019-08-14
define padding with hypothetical Pad type

RalfJung:
Closes https://github.com/rust-lang/unsafe-code-guidelines/issues/183.
Closes https://github.com/rust-lang/unsafe-code-guidelines/issues/174 by saying that a padding byte can be anything.

### DIFF
--- a/reference/src/glossary.md
+++ b/reference/src/glossary.md
@@ -192,7 +192,7 @@ for all values `v` and lists of bytes `b` such that `v` and `b` are related at `
 changing `b` at index `i` to any other byte yields a `b'` such `v` and `b'` are related (`Vrel_T(v, b')`).
 In other words, the byte at index `i` is entirely ignored by `Vrel_T` (the value relation for `T`), and two lists of bytes that only differ in padding bytes relate to the same value(s), if any.
 
-This definition works fine for structs.
+This definition works fine for product types (structs, tuples, arrays, ...).
 The desired notion of "padding byte" for enums and unions is still unclear.
 
 #### Place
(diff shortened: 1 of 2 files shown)


gnzlbg:
Thanks for working on this.

RalfJung:
I have added a remark that for unions and enums, we don't really know what we are doing yet.

reference/src/glossary.md
```diff
@@ -177,6 +177,24 @@ zero-sized type", to refer to zero-sized types with an alignment requirement of
...
+* `Pad` is valid for any byte, i.e., it has the same validity invariant as `MaybeUninit<u8>`.
+* Copying `Pad` ignores the source byte, and writes *any* value to the target byte. Or, equivalently (in terms of Abstract Machine behavior), copying `Pad` marks the target byte as uninitialized.
+
+We can also define padding in terms of the [representation relation]:
+A byte at index `i` is a padding byte for type `T` if,
+for all values `v` and lists of bytes `b` such that `v` and `b` are related at `T` (let's write this `Vrel_T(v, b)`),
+changing `b` at index `i` to any other byte yields a `b'` such `v` and `b'` are related (`Vrel_T(v, b')`).
+In other words, the byte at index `i` is entirely ignored by `Vrel_T` (the value relation for `T`), and two lists of bytes that only differ in padding bytes relate to the same value(s), if any.
+
+This definition works fine for structs.
```
gnzlbg: More generally, this works fine for all product types right? (arrays, tuples, structs, ... SIMD vectors don't have it but..).

RalfJung: Yeah, I always use "struct" as a synonym for all of that. They're all an instance of the same mechanism in terms of layout, after all...

Issue #72, opened 2019-01-10
Validity of function pointers


RalfJung:
Discussing the validity of function pointers.

Clearly, the must be non-NULL.  Since we exclude some bit patterns, we likely also do not want to allow the entire value to be uninitialized.  We could allow some bits to be uninitialized as long as there is at least one bit initialized to 1, but, uh, why?^^

Anything else?  Do fn ptrs have to point to allocated executable memory?  [This discussion](https://rust-lang.zulipchat.com/#narrow/stream/136281-wg-unsafe-code-guidelines/topic/should.20valid.20fn.20be.20callable.20.3F) concludes that the answer ought to be "no", because there's little to no benefit and it would interact badly with unloading shared libraries.

nikomatsakis:
I think we should not require that they point to allocated, executable memory and we should forbid NULL, but otherwise permit any **initialized** value. As you say, I don't see much value to permitting *some* bits to be uninitialized.

RalfJung:
I [recently learned](https://github.com/rust-lang/rust/pull/62474#issuecomment-509965054) that function pointers could, in principle, have alignment requirements. However, currently the alignment seems to be 1 for all supported platforms. The [fn docs](https://doc.rust-lang.org/std/primitive.fn.html) call out the non-NULL property as does [the Nomicon](https://doc.rust-lang.org/nightly/nomicon/what-unsafe-does.html), but they do not call out alignment. Miri also does not check fn ptr alignment as it has no good way to figure out what the requirements for that are, and anway it currently would not do anything.

So should those docs be amended to call out the alignment situation? Which place should they point at if someone wants to figure out the fn ptr alignment for their platform? There is no such thing as `align_of` for this.

Also, should https://doc.rust-lang.org/reference/types/function-pointer.html or https://doc.rust-lang.org/book/ch19-05-advanced-functions-and-closures.html call out the non-NULL property and/or the alignment?

rkruppe:
I don't think we should put any alignment requirement into the validity invariant. LLVM having this notion is no reason to -- in fact, the [function pointer alignment patch](https://reviews.llvm.org/D57335) *fixed* a wrong assumption made by LLVM's optimizations about what happens when you take a pointer to a *known* function. Trying to establish alignment requirements for arbitrary pointers to unknown functions would go far beyond that and is too much headache for too little reward:

[...]

Issue #181, opened 2019-07-26
Define: Byte


RalfJung:
We use the term "byte" in the glossary without defining it. See https://github.com/rust-lang/unsafe-code-guidelines/pull/180#discussion_r307431492 and https://github.com/rust-lang/unsafe-code-guidelines/pull/175 for why that might not be as trivial as it might seem.

RalfJung:
"abyte" or "ambyte" as a short for "abstract (machine) byte" might be a term we could use instead.

gnzlbg:
Considering just real hardware, it isn't uncommon to have to look up what kind of "byte" we are talking about. For example, whether a byte is 8, 16 (DSPs), or N bits wide, whether a byte can be 0xUU (e.g. Itanium), whether a byte can carry provenance information (e.g. CHERI), etc. depends on the machine (e.g. CHERI).

What matters for Rust programs is what a "byte" is in the Rust abstract machine, not what it is in the hardware the binary runs on. So I would just define "byte" with the meaning of it for the abstract machine, and then see if we also need to introduce a different term to specifically talk about "hardware bytes". Is there a part of the spec in which we actually need to talk specifically about "hardware bytes" differentiating them from "abstract machine bytes"?

RalfJung:
> whether a byte can be 0xUU (e.g. Itanium)

Itanium's NaT are not exactly like 0xUU, I think. But I get your point.

The fact remains though that the vast majority of people, and plenty of online material, equates "byte" and `u8`. Some people would say a "byte" is "N bits" where N might in rare cases be different from 8, but that's not enough flexibility for the Rust Abstract Machine byte (at least if you say a bit is 0 or 1, and changing that seems worse than changing "byte"). Itanium, CHERI and friends are extremely niche architectures that barely anyone knows much about.

(FWIW, we could define a Rust Abstract Machine byte to be 8 Rust Abstract Machine bits, but then a "bit" would be 0, 1, U (uninitialized), or a fragment of a pointer.)

gnzlbg:
> The fact remains though that the vast majority of people, and plenty of online material, equates "byte" and u8.

Sure, but isn't this actually correct? If an `u8` can be `0xUU` and carry provenance information, then that's exactly what a byte is in the abstract machine. 

That is, when we define a "byte" in the abstract machine, a `u8` should be defined as a "byte" as well, and the docs for `u8` should explain all of this, such that we end with the spec, API-docs, the book, etc. all in sync.

[...]

PR #197, opened 2019-08-25
RFC for fn ptr validity

RalfJung:
This is a draft RFC based on the consensus in https://github.com/rust-lang/unsafe-code-guidelines/issues/72. I hope to get some feedback from the WG before officially proposing it.

### DIFF
--- a/rfcs/0000-unaligned-fn-ptr.md
+++ b/rfcs/0000-unaligned-fn-ptr.md
@@ -1,0 +1,92 @@
+- Feature Name: `unaligned_fn_ptr`
+- Start Date: 2018-08-25
+- RFC PR: [rust-lang/rfcs#0000](https://github.com/rust-lang/rfcs/pull/0000)
+- Rust Issue: [rust-lang/rust#0000](https://github.com/rust-lang/rust/issues/0000)
+
+# Summary
+[summary]: #summary
+
+Declare that function pointers do not make any alignment or other requirements to be "valid", beyond being non-null.
+
+# Motivation
+[motivation]: #motivation
+
+Rust requires every value that is "produced" to be "valid" according to its type, even if the value is never actually "used" (see [Reference][ref-ub], [Nomicon][nom-ub]).
+Unsafe code needs to know these rules to be sure it does not cause undefined behavior.
+So what should the rules be for function pointers (`fn` types)?
+They definitely must not be null, but is there anything else?
+Possible options for additional requirements include alignment, and some kind of well-formedness of the data it points to ("it must be executable code", or something like that).
+This is currently an open question.
+
+The motivation to settle this question is that we have to say *something* in the Reference and the Nomicon that is at least a reasonable "upper bound" for the validity requirements of function pointers.
+Otherwise, unsafe code creating function pointers is left without rules to follow.
+With [this PR for the Reference](https://github.com/rust-lang-nursery/reference/pull/659), we have reasonable upper bounds for *most* types of Rust (these have not gone through a formal consensus process, but to my knowledge they form an upper bound for every proposal that has been made for validity of any type thus far).
+The only exception is function pointers.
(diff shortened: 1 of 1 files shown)


Lokathor:
[I think this is the right link for the rendered form](https://github.com/RalfJung/unsafe-code-guidelines/blob/fn-ptr/rfcs/0000-unaligned-fn-ptr.md)

Lokathor:
**Very** minor technicality: ARM uses 2 _or_ 4 alignment depending on instruction set.

RalfJung:
> Very minor technicality: ARM uses 2 or 4 alignment depending on instruction set.

That mans 2 is the only alignment we can assume *in general*, right?

rkruppe:
There are ARM chips which don't implement any ISA with 2-byte alignment. If targeting only those, you could assume 4 byte alignment (and also wouldn't need to worry about the LSB being used to indicate Thumb mode), provided you find or define a target triple that is only compatible with those cores and not with any that implement Thumb or Thumb 2. That seems very shaky, tbh.

rfcs/0000-unaligned-fn-ptr.md
```diff
@@ -0,0 +1,92 @@
...
+fn bad() {
+    let x: fn() = unsafe { mem::transmute(0usize) } // This is UB!
+}
+```
+
+Transmuting any non-null value to a function pointer is defined behavior.
+
+```rust
+fn good() {
+    let x: fn() = unsafe { mem::transmute(1usize) } // This not UB.
```
Ixrec:
```suggestion
    let x: fn() = unsafe { mem::transmute(1usize) } // This is not UB.
```

rfcs/0000-unaligned-fn-ptr.md
```diff
@@ -0,0 +1,92 @@
...
+}
+```
+
+Further requirements, outside the scope of this RFC, come into play when the function pointer gets called.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+A function pointer is "valid" (in the sense that it can be produced without causing immediate UB) if and only if it is non-null.
+
```
gnzlbg: Could we add this to the validity chapter of the UCG reference ?  Maybe with the examples from above ?

RalfJung: Sure, I can do that once we merged this PR.

RalfJung:
@rkruppe are you saying we should explain that in the RFC?

rkruppe:
No, I don't think it's that important. To resolve @Lokathor's nit, it would be enough to talk about something like "most popular ARM targets" (cores in recent memory without Thumb or other form of 16 bit instructions are a rarity, Thumb is >20 years old and code density really sucks with only 32 bit instructions).

Lokathor:
Yeah I was thinking something like "ARM targets have a code alignment of 2 or 4, depending on platform details".

I wanted the alignment to seem _even less_ certain, to drive home that non-null is the only realistic guarantee we can give.

RalfJung:
"2 or 4" is just as certain as "2" though in terms of guarantees.

But sure, I can adjust the text for this.

RalfJung:
Reading through this again, I am getting worried about how this relates with the [open questions around provenance in LLVM semantics](https://github.com/rust-lang/unsafe-code-guidelines/issues/181#issuecomment-519860562). Specifically, we might end up having to declare that it is UB to transmute data to `fn()` that has provenance, but does not form a complete pointer (e.g. when the first 4 bytes are from a pointer but the last 4 are not).

(For integers, we have the escape hatch of declaring that to result in `poison`, which is okay if we also permit uninitialized integers. But we cannot permit uninitialized function pointers.)

That would however mean that there are situations where transmuting a non-0 byte string to `fn()` is UB.  The problem is not at all unique to `fn()`; e.g. `bool` has the same problem (if a byte is `0x01` but carries provenance, we might have to make it UB to transmute that to `bool`).

Unfortunately it is entirely unclear what LLVM even *should* do for such cases, let alone what it *will* do. That makes it very hard for us here to make any commitment.

Issue #202, opened 2019-09-11
Clarify the definition of undefined behavior


gnzlbg:
I think we should clarify in the definition of undefined behavior that:

* While this document (and the reference) often call out the behavior that is undefined, the behavior that is not explicitly defined in these documents is also undefined.

* Undefined behavior cannot be assumed to remain undefined. 

To expand on this second point, we currently say that creating a `&mut T` with value `0` is undefined behavior, but we guarantee that this will always be undefined behavior (because the type has a niche), and therefore, this code is sound:

```rust
fn foo(x: &mut T) { unsafe {
   if transmute(x) == 0_usize {
      *std::ptr::null() // never invokes UB
   }
}}
```

Now follow me in this thought experiment, and imagine that, instead of guaranteeing that `&mut T` are never `0`, we were just to say that creating a `&mut T` with value `0` is UB.

Would the example above be sound? I don't think so. If that were the only thing we say about `&mut T` with value `0`, we could, in a future release of Rust, make this undefined behavior defined in such a way that would make that code unsound. For example, we could say that creating a `&mut T` with value `0` is ok, but dereferencing it panics. 

That is, while the compiler can assume that undefined behavior never happens when optimizing Rust code, users cannot assume that undefined behavior will remain undefined when writing Rust code.

RalfJung:
> While this document (and the reference) often call out the behavior that is undefined, the behavior that is not explicitly defined in these documents is also undefined.

I think that is one of the worst possible definitions of UB.
IMO, it is a spec bug if the spec does not cover a case. That's just like having a non-exhaustive `match`. A mathematical definition has to cover all cases, otherwise it is not a valid definition. The same goes for a proper spec.

gnzlbg:
So what do we call behavior not defined in the spec then ?

gnzlbg:
Or are you suggesting that the spec is not valid until it becomes complete ?

RalfJung:
"Not yet specified behavior"? It's basically a white spot on the map. That's very different from UB, for which there should be a precise (and conscious) definition. UB is where we *know* there are dragons; a white spot is where there might be dragons or not.

RalfJung:
> Or are you suggesting that the spec is not valid until it becomes complete ?

Not sure what you mean by "valid", but is certainly not complete until there are no more missing cases.

gnzlbg:
> "Not yet specified behavior"? 

This sounds good to me. 

[...]

PR #203, opened 2019-09-11
Update definition of UB in the glossary

gnzlbg:
Partially addresses #202 .

### DIFF
--- a/reference/src/glossary.md
+++ b/reference/src/glossary.md
@@ -150,7 +150,7 @@ For some more information, see [this blog post](https://www.ralfj.de/blog/2018/0
 [ub]: #undefined-behavior
 
 *Undefined Behavior* (UB) is the behavior for which the Rust language provides no guarantees. Rust programs that do **not** exhibit *undefined behavior* execute according to the semantics of the Rust Abstract Machine on real hardware.
-
+Rust imposes no requirements on programs that exhibit undefined behavior; these programs are not bound by any specification.
 Rust promises that *safe* Rust is [sound][soundness], that is, that it does not exhibit *undefined behavior*. `unsafe` Rust code that allows *safe* Rust code to exhibit *undefined behavior* is [unsound][soundness].
 
 A list of behavior considered undefined is available in the [Rust Language Reference](https://doc.rust-lang.org/reference/behavior-considered-undefined.html).
--- a/reference/src/glossary.md
+++ b/reference/src/glossary.md
@@ -159,7 +159,7 @@ A list of behavior considered undefined is available in the [Rust Language Refer
 > If the source code does not exhibit UB, the implementation promises to compile it to a program that will do on real hardware what the source code says it should do on the Rust Abstract Machine. 
 > However, if the source code exhibits UB, this contract is void, and the implementation can do *anything*. 
 >
-> When this document says that "the behavior of `X` is undefined" it is saying that "there are no guarantees about what the behavior of `X` is". 
+> When this document says that "the behavior of `X` is undefined" that should be read as "there are no guarantees about what the behavior of `X` is". 
 > It could fail to compile, compile but not produce an executable binary, produce an executable binary that does garbage, or even produce an executable binary that appears to do something reasonable.
 > It can also do something different any time. 
 >
--- a/reference/src/glossary.md
+++ b/reference/src/glossary.md
@@ -164,7 +164,7 @@ A list of behavior considered undefined is available in the [Rust Language Refer
 > It can also do something different every time. 
 >
 > It is backward compatible to change the specification from making no guarantees (UB) to guaranteeing something (not UB). 
-> That is, Rust source code is not allowed to rely on UB being UB forever and therefore something that cannot ever happen, because that can change. 
+> That is, Rust source code is not allowed to rely on UB being UB forever. 
 
 #### Soundness (of code / of a library)
 [soundness]: #soundness-of-code--of-a-library
--- a/reference/src/glossary.md
+++ b/reference/src/glossary.md
@@ -149,7 +149,7 @@ For some more information, see [this blog post](https://www.ralfj.de/blog/2018/0
 #### Undefined Behavior
 [ub]: #undefined-behavior
 
-*Undefined Behavior* (UB) is not bound by any specification and this document imposes no requirements on Rust programs that exhibit *undefined behavior*. Rust programs that do **not** exhibit *undefined behavior* preserve the semantics of the Rust Abstract Machine.
+*Undefined Behavior* (UB) is the behavior for which the Rust language imposes no requirements and provides no guarantees. Rust programs that do **not** exhibit *undefined behavior* preserve the semantics of the Rust Abstract Machine.
 
 Rust promises that *safe* Rust is [sound][soundness], that is, that it does not exhibit *undefined behavior*. `unsafe` Rust code that allows *safe* Rust code to exhibit *undefined behavior* is [unsound][soundness].
 
--- a/reference/src/glossary.md
+++ b/reference/src/glossary.md
@@ -149,23 +149,15 @@ For some more information, see [this blog post](https://www.ralfj.de/blog/2018/0
 #### Undefined Behavior
 [ub]: #undefined-behavior
 
-*Undefined Behavior* (UB) is the behavior for which the Rust language provides no guarantees. Rust programs that do **not** exhibit *undefined behavior* execute according to the semantics of the Rust Abstract Machine on real hardware.
-Rust imposes no requirements on programs that exhibit undefined behavior; these programs are not bound by any specification.
+Undefined Behavior (UB) is an error condition of the Rust Abstract Machine.
+The Rust language provides no guarantees for programs that can reach this error condition.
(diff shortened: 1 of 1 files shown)


reference/src/glossary.md
```diff
@@ -149,16 +149,23 @@ For some more information, see [this blog post](https://www.ralfj.de/blog/2018/0
...
-
-Also see: [Soundness][soundness].
+*Undefined Behavior* (UB) is the behavior for which the Rust language provides no guarantees. Rust programs that do **not** exhibit *undefined behavior* execute according to the semantics of the Rust Abstract Machine on real hardware.
+Rust imposes no requirements on programs that exhibit undefined behavior; these programs are not bound by any specification.
+
+Rust requires that *safe* Rust is not able to exhibit *undefined behavior* (see [soundness][soundness]). It is up to the implementation and `unsafe` Rust code to ensure that this is the case.
+
+A list of behavior considered undefined is available in the [Rust Language Reference](https://doc.rust-lang.org/reference/behavior-considered-undefined.html).
+
+> **Note:** In practice, UB is a contract between the Rust source code and the implementation.
```
RalfJung: The implementation of what? I think this is your C++ standard mode showing. ;)

I find "compiler" much more understandable.

reference/src/glossary.md
```diff
@@ -149,16 +149,23 @@ For some more information, see [this blog post](https://www.ralfj.de/blog/2018/0
...
-Rust promises that safe code cannot cause Undefined Behavior---the compiler and authors of unsafe code takes the burden of this contract on themselves.
-For unsafe code, however, the burden is still on the programmer.
-
-Also see: [Soundness][soundness].
+*Undefined Behavior* (UB) is the behavior for which the Rust language provides no guarantees. Rust programs that do **not** exhibit *undefined behavior* execute according to the semantics of the Rust Abstract Machine on real hardware.
+Rust imposes no requirements on programs that exhibit undefined behavior; these programs are not bound by any specification.
+
+Rust requires that *safe* Rust is not able to exhibit *undefined behavior* (see [soundness][soundness]). It is up to the implementation and `unsafe` Rust code to ensure that this is the case.
+
+A list of behavior considered undefined is available in the [Rust Language Reference](https://doc.rust-lang.org/reference/behavior-considered-undefined.html).
```
RalfJung: Please keep the Nomicon link, I think it is useful to read this in both styles.

As long as both lists exist, I think we should link them both here.

reference/src/glossary.md
```diff
@@ -149,16 +149,23 @@ For some more information, see [this blog post](https://www.ralfj.de/blog/2018/0
...
-If it turns out the program *does* have undefined behavior, the contract is void, and the program produced by the compiler is essentially garbage (in particular, it is not bound by any specification; the program does not even have to be well-formed executable code).
-
-In Rust, the [Nomicon](https://doc.rust-lang.org/nomicon/what-unsafe-does.html) and the [Reference](https://doc.rust-lang.org/reference/behavior-considered-undefined.html) both have a list of behavior that the language considers undefined.
-Rust promises that safe code cannot cause Undefined Behavior---the compiler and authors of unsafe code takes the burden of this contract on themselves.
-For unsafe code, however, the burden is still on the programmer.
-
-Also see: [Soundness][soundness].
+*Undefined Behavior* (UB) is the behavior for which the Rust language provides no guarantees. Rust programs that do **not** exhibit *undefined behavior* execute according to the semantics of the Rust Abstract Machine on real hardware.
+Rust imposes no requirements on programs that exhibit undefined behavior; these programs are not bound by any specification.
```
RalfJung: So, I like your start, because it is more concrete than mine. However, I think it is very important that the term "contract" comes up in the first paragraph already. That is the mindset needed to understand UB, it should not be an afterthought degraded to a "note".
Also, your text starts by explaining things about the negation, which is confusing.

But I realized that I am not happy how the Abstract Machine comes in in either version -- after all, it is the abstract machine that *defines* UB; the text makes it sound like the Abstract Machine only enters the picture when there is no UB.

So here's another shot, merging your and my versions:

> *Undefined Behavior* (UB) is an error condition of the Rust Abstract Machine.
> The Rust language provides no guarantees for programs that can reach this error condition.
> As part of the contract between the Rust programmer and the compiler, is thus the obligation of the programmer to ensure that their code, when run according to the rules of the Rust Abstract Machine, does not exhibit undefined behavior.
> In return, the compiler guarantees that programs that do **not** exhibit undefined behavior execute on real hardware according to the semantics of the Rust Abstract Machine.
> Rust imposes no requirements on the behavior of programs that exhibit undefined behavior; these programs are not bound by any specification (in particular, the program produced by the compiler is essentially garbage, it does not even have to be well-formed executable code).

reference/src/glossary.md
```diff
@@ -149,16 +149,23 @@ For some more information, see [this blog post](https://www.ralfj.de/blog/2018/0
...
 [ub]: #undefined-behavior
 
-*Undefined Behavior* is a concept of the contract between the Rust programmer and the compiler:
-The programmer promises that the code exhibits no undefined behavior.
-In return, the compiler promises to compile the code in a way that the final program does on the real hardware what the source program does according to the Rust Abstract Machine.
-If it turns out the program *does* have undefined behavior, the contract is void, and the program produced by the compiler is essentially garbage (in particular, it is not bound by any specification; the program does not even have to be well-formed executable code).
-
-In Rust, the [Nomicon](https://doc.rust-lang.org/nomicon/what-unsafe-does.html) and the [Reference](https://doc.rust-lang.org/reference/behavior-considered-undefined.html) both have a list of behavior that the language considers undefined.
-Rust promises that safe code cannot cause Undefined Behavior---the compiler and authors of unsafe code takes the burden of this contract on themselves.
-For unsafe code, however, the burden is still on the programmer.
```
RalfJung: I like the term "burden" here, but if y'all agree with @gnzlbg I am fine removing it.

reference/src/glossary.md
```diff
@@ -149,16 +149,23 @@ For some more information, see [this blog post](https://www.ralfj.de/blog/2018/0
...
+> **Note:** In practice, UB is a contract between the Rust source code and the implementation. 
+> If the source code does not exhibit UB, the implementation promises to compile it to a program that will do on real hardware what the source code says it should do on the Rust Abstract Machine. 
+> However, if the source code exhibits UB, this contract is void, and the implementation can do *anything*. 
+>
+> When this document says that "the behavior of `X` is undefined" that should be read as "there are no guarantees about what the behavior of `X` is". 
+> It could fail to compile, compile but not produce an executable binary, produce an executable binary that does garbage, or even produce an executable binary that appears to do something reasonable.
+> It can also do something different every time. 
+>
+> It is backward compatible to change the specification from making no guarantees (UB) to guaranteeing something (not UB). 
+> That is, Rust source code is not allowed to rely on UB being UB forever.
```
RalfJung: Why is that part just a "note"?

reference/src/glossary.md
rkruppe: I also think this is a more useful framing than the closest equivalent that's in the new text.

reference/src/glossary.md
gnzlbg: Maybe. Is it only the compiler ? It is also a contract with the implementation of the abstract machine (e.g. miri, the hardware, etc.).

reference/src/glossary.md
gnzlbg: Currently the Nomicon link is way outdated. I'll keep both links, but maybe the nomicon should be updated to just refer to the language reference at some point.

reference/src/glossary.md
RalfJung: I think so far in the UCG documents we have always talked about "the compiler"?
I agree the proper spec might have to be more careful with terminology. But we should be consistent at least. So probably best to keep saying "the compiler" and file an issue to maybe `s/compiler/Rust implementation/` or so?

reference/src/glossary.md
RalfJung: I just (over the last few weeks) updated both of them, and the lists should actually be in sync. Why do you say it is outdated?

> maybe the nomicon should be updated to just refer to the language reference at some point.

Agreed, but for now this is the situation we got.

gnzlbg: > I just (over the last few weeks) updated both of them, and the lists should actually be in sync. Why do you say it is outdated?

I didn't realize that the nomicon PR was already merged.

gnzlbg:
@rkruppe @RalfJung is this better ?

RalfJung:
Considering https://github.com/rust-lang/unsafe-code-guidelines/issues/202#issuecomment-543673604, I think putting the error state first here does not reflect our current consensus any more.

gnzlbg:
No, i'm going to try to come up with a way to explain the two different types of UB that are discussed there.

RalfJung:
While thinking about this I came up with the following introduction, which I think is a nice starting point:

> The Rust compiler guarantees correct compilation only under certain assumptions that must be satisfied by the programmer. Any violation of these assumptions is called *Undefined Behavior*.

RalfJung:
Closing due to inactivity, though updating this definition along the lines of https://github.com/rust-lang/unsafe-code-guidelines/issues/202#issuecomment-543673604 is probably still a good idea.

PR #207, opened 2019-09-28
Change note about usize incompatibility with C type

comex:
...to mention `size_t` rather than `unsigned`.

Neither `size_t` nor `unsigned` is guaranteed to be compatible with `usize`, so we could have the note mention both.  But as I see it, when experienced C programmers see `usize` for the first time, they will think of `size_t`, since it has a similar name and is used for similar purposes.  At that point they will be unsurprised to hear that `usize` is different from `unsigned` since, on the most common platforms in use, `size_t` is 64-bit while `unsigned` is 32-bit.  As such, the existing warning doesn't add all that much.  On the other hand, I believe `usize` is compatible with `size_t` on all currently supported targets, so the fact that it might be incompatible is much more surprising.  So surprising, in fact, that I think it's worth removing the mention of `unsigned` just to make the warning stand out more.  YMMV...

### DIFF
--- a/reference/src/layout/scalars.md
+++ b/reference/src/layout/scalars.md
@@ -37,14 +37,14 @@ The `isize` and `usize` types are pointer-sized signed and unsigned integers.
 They have the same layout as the [pointer types] for which the pointee is
 `Sized`, and are layout compatible with C's `uintptr_t` and `intptr_t` types.
 
+> **Note**: As a result, Rust's `usize` is **not** necessarily layout
+> compatible with C's `size_t`, since `uintptr_t` may be different from
+> `size_t`.
+
 > **Note**: C99 [7.18.2.4](https://port70.net/~nsz/c/c99/n1256.html#7.18.2.4)
 > requires `uintptr_t` and `intptr_t` to be at least 16-bit wide. All 
 > platforms we currently support have a C platform, and as a consequence,
 > `isize`/`usize` are at least 16-bit wide for all of them.
-
-> **Note**: Rust's `usize` and C's `unsigned` types are **not** equivalent. C's
-> `unsigned` is at least as large as a short, allowed to have padding bits, etc.
-> but it is not necessarily pointer-sized.
 
 > **Note**: in the current Rust implementation, the layouts of `isize` and
 > `usize` determine the following:


reference/src/layout/scalars.md
```diff
@@ -37,15 +37,15 @@ The `isize` and `usize` types are pointer-sized signed and unsigned integers.
 They have the same layout as the [pointer types] for which the pointee is
 `Sized`, and are layout compatible with C's `uintptr_t` and `intptr_t` types.
 
+> **Note**: As a result, Rust's `usize` is **not** necessarily layout
+> compatible with C's `size_t`, since `uintptr_t` may be different from
+> `size_t`.
```
gnzlbg: An example of such an architecture might be CHERI, where `uintptr_t` is 128-bit, but `size_t` is 64-bit. Maybe @parched knows more about that. 

From my experience interacting with `libc` users, a lot of them write code under the assumption that `usize == uintptr_t == size_t`, which is indeed true for AFAIK all architectures that Rust supports today.

nw0: Earlier this year I made some effort to get Rust (just `libcore`) to compile for CHERI. In many ways it "just works", although you have to fudge the 64-bit index size (generated from `ptrtoint`) with 128-bit `usize`, for actual indexing. Then there are the problems like 128-bit indices into `memcpy` intrinsics.

Aside: proper support absolutely requires an index type as well as a pointer size type. I think the latter could be implicit (set only in the data layout string and `target_ptr_width`), although someone will probably cook up some Unsafe code to prove me wrong.

I write this here because there really hasn't been any appropriate place to have this discussion yet.

reference/src/layout/scalars.md
```diff
@@ -37,15 +37,15 @@ The `isize` and `usize` types are pointer-sized signed and unsigned integers.
...
+
 > **Note**: C99 [7.18.2.4](https://port70.net/~nsz/c/c99/n1256.html#7.18.2.4)
 > requires `uintptr_t` and `intptr_t` to be at least 16-bit wide. All 
 > platforms we currently support have a C platform, and as a consequence,
 > `isize`/`usize` are at least 16-bit wide for all of them.
 
-> **Note**: Rust's `usize` and C's `unsigned` types are **not** equivalent. C's
-> `unsigned` is at least as large as a short, allowed to have padding bits, etc.
```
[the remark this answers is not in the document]
RalfJung: No strong opinion from my side. I am not sure removing this information will help anyone though. This is a rather long and somewhat messy list of notes anywhere, maybe it can be organized better?

reference/src/layout/scalars.md
gnzlbg: @nw0 you probably want to open a rust-lang/rust issue about this, since it impacts to many parts of the toolchain, and move the discussion there. We currently use `usize` for indexing and for converting pointers to ints (`ptr as usize`), so I'm not sure how your "implicit" pointer size type is compatible with both.

RalfJung:
@comex looks like we lost track of this *oops* sorry for that.
In case you still want to complete this PR... are we sure this is correct? AFAIK on all platforms that Rust supports, `uintptr_t == size_t`. And in practice people do (have to) rely on `usize` being compatible with `size_t`.

Recent developments around Strict Provenance indicate that for CHERI, `usize` should probably still be 64bit wide, meaning it would actually be compatible with `size_t` but *not* `uintptr_t` on that platform. That would be a special exception, CHERI is not an entirely portable target for Rust and we'd have to find ways to express that. (It's not entirely portable for real-world C code either, since real-world C code makes tons of assumptions about the C integer hierarchy.)

RalfJung:
Closing due to inactivity. Feel free to reopen if you want to get back to this, but note that the UCG book isn't really maintained these days. That said I am not sure where else such information would go so I am still fine with keeping it here.

PR #217, opened 2019-10-21
note on uninhabited structs

RalfJung:
I just realized that our current wording excludes making `(i32, !)` zero-sized. That's currently true and there are good arguments for keeping it that way, but we have no guarantee yet.

### DIFF
--- a/reference/src/layout/structs-and-tuples.md
+++ b/reference/src/layout/structs-and-tuples.md
@@ -72,17 +72,12 @@ struct Foo {
 (In fact, one may use such field names in patterns or in accessor
 expressions like `foo.0`.)
 
-The degrees of freedom the compiler has when computing the layout of a struct or
-tuple is to determine the order of the fields, and the "gaps" (often called
-*padding*) before, between, and after the fields. The layout of these fields
-themselves is already entirely determined by their types, and since we intend to
-allow creating references to fields (`&s.f1`), structs do not have any
-wiggle-room there.
-
-**Note:** This is only true if the struct is inhabited.  For structs like `(i32,
-!)` that do not have a valid inhabitant, the compiler has more freedom.  After
-all, no references to fields can ever be taken.  For example, such structs might
-be zero-sized.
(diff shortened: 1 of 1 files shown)


rkruppe:
Should not be a note IMO (it's not just clarifying something or a digression, it's on the same level as the preceding paragraph) but otherwise LGTM.

reference/src/layout/structs-and-tuples.md
```diff
@@ -79,6 +79,11 @@ themselves is already entirely determined by their types, and since we intend to
 allow creating references to fields (`&s.f1`), structs do not have any
 wiggle-room there.
 
+**Note:** This is only true if the struct is inhabited.  For structs like `(i32,
+!)` that do not have a valid inhabitant, the compiler has more freedom.  After
+all, no references to fields can ever be taken.  For example, such structs might
+be zero-sized.
+
```
gnzlbg: This makes sense and I suppose we could actually make this a guarantee already, but I'm left wondering if this guarantee "propagates" back to `!`. E.g. if `struct A(!)` can be zero-sized, does that mean that `!` can be zero-sized ?

RalfJung:
> Should not be a note IMO (it's not just clarifying something or a digression, it's on the same level as the preceding paragraph) but otherwise LGTM.

Better now?

reference/src/layout/structs-and-tuples.md
RalfJung: `!` is zero-sized... so yes it can be...

gnzlbg: Could we word that as well somewhere? Maybe add a `uninhabited.md` document saying at least that ?

gnzlbg: Ah you reworded things, LGTM.

RalfJung:
@gnzlbg are you okay with merging this?

gnzlbg:
Yes!

PR #220, opened 2019-12-05
Guarantee layout of homogeneous tuples

### LINKED ISSUE #36: Layout of homogeneous structs
From #31: If you have homogeneous structs, where all the `N` fields are of a single type `T`, can we guarantee a mapping to the memory layout of `[T; N]`? How do we map between the field names and the indices? What about zero-sized types?

A specific proposal:

- If all fields have the same type `T` (ignoring zero-sized types), then the layout will be the same as `[T; N]` where `N` is the number of fields
- If the struct is defined with named fields, the mapping from fields to their indices is undefined (so `foo.bar` maps an undefined index)
- If the struct is defined as a tuple struct (or a tuple), then the indices are derived from the definition (so `foo.0` maps to `[0]`)

This is basically because it's convenient but also because it's about the only sensible thing to do (unless you imagine we might want to start inserting padding between random fields or something, which connects to #35).

Note that for tuple structs (and tuples), the ordering of (public) fields is also significant for semver and other purposes -- changing the ordering will affect your clients. The same is not true for named fields and so this proposal attempts to avoid making it true.

gnzlbg:
Not much has happened on the issue #36, but this is how the proposed change would look like.

Closes #36 .

### DIFF
--- a/reference/src/layout/packed-simd-vectors.md
+++ b/reference/src/layout/packed-simd-vectors.md
@@ -45,12 +45,12 @@ That is, two distinct `repr(simd)` vector types that have the same `T` and the
 same `N` have the same size and alignment.
 
 Vector elements are laid out in source field order, enabling random access to
-vector elements by reinterpreting the vector as an array:
+vector elements by reinterpreting the vector as an array or as an homogeneous tuple:
 
 ```rust,ignore
 union U {
    vec: Vector<T, N>,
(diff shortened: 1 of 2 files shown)


rkruppe:
#36 is about structs or aggregates more generally, not (just) tuples. Plus, given the state of discussion (and lack of consensus) in #36 I don't think we should merge this.

PR #222, opened 2019-12-10
Use webarchive link for c17 specification

kngwyu:
Unfortunately, [this link](http://www.open-std.org/jtc1/sc22/wg14/www/abq/c17_updated_proposed_fdis.pdf) looks broken.
I borrowed this web archive link from http://www.iso-9899.info/wiki/The_Standard.


RalfJung:
@gnzlbg do you know a more authoritative/stable location of this standard?

gnzlbg:
@RalfJung no I don't, that's the same source that I used back then.

PR #229, opened 2020-03-20
Prefer ::MAX associated constant over `max_value()`

lzutao:
As `max_value()` is being soft-deprecated now and they will be deprecated
in a next few Rust releases.


RalfJung:
So much better :)

Lokathor:
Oh no, I think clippy has been directing people to max_value for months now XD

PR #230, opened 2020-03-20
Minor raw markdown changes

lzutao:


reference/src/layout/scalars.md
```diff
@@ -22,7 +23,7 @@ taking on the values 1 (`true`) or 0 (`false`).
 ## `char`
 
 Rust char is 32-bit wide and represents an [unicode scalar value]. The alignment
-of `char` is _implementation-defined_.
+of `char` is _implementation-defined_ (Citation needed).
```
lzutao: I wonder that whether `char` alignment is really implementation defined.

RalfJung: Well is it explicitly specified anywhere?

reference/src/layout/scalars.md
```diff
@@ -48,17 +49,17 @@ They have the same layout as the [pointer types] for which the pointee is
...
+>
+> * The maximum size of Rust _allocations_ is limited to `isize::MAX`.
 >   The LLVM `getelementptr` instruction uses signed-integer field offsets. Rust
 >   calls `getelementptr` with the `inbounds` flag which assumes that field
 >   offsets do not overflow,
 >
-> * the maximum number of elements in an array is `usize::MAX` (`[T; N:
->   usize]`. Only ZST arrays can probably be this large in practice, non-ZST
->   arrays are bound by the maximum size of Rust values,
+> * The maximum number of elements in an array is `usize::MAX` (`[T; N: usize]`).
```
RalfJung: This continues the sentence from the previous bullet. I think it was deliberately lower-case.

reference/src/layout/scalars.md
lzutao: It doesn't. But I wonder can we define this information in the Reference book?
And/Or explain why char is implementation defined.

lzutao: Hi, I add citation to the Rust reference.

PR #234, opened 2020-06-08
add case of libstd fix from Stacked Borrows

JOE1994:
This PR adds a new case to the list of
**'adjustments made to libstd (due to Stacked Borrows violation)'**.

A small fix was made to libstd in rust-lang/rust#70479 (back in March).
(Miri reported UB due to Stacked Borrows violation - [link to Miri error log](https://github.com/rust-lang/miri/pull/1225#discussion_r397830221))

Thank you for reviewing :+1:


RalfJung:
Oh, I forgot about this list, good catch!

But we also have a list at https://github.com/rust-lang/miri#bugs-found-by-miri, which is better maintained. Maybe it would make more sense to link there instead? (Looks like this one is also missing from that list, though.)

JOE1994:
> Oh, I forgot about this list, good catch!
> 
> But we also have a list at https://github.com/rust-lang/miri#bugs-found-by-miri, which is better maintained. Maybe it would make more sense to link there instead? (Looks like this one is also missing from that list, though.)

Then I will close this PR and make a PR to Miri :+1:

RalfJung:
@JOE1994 thanks. :)  Can you also do a PR here to link to Miri instead of having a separate list?

PR #247, opened 2020-08-15
note exception to 1-ZST-are-ignored rule

RalfJung:
This is based on what @eddyb [wrote here](https://github.com/rust-lang/rust/pull/73453#issuecomment-674310011).

### DIFF
--- a/reference/src/layout/structs-and-tuples.md
+++ b/reference/src/layout/structs-and-tuples.md
@@ -177,11 +177,15 @@ A struct with only one field has the same layout as that field.
 
 #### Structs with 1-ZST fields
 
-For the purposes of struct layout [1-ZST] fields are ignored.
+For the purposes of struct layout [1-ZST] fields are ignored, *except* when it
+is the final field and could be subject to an unsizing coercion.
 
 In particular, if all but one field are 1-ZST, then the struct is equivalent to
 a [single-field struct][single-field structs].  In other words, if all but one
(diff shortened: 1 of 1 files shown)


eddyb:
I'd hold off on merging this, as at least for the call ABI we could have 1-ZST independence, as per the second half of https://github.com/rust-lang/rust/pull/73453#issuecomment-674310011.

IMO it's more important to point out that ZST fields *still have offsets*, but we don't guarantee anything about them.
(except when field order is guaranteed, I guess? or is `last_field: [T; 0]` a really bad idea?)

RalfJung:
Of course they still have offsets, all fields do.

> The degrees of freedom the compiler has when computing the layout of an inhabited struct or tuple is to determine the order of the fields, and the "gaps" (often called padding) before, between, and after the fields. The layout of these fields themselves is already entirely determined by their types, and since we intend to allow creating references to fields (&s.f1), structs do not have any wiggle-room there.

eddyb:
Can this be closed once https://github.com/rust-lang/rust/pull/73453 lands? @erikdesjardins switched the PR over to fix codegen instead.

RalfJung:
If that lifts the unsizing restriction, then yes, we won't need this docs change any more :)

RalfJung:
https://github.com/rust-lang/rust/pull/73453 landed so let's close this.

PR #279, opened 2021-03-15
Fix typo: &dyn mut -> &mut dyn

Shadlock0133:


RalfJung:
Good catch, thanks. :)

PR #290, opened 2021-06-24
Add missing `repr(C)` to a struct exposed to C

vi:
This struct is supposed to be shared between Rust and C code, but suspiciously lacks `#[repr(C)]`.


RalfJung:
Good catch. :)

PR #332, opened 2022-04-27
reference: add link to github repo

johannst:


reference/book.toml
```diff
@@ -3,3 +3,7 @@ authors = ["The Rust Project Developers"]
 multilingual = false
 src = "src"
 title = "Unsafe Code Guidelines Reference"
+
+[output.html]
+git-repository-url = "https://github.com/rust-lang/unsafe-code-guidelines"
+git-repository-icon = "fa-github"
```
RalfJung, line 9: What does this "icon" thing do?
https://github.com/rust-lang/reference/blob/master/book.toml does not seem to have it.

johannst: It configures the icon, but I just saw that it defaults to `fa-github` (https://rust-lang.github.io/mdBook/format/configuration/renderers.html#html-renderer-options).
I'll remove it.

RalfJung:
Thanks!

PR #368, opened 2022-10-15
Union validity

### LINKED ISSUE #352: What is the safety invariant, if any, for unions?

alercah:
Closes #352.

### DIFF
--- a/reference/src/validity/unions.md
+++ b/reference/src/validity/unions.md
@@ -1,13 +1,243 @@
 # Validity of unions
 
-**Disclaimer**: This chapter is a work-in-progress. What's contained here
-represents the consensus from issue [#73]. The statements in here are not (yet)
-"guaranteed" not to change until an RFC ratifies them.
+**Disclaimer**: This chapter is a work-in-progress.
+What's contained here represents the consensus from [various issues][union
+discussion].
+The statements in here are not (yet) "guaranteed" not to change until an RFC
+ratifies them.
 
-## Validity of unions with zero-sized fields
+[union discussion]: https://github.com/rust-lang/unsafe-code-guidelines/blob/master/active_discussion/unions.md
 
-A union containing a zero-sized field can contain any bit pattern. An example of such
-an union is [`MaybeUninit`].
+**Note**: For ease of reading the examples, the hypothetical type `Padded<T>` is
+used, which behaves identically to `T` except that writing to it clobbers any
+subsequent padding.
+This could actually be accomplished using an overaligned newtype struct, but
+that would make the examples harder to read.
+Additionally, layout of union fields is put in some comments; in these comments,
+the notation `n*pad` means "`n` padding bytes".
 
+## Value model
+
+The possible values of unions are not defined in terms of the values of the
+fields, but rather, a union's possible values are lists of bytes.
+The [representation relation] is trivial in both directions, except for [padding
+bytes] which are uninitialized in all values.
+
+<details><summary><b>Rationale</b></summary>
+
+The following examples must be supported, and therefore impose constraints on union value behaviour. The simplest solution, by far, is to treat the value representation of unions as merely being lists of bytes. While we do not discuss every possible angle here, it should be rapidly clear from just these two examples that any other example is significantly more complicated. Please trust us that every other alternatives that sounds even half-reasonable has been examined and has some critical flaw or another.
+
+**We cannot require all fields, or even only one field, to be valid at all times.**
+
+```rust
+union Padding {
+    left: (Padded<u8>, u16), // [[u8, 1*pad], u16]
+    right: (u16, Padded<u8>), // [u16, [u8, 1*pad]]
+}
+let p = Padding{left: (0, 0)}; // resulting bytes: [0, uninit, 0, 0]
+p.right.1 = 1; // resulting bytes: [0, uninit, 1, uninit]
+fn f(_: Padding) {}
+f(p);
+```
+
+By the end of this example, the resulting union has no valid fields, because every field contains uninit, non-padding bytes. So therefore no field is even fully initialized, despite the fact that `p` was fully initialized as can be witnessed by the fact that the compiler allows it to be moved into `f`. This is all stable, Safe Rust, so these semantics being sound are a hard constraint.
+
+**Unions must preserve provenance.**
+
+```rust
+union Provenance<'a> {
+    raw: *const u32,
+    reference: &'a u32,
+}
+let x: u32;
+let u = Provenance(raw: &x);
+let y = unsafe{ *u.reference };
+```
+
+We must be able to carry provenance between the `raw` and `reference` fields in order for the assignment via `u.reference` to be valid. While this uses Unsafe Rust, this code is "obviously" sound. And therefore the union must be able to maintain provenance between the two pointers---even should the pointers be nested deeply within structs.
+
+[representation relation]: ../glossary.md#representation-relation
+[padding bytes]: ../glossary.md#padding-byte
+
+</details>
+
+### Niches
+
+## Value space for #[repr(C)] and Raw-repr unions
+
+`#[repr(C)]` and [Raw-repr][raw repr] unions can store any byte value, except for [padding bytes] which are always uninitialized.
+
+<details><summary><b>Rationale</b></summary>
+
+The purpose of the Raw-repr is to provide these semantics, which are easy to reason about. Furthermore, C programmers are used to being able to treat unions like bags of bytes, more or less, and Rust programmers are similarly used to the same with `#[repr(C)]` unions. Therefore, they should both accept any arbitrary byte pattern.
+
+</details>
+
+## Possible niche values
+
+The presence of [padding bytes], and writes to individual fields in general, makes niches hard to come by in unions. A niche representation of a union would have to not only be invalid for every single one of its fields, but also impossible to construct in Safe Rust with any combination of writes to any of its fields.
+
+For reprs other than `#[repr(C)]` and the [Raw-repr], values not constructible from safe Rust are consequently [**TBD**][#73] whether or not they are valid. The following example assumes that `#[repr(Rust)]` is not the Raw-repr:
+
+```rust
+#[repr(Rust)]
+union U {
+    a: (u16, u16),
+    b: u32,
+}
+MaybeUninit<U>::uninit().assume_init(); // Unsound: assumes that U can be uninit
+fn get_b(u: U) -> u32 {
+    unsafe { u.b } // Unsound: assumes that U cannot be uninit
+}
+let u: U;
+get_b(u); // Compile error: u is not initialized.
+```
+
+Because all bytes of `U` must be initialized for the value to be valid, and this is enforced by the compiler's initialization checks, it might be tempting to assume that `U`'s bytes must always be defined, but this is not a valid assumption. It is equally invalid, however, to assume that `U`'s bytes can be undefined.
+
+<details><summary><b>Rationale</b></summary>
+
+We have not yet reached consensus on whether or not we wish to leave the door open for the possibility that unions with safe field access, or `#[repr(transparent)]` unions with no ZSTs, contain niches:
+
+```rust
+#[repr(transparent)]
+union U { b: bool };
+assert_eq!(size_of::<Option<U>>(), 1); // Requires a niche, which in turn requires that U must be initialized to be valid.
+```
+
+We are **not** describing this case as unspecified, but instead as TBD."Constructible with Safe Rust" is a poorly-defined and very complex invariant, which falls short of the UCG's goals of easily checked, easy to understand (such as it were) semantics, and therefore we are not comfortable leaving the language in this state on an indefinite basis.
+
+The main saving grace here is that `#[repr(Rust)]` unions are presently nearly impossible to use correctly anyways, because they do not even guarantee fields at offset 0.
+
+</details>
+
+## Validity of sometimes-padding bytes
+
+We can say that a byte is *sometimes padding* for a union `U` if there is *some* inhabited field `f` such that the byte is either padding for `f` or not a part of `f`.
+
+In that case, the byte will be uninitialized in the value `U{f: /* some value */ }`. By the [monotonicity property], therefore, all sometimes-padding bytes can contain any byte value, be it undefined or any bit pattern with any provenance. Likewise, if multiple bytes are padding for the same field, then they can take on any possible combination of byte values between them.
+
+It follows that a union containing an inhabited zero-sized field can contain any bit pattern whatsoever, because all bytes are sometimes-padding bytes. An example of such
+an union is [`MaybeUninit<T>`], which is a union of `T` and `()`.
+
+As per the previous section, however, just because a byte is a sometimes-padding byte does not mean it can always safely be set to uninitialized (or any other value), if this can produce a value not reachable from Safe Rust.
+
+For instance, the following is presently unsound (assuming that `#[repr(Rust)]` is not the [Raw-repr]), even assuming that all fields are placed at offset 0:
+
+```rust
+struct B {
+
+}
+#[repr(Safe)]
+Union u {
+    b: bool, // [bool, 1*pad]
+    u: u16, // [u16]
+}
+let u = U{u: (0xff00, 0}; // resulting bytes: [0xff, 0]
+unsafe { (&mut u.b as *mut u8 as *mut MaybeUninit<u8>).offset(1).write(MaybeUninit::uninit()) }; // resulting bytes: [0xff, uninit]
+```
+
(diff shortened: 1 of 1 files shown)


reference/src/validity/unions.md
```diff
@@ -1,13 +1,243 @@
...
+The possible values of unions are not defined in terms of the values of the
+fields, but rather, a union's possible values are lists of bytes.
+The [representation relation] is trivial in both directions, except for [padding
+bytes][padding byte] which are uninitialized in all values.
+
+<details><summary><b>Rationale</b></summary>
+
+The following examples must be supported, and therefore impose constraints on union value behaviour. The simplest solution, by far, is to treat the value representation of unions as merely being lists of bytes. While we do not discuss every possible angle here, it should be rapidly clear from just these two examples that any other example is significantly more complicated. Please trust us that every other alternatives that sounds even half-reasonable has been examined and has some critical flaw or another.
+
+**We cannot require all fields, or even only one field, to be valid at all times.**
```
joshtriplett, line 30: :+1:

reference/src/validity/unions.md
```diff
@@ -1,13 +1,243 @@
...
+
+### Niches
+
+## Valid values for #[repr(C)] and Raw-repr unions
+
+`#[repr(C)]` and [Raw-repr][raw repr] unions can take on any byte value.
+
+<details><summary><b>Rationale</b></summary>
+
+The purpose of the Raw-repr is to provide these semantics, which are easy to raeson about. Furthermore, C programmers are used to being able to treat unions like bags of bytes, more or less, and Rust programmers are similarly used to the same with `#[repr(C)]` unions. Therefore, they should both accept any arbitrary byte pattern.
```
joshtriplett, line 72:
```suggestion
The purpose of the Raw-repr is to provide these semantics, which are easy to reason about. Furthermore, C programmers are used to being able to treat unions like bags of bytes, more or less, and Rust programmers are similarly used to the same with `#[repr(C)]` unions. Therefore, they should both accept any arbitrary byte pattern.
```

joshtriplett: Further rationale: some unions are effectively non-exhaustive, because the program may not know about every variant. Consider the case of a u8 discriminant followed by a union, for which the program is expected to ignore unknown variants. Also consider the case where the u8 is *part* of the union at the start of every structure; it may be guaranteed valid but no other field may be.

joshtriplett:
The TBD aspects of this seem to leave enough room for further refinement.

reference/src/validity/unions.md
alercah: Applied the typo fix. I expanded on the rationale in #365 where the rationale for the Raw-repr is actually present. This section is just the rationale for its range of possible values; I clarified as well that padding is not preserved here.

alercah:
Thanks for pointing out the rationale that unions are sometimes supposed to be treated as non-exhuastive so that they can preserve unknown variants. This is a strong rationale for removing tail padding from Raw-repr unions, since it could similarly break forwards compatibility, so I have updated the doc to account for that (except when the union is deliberately over-aligned).

reference/src/validity/unions.md
```diff
@@ -1,13 +1,243 @@
...
+that would make the examples harder to read.
+Additionally, layout of union fields is put in some comments; in these comments,
+the notation `n*pad` means "`n` padding bytes".
 
+## Value model
+
+The possible values of unions are not defined in terms of the values of the
+fields, but rather, a union's possible values are lists of bytes.
+The [representation relation] is trivial in both directions, except for [padding
+bytes] which are uninitialized in all values.
```
RalfJung, line 24:
```suggestion
The [representation relation] is trivial in both directions, except for [padding
bytes] which are uninitialized in all values.
```
Usually padding just isn't part of the value, and hence becomes uninitialized in `encode` (representing a value as bytes). Why do it differently here?

reference/src/validity/unions.md
```diff
@@ -1,13 +1,243 @@
...
+[representation relation]: ../glossary.md#representation-relation
+[padding bytes]: ../glossary.md#padding-byte
+
+</details>
+
+### Niches
+
+## Value space for #[repr(C)] and Raw-repr unions
+
+`#[repr(C)]` and [Raw-repr][raw repr] unions can store any byte value, except for [padding bytes] which are always uninitialized.
```
RalfJung, line 68: A padding byte is not a value so I don't understand what this means.

(reads a 3rd time)

Oh... I parsed ("value, except for padding bytes which are...") as the thing that can be stored, but that's not what you mean. Please reformulate.

Actually, I am not even sure what you mean by "value space". You defined the "value model", as in the set of `Value`s (in MiniRust terms) that make up unions, above. I can only guess that this here is describing the set of `List<AbstractByte>` that can encode a union, which is technically redundant with the previous section but seems worth keeping. However, padding bytes can be represented by any AbstractByte, so in that case the sentence should say that any sequence of bytes is a valid union encoding.

reference/src/validity/unions.md
```diff
@@ -1,13 +1,243 @@
...
+
+<details><summary><b>Rationale</b></summary>
+
+The purpose of the Raw-repr is to provide these semantics, which are easy to reason about. Furthermore, C programmers are used to being able to treat unions like bags of bytes, more or less, and Rust programmers are similarly used to the same with `#[repr(C)]` unions. Therefore, they should both accept any arbitrary byte pattern.
+
+</details>
+
+## Possible niche values
+
+The presence of [padding bytes], and writes to individual fields in general, makes niches hard to come by in unions. A niche representation of a union would have to not only be invalid for every single one of its fields, but also impossible to construct in Safe Rust with any combination of writes to any of its fields.
```
RalfJung, line 78: The two sections above already imply that there can be no niche. I am confused by this third discussion of the same decision.

reference/src/validity/unions.md
```diff
@@ -1,13 +1,243 @@
...
+
+We have not yet reached consensus on whether or not we wish to leave the door open for the possibility that unions with safe field access, or `#[repr(transparent)]` unions with no ZSTs, contain niches:
+
+```rust
+#[repr(transparent)]
+union U { b: bool };
+assert_eq!(size_of::<Option<U>>(), 1); // Requires a niche, which in turn requires that U must be initialized to be valid.
+```
+
+We are **not** describing this case as unspecified, but instead as TBD."Constructible with Safe Rust" is a poorly-defined and very complex invariant, which falls short of the UCG's goals of easily checked, easy to understand (such as it were) semantics, and therefore we are not comfortable leaving the language in this state on an indefinite basis.
```
RalfJung, line 108:
```suggestion
We are **not** describing this case as unspecified, but instead as TBD. "Constructible with Safe Rust" is a poorly-defined and very complex invariant, which falls short of the UCG's goals of easily checked, easy to understand (such as it were) semantics, and therefore we are not comfortable leaving the language in this state on an indefinite basis.
```

reference/src/validity/unions.md
```diff
@@ -1,13 +1,243 @@
...
+}
+let u: U;
+get_b(u); // Compile error: u is not initialized.
+```
+
+Because all bytes of `U` must be initialized for the value to be valid, and this is enforced by the compiler's initialization checks, it might be tempting to assume that `U`'s bytes must always be defined, but this is not a valid assumption. It is equally invalid, however, to assume that `U`'s bytes can be undefined.
+
+<details><summary><b>Rationale</b></summary>
+
+We have not yet reached consensus on whether or not we wish to leave the door open for the possibility that unions with safe field access, or `#[repr(transparent)]` unions with no ZSTs, contain niches:
```
RalfJung, line 100: If they do, however, we need to adjust the value representation / the set of valid representations of a union value. So if we want to keep the door open as you do here, that needs to already be said earlier when the value domain / representation is defined.

reference/src/validity/unions.md
```diff
@@ -1,13 +1,243 @@
...
+
+The main saving grace here is that `#[repr(Rust)]` unions are presently nearly impossible to use correctly anyways, because they do not even guarantee fields at offset 0.
+
+</details>
+
+## Validity of sometimes-padding bytes
+
+We can say that a byte is *sometimes padding* for a union `U` if there is *some* inhabited field `f` such that the byte is either padding for `f` or not a part of `f`.
+
+In that case, the byte will be uninitialized in the value `U{f: /* some value */ }`. By the [monotonicity property], therefore, all sometimes-padding bytes can contain any byte value, be it undefined or any bit pattern with any provenance. Likewise, if multiple bytes are padding for the same field, then they can take on any possible combination of byte values between them.
```
RalfJung, line 118:
```suggestion
In that case, the byte will be uninitialized in the value `U{f: /* some value */ }`. By the [monotonicity property], therefore, all sometimes-padding bytes can contain any byte value, be it undefined or any bit pattern with any provenance.
```
That last sentence is confusing and I don't know what it is supposed to achieve.

reference/src/validity/unions.md
```diff
@@ -1,13 +1,243 @@
...
+```rust
+struct B {
+
+}
+#[repr(Safe)]
+Union u {
+    b: bool, // [bool, 1*pad]
+    u: u16, // [u16]
+}
+let u = U{u: (0xff00, 0}; // resulting bytes: [0xff, 0]
```
RalfJung, line 136: This doesn't parse... did you mean
```suggestion
let u = U { u: 0xff00 }; // resulting bytes: [0xff, 0]
```
?

reference/src/validity/unions.md
```diff
@@ -1,13 +1,243 @@
...
+We can say that a byte is *sometimes padding* for a union `U` if there is *some* inhabited field `f` such that the byte is either padding for `f` or not a part of `f`.
+
+In that case, the byte will be uninitialized in the value `U{f: /* some value */ }`. By the [monotonicity property], therefore, all sometimes-padding bytes can contain any byte value, be it undefined or any bit pattern with any provenance. Likewise, if multiple bytes are padding for the same field, then they can take on any possible combination of byte values between them.
+
+It follows that a union containing an inhabited zero-sized field can contain any bit pattern whatsoever, because all bytes are sometimes-padding bytes. An example of such
+an union is [`MaybeUninit<T>`], which is a union of `T` and `()`.
+
+As per the previous section, however, just because a byte is a sometimes-padding byte does not mean it can always safely be set to uninitialized (or any other value), if this can produce a value not reachable from Safe Rust.
+
+For instance, the following is presently unsound (assuming that `#[repr(Rust)]` is not the [Raw-repr]), even assuming that all fields are placed at offset 0:
```
RalfJung, line 125: "unsound" is a property of safe APIs so this sentence is a category error.
I'd just say that this creates a union value that could not be created in safe Rust and so passing it to unsuspecting code might cause UB -- for example, someone might rely on the fact that "if the first byte is at least 2 then the 2nd byte is initialized". I don't think it is clear whether the crate making the assumption or the crate producing the strange value is wrong -- ultimately only the crate defining `u` can make this call. (Maybe add forward reference to next section, or reorder the sections.)

reference/src/validity/unions.md
```diff
@@ -1,13 +1,243 @@
...
+
+```rust
+struct B {
+
+}
+#[repr(Safe)]
+Union u {
```
RalfJung, line 132:
```suggestion
Union U {
```

reference/src/validity/unions.md
```diff
@@ -1,13 +1,243 @@
...
+    b: bool, // [bool, 1*pad]
+    u: u16, // [u16]
+}
+let u = U{u: (0xff00, 0}; // resulting bytes: [0xff, 0]
+unsafe { (&mut u.b as *mut u8 as *mut MaybeUninit<u8>).offset(1).write(MaybeUninit::uninit()) }; // resulting bytes: [0xff, uninit]
+```
+
+This value is impossible to reach in Safe Rust: the only way to write uninit to the padding is to write to the boolean field. Writing to the integer field must initialize
```
RalfJung, line 140: In C, as far as I know, writing to a field can clobber neighboring padding bytes. So actually, maybe there *is* a way to construct the bad value in safe code?

reference/src/validity/unions.md
```diff
@@ -1,13 +1,243 @@
...
+    pub i: i32
+}
+// Crate b
+pub fn get_i(u: a::U) -> i32 {
+    // Safe: u.i cannot be uninit in Safe Rust.
+    unsafe { u.i } // UNSOUND!
+}
+```
+
+Making this field access safe would require additional an safety invariant that can be understood by the compiler. The UCG WG does not oppose such a safety invariant, but believes it should be opt-in, and an RFC for such a feature is beyond our remit.
```
RalfJung, line 160:
```suggestion
This code could be made sound by convincing the author of crate a to add explicit documentation stating that `i` is always initialized.

Making this field access *safe* would require additional an safety invariant that can be understood by the compiler. The UCG WG does not oppose such a safety invariant, but believes it should be opt-in, and an RFC for such a feature is beyond our remit.
```

reference/src/validity/unions.md
```diff
@@ -1,13 +1,243 @@
...
+// Crate b
+pub fn get_i(s: a::S) -> i32 {
+    // Safe: s.i cannot be uninit in Safe Rust.
+    unsafe { s.i } // Sound.
+}
+```
+
+This struct code, however, is absolutely sound, even in the absence of a safety invariant documented by `S`, because of `S`'s validity invariant: for `S` to be valid, all its fields must be valid, and therefore `i` must be initialized. If it weren't, the definition of `get_i` wouldn't be the problem: the caller would be committing UB by passing an uninitialized `S`. Consequently, the `unsafe` block is redundant.
+
+But for the union type `U`, its validity invariant is not transitive to its fields. `u.i` has no guarantee of validity for `U` to be valid.
```
RalfJung, line 182:
```suggestion
But for the union type `U`, its validity invariant does not imply the one for its fields. `u.i` has no guarantee of validity for `U` to be valid.
```
This is not really a case of "transitivity" IMO.

RalfJung:
Closing due to inactivity. Note that the UCG book isn't really maintained these days -- probably the proper thing would be to write up a markdown document, schedule a t-opsem meeting, use that to update the document, and then FCP the result.

PR #369, opened 2022-10-15
Union discussions

alercah:
Assumes #365 and #368 are merged.

### DIFF
--- a/active_discussion/unions.md
+++ b/active_discussion/unions.md
@@ -1,3 +1,29 @@
 # Unions
 
-TBD
+## Outstanding questions
+
+* Is `#[repr(Rust)]` the bag-o-bytes union repr, or do we want to propose a new repr?
+    * *Discussion:* [#73: Validity of unions][#73]
+* The following questions are all implicitly answered if `#[repr(Rust)]` is the bag-o-bytes repr, but remain open if not:
+    * Do `#[repr(Rust)]` unions guarantee all fields at offset 0?
+        * *Discussion*: [#353: Offsets of union fields][#353]
+    * Do `#[repr(Rust)]` unions have internal padding?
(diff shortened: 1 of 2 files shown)


active_discussion/unions.md
```diff
@@ -1,3 +1,29 @@
 # Unions
 
-TBD
+## Outstanding questions
+
+* Is `#[repr(Rust)]` the bag-o-bytes union repr, or do we want to propose a new repr?
+    * *Discussion:* [#73: Validity of unions][#73]
+* The following questions are all implicitly answered if `#[repr(Rust)]` is the bag-o-bytes repr, but remain open if not:
+    * Do `#[repr(Rust)]` unions guarantee all fields at offset 0?
```
RalfJung, line 8: Even bag-of-bytes unions could have fields at non-0 offsets. Not sure if there is any reason to do that, but this makes it sound like there is a necessary logical implication from bag-of-bytes to 0-offset, which I don't think is the case.

active_discussion/unions.md
```diff
@@ -1,3 +1,29 @@
...
+        * *Discussion*: [#353: Offsets of union fields][#353]
+    * Do `#[repr(Rust)]` unions have internal padding?
+        * *Discussion*: [#354: Do #[repr(Rust)] unions have internal padding?][#354]
+* Do `#[repr(transparent)]` unions ever have niches?
```
RalfJung, line 12: This one, OTOH, *is* answered by bag-of-bytes -- right?

RalfJung:
Closing due to inactivity. Note that the UCG book isn't really maintained these days -- probably the proper thing would be to write up a markdown document, schedule a t-opsem meeting, use that to update the document, and then FCP the result.

Issue #176, opened 2019-07-25
Remove size != stride distinction for array layout


gnzlbg:
There was some discussion here about removing the size != stride distinction from the array layout chapter: https://github.com/rust-lang/unsafe-code-guidelines/pull/162

@rkruppe commented:

> Even if Rust eventually adds a notion of stride distinct from size, whoever designs and introduces that feature already needs update all our docs to account for it, and there's far more of that than just UCG.

And I find it hard to argue against that.

RalfJung:
IMO we should keep the option open or basically make an RFC that says "we will never do this". Because if we close the option here we kind of sneakily make almost the same decision but without consulting anyone.

Cc @petrochenkov @Amanieu

Lokathor:
If we remove it here we just remove a lot of words that have nothing to do with Rust as it is today, without needing to say either way about the future.

The current page is _actually_ confusing and hard to follow because of all the wording about a potential future we probably will not even see.

rkruppe:
As I wrote in #162, I don't think removing the currently present language is the forward compatibility hazard that @RalfJung says it is. IIUC the concern is, if we have only one notion of size and document all our guarantees in terms of it, we don't have the leeway to tease them apart later where needed to support stride!=size. But we're not actually doing a good job at being forward compatible in that sense, e.g. as I wrote in #162:

> Struct layout would probably be affected too (consider `struct Foo(u16, u8); struct Bar(Foo, u8);`) but there's no mention of the stride/size distinction in the corresponding chapter.

I don't see anyone trying to fix this, or even worrying loudly about it, so IMO this is a half-hearted attempt which just complicates UCG without actually achieving anything. What's more, I think this is pretty inherent to this thing -- you'd have to basically design the feature and the migration strategy completely to be able to say how it interacts with all the matters we're covering in this document (e.g., do `<*T>::offset` and `<*T>::add` add the stride or size?).

I'm also repeating myself pointing out that stride==size is greatly entrenched in other documents (reference, rustdoc examples, and probably unofficial resources), in the collective knowledge of Rust programmers, and in lots of unsafe code in the wild. Belief that stride!=size is possible to introduce into Rust today or in the future requires belief that we can somehow solve these all in a backwards compatible fashion. But if that's true, then we can surely also fix the unsafe code guidelines in the same fashion, right?

[...]

PR #447, opened 2023-08-13
Clarify size == stride

JakobDegen:
Blocked pending FCP in #176

### DIFF
--- a/reference/src/layout/arrays-and-slices.md
+++ b/reference/src/layout/arrays-and-slices.md
@@ -1,15 +1,10 @@
 # Layout of Rust array types and slices
 
-## Editor's Note
-
-Issue #176 clarified that as of this writing, Rust does not distinguish between size and stride, ie
-they are always equal.
-
 ## Layout of Rust array types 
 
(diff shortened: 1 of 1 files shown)


reference/src/layout/arrays-and-slices.md
```diff
@@ -1,5 +1,10 @@
 # Layout of Rust array types and slices
 
+## Editor's Note
+
+Issue #176 clarified that as of this writing, Rust does not distinguish between size and stride, ie
+they are always equal.
```
RalfJung, line 6: At the very least, the "Guaranteeing stride == size" section should probably be removed? And generally the entire file should be updated to assume size == stride I think -- right now different parts of the document seem to contradict each other.

RalfJung:
@JakobDegen I have updated the MR to entirely remove references to stride as a separate concept.

RalfJung:
Since we had an FCP, and nobody suggested wording tweaks, I will go ahead and land this.

Also Cc https://github.com/rust-lang/lang-team/pull/216

PR #448, opened 2023-08-13
add collection of deliberate cases of UB

RalfJung:
Fixes https://github.com/rust-lang/unsafe-code-guidelines/issues/158

### DIFF
--- a/resources/deliberate-ub.md
+++ b/resources/deliberate-ub.md
@@ -16,16 +16,17 @@ We should evaluate whether there truly is some use-case here that is not current
   The LLVM IR we generate here contains `noundef` so this UB is not just a specification artifact.<br>
   Furthermore, `compare_exchange` will compare padding bytes, which leads to UB.
   It is not clear how to best specify a useful `compare_exchange` that can work on padding bytes.
-* `bytes` does [a non-atomic plain load that races](https://github.com/tokio-rs/bytes/blob/dea868a4b0eec28877e9013702c0ae12dbc40c4b/src/bytes.rs#L2508),
+* `bytes` used to do [a non-atomic plain load that races](https://github.com/tokio-rs/bytes/blob/dea868a4b0eec28877e9013702c0ae12dbc40c4b/src/bytes.rs#L2508),
   because relaxed loads are claimed to cost too much performance (also see [this LLVM issue](https://github.com/llvm/llvm-project/issues/37064)).
   (Note that LLVM's handling of data races is not enough here, data races still return garbage data -- so this is UB in the generated LLVM IR. Also see [this thread](https://internals.rust-lang.org/t/unordered-as-a-solution-to-bit-wise-reasoning-for-atomic-accesses/11079) on using "unordered".)
   The best fix is probably to improve LLVM optimizations on relaxed loads, so that they are not considered too slow for this use-case.
   However, it is not clear if all the desired optimizations are even legal for relaxed loads -- there is a significant semantic difference between relaxed loads and non-atomic loads, so seeing *some* performance difference is expected.
   Code that works with concurrency can't just expect not to have to pay a cost for that.
+  And anyway current `bytes` does not seem to have any atomics in `ByteMut` any more, so possibly this is resolved.
 * `smol-rs/event-listener` needs an SC fence, but on x86 they prefer a `lock cmpxchg` over an `mfence`.
   To achieve this, they do an [SC `compare_exchange` on a local variable](https://github.com/smol-rs/event-listener/blob/0ea464102e74219aab2932f9eff14418a13268d4/src/notify.rs#L574-L577).
   That is UB (both in Rust and LLVM) since such an operation has no synchronization effect; the compiler could easily see that this variable is never accessed from outside this function and hence optimize it away entirely.
-  The best fix is probably an inline assembly block.
+  The best fix is probably an [inline assembly block](https://github.com/smol-rs/event-listener/pull/71).
 
 ### Cases related to aliasing
 
--- a/resources/deliberate-ub.md
+++ b/resources/deliberate-ub.md
@@ -8,24 +8,27 @@ We should evaluate whether there truly is some use-case here that is not current
 * crossbeam's `AtomicCell` implements an SeqLock, which [is well-known to not be compatible with the C++ memory model](http://www.hpl.hp.com/techreports/2012/HPL-2012-68.pdf).
   Specifically the problem is [this non-atomic volatile read](https://github.com/crossbeam-rs/crossbeam/blob/5d07fe43540d7f21517a51813acd9332744e90cb/crossbeam-utils/src/atomic/atomic_cell.rs#L980) which can cause data races and hence UB.
   This would be fine if we either (a) adopted LLVM's handling of memory races (then the problematic read would merely return `undef` instead of UB due to a data race), or (b) added [bytewise atomic memcpy](https://github.com/rust-lang/rfcs/pull/3301) and used that instead of the non-atomic volatile load.
+  This is currently *not* UB in the LLVM IR we generate, due to LLVM's different handling of read-write races and because the code carefully uses `MaybeUninit`.
 * crossbeam's `AtomicCell` also uses the standard library `Atomic*` types to do atomic reads and writes of [*any type* that has the right size](https://github.com/crossbeam-rs/crossbeam/blob/master/crossbeam-utils/src/atomic/atomic_cell.rs#L928-L932).
   However, doing an `AtomicU32` atomic read on a `(u16, u8)` is unsound because the padding byte can be uninitialized.
   (Also, pointer provenance is lost.)
   To fix this we need to be able to perform atomic loads at type `MaybeUninit<u32>`.
+  The LLVM IR we generate here contains `noundef` so this UB is not just a specification artifact.
 * `bytes` does [a non-atomic plain load that races](https://github.com/tokio-rs/bytes/blob/dea868a4b0eec28877e9013702c0ae12dbc40c4b/src/bytes.rs#L2508),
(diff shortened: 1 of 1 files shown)


CAD97 (approved):
LGTM, though two potential future refinements:

- Differentiate somehow between cases which cause LLVM UB and which are UB in the frontend but not in the backend yet; the former is more pressing than the latter, which could be argued as "just" relying on implementation details.
- Add some examples of patterns that SB considers UB but people want to do anyway (e.g. `&Header`, probably "work in progress" given TB resolves it).

But this is useful even without, and these both are more about the "potentially UB" class than the "definite UB" class that this is more about.

resources/deliberate-ub.md
```diff
@@ -0,0 +1,41 @@
...
+  This would be fine if we either (a) adopted LLVM's handling of memory races (then the problematic read would merely return `undef` instead of UB due to a data race), or (b) added [bytewise atomic memcpy](https://github.com/rust-lang/rfcs/pull/3301) and used that instead of the non-atomic volatile load.
+* crossbeam's `AtomicCell` also uses the standard library `Atomic*` types to do atomic reads and writes of [*any type* that has the right size](https://github.com/crossbeam-rs/crossbeam/blob/master/crossbeam-utils/src/atomic/atomic_cell.rs#L928-L932).
+  However, doing an `AtomicU32` atomic read on a `(u16, u8)` is unsound because the padding byte can be uninitialized.
+  (Also, pointer provenance is lost.)
+  To fix this we need to be able to perform atomic loads at type `MaybeUninit<u32>`.
+* Similar to the previous case, the `atomic-memcpy` crate uses the [standard `Atomic*` types to load potential padding or pointer bytes](https://github.com/taiki-e/atomic-memcpy/blob/b7b82bface0f24c5a9171f2d0626871c61302152/src/lib.rs#L323).
```
taiki-e, line 15: Well, the atomic-memcpy's API is unsafe and passing a type containing uninitialized bytes violates its safety requirements.

resources/deliberate-ub.md
```diff
@@ -0,0 +1,41 @@
...
+  This is a user-space attempt to implement bytewise atomic memcpy, so [adding that as a native operation](https://github.com/rust-lang/rfcs/pull/3301) should fix this.
+* `bytes` does [a non-atomic plain load that races](https://github.com/tokio-rs/bytes/blob/dea868a4b0eec28877e9013702c0ae12dbc40c4b/src/bytes.rs#L2508),
```
taiki-e, line 17: IIRC, this no longer exists in the latest version of bytes.

RalfJung:
> Add some examples of patterns that SB considers UB but people want to do anyway (e.g. &Header, probably "work in progress" given TB resolves it).

We already have an issue open for that, so not sure it's worth also going out and hunting for examples. Other people can make PRs after this one lands. :)

resources/deliberate-ub.md
RalfJung: Ah, I hadn't realized that this is unsafe. I'll remove it then, thanks.

resources/deliberate-ub.md
RalfJung: Interesting, I can't find it either. Do you know what they are doing instead? I can't find any atomics in `BytesMut`, did they just give up on having the mutable version be thread-safe?

RalfJung: This seems to have happened with a [big refactor](https://github.com/tokio-rs/bytes/commit/43ac8e5494e9404327f971937edb092b8cae6a2b). @seanmonstar is the old problem with atomics in `BytesMut::kind` completely gone or is that now just somehow moved somewhere else?

RalfJung:
If the `bytes` issue is truly resolved, then it looks like every item in the list has a least an RFC that would resolve it -- except for needing to do atomic loads at types like `MaybeUninit<i32>`.

resources/deliberate-ub.md
```diff
@@ -0,0 +1,42 @@
+When a crate explicitly acknowledges that what it does is UB, but prefers keeping that code over UB-free alternatives (or there are no UB-free alternatives), that is always a concerning sign.
+We should evaluate whether there truly is some use-case here that is not currently achievable in well-defined Rust, and work with crate authors on providing a UB-free solution.
+
+## Known cases of deliberate UB
+
+### Cases related to concurrency
+
+* crossbeam's `AtomicCell` implements an SeqLock, which [is well-known to not be compatible with the C++ memory model](http://www.hpl.hp.com/techreports/2012/HPL-2012-68.pdf).
+  Specifically the problem is [this non-atomic volatile read](https://github.com/crossbeam-rs/crossbeam/blob/5d07fe43540d7f21517a51813acd9332744e90cb/crossbeam-utils/src/atomic/atomic_cell.rs#L980) which can cause data races and hence UB.
+  This would be fine if we either (a) adopted LLVM's handling of memory races (then the problematic read would merely return `undef` instead of UB due to a data race), or (b) added [bytewise atomic memcpy](https://github.com/rust-lang/rfcs/pull/3301) and used that instead of the non-atomic volatile load.
```
RalfJung, line 10: @taiki-e if I have you around anyway, could you briefly say if this is correct -- if Rust gets atomic bytewise memcpy, then AtomicCell could use them?

resources/deliberate-ub.md
```diff
@@ -0,0 +1,42 @@
...
+  This is currently *not* UB in the LLVM IR we generate, due to LLVM's different handling of read-write races and because the code carefully uses `MaybeUninit`.
+* crossbeam's `AtomicCell` also uses the standard library `Atomic*` types to do atomic reads and writes of [*any type* that has the right size](https://github.com/crossbeam-rs/crossbeam/blob/master/crossbeam-utils/src/atomic/atomic_cell.rs#L928-L932).
+  However, doing an `AtomicU32` atomic read on a `(u16, u8)` is unsound because the padding byte can be uninitialized.
+  (Also, pointer provenance is lost.)
+  To fix this we need to be able to perform atomic loads at type `MaybeUninit<u32>`.
```
RalfJung, line 15: @taiki-e I am curious, given your crate https://github.com/taiki-e/atomic-maybe-uninit/ (which IIUC uses inline assembly to add the missing atomic `MaybeUninit` loads) -- why does crossbeam not use that?

resources/deliberate-ub.md
RalfJung: Also, couldn't atomic-memcpy use https://github.com/taiki-e/atomic-maybe-uninit/ to avoid those safety requirements (potentially making it an option for use in `AtomicCell`)?

resources/deliberate-ub.md
taiki-e: Yeah, RFC 3301 should be enough to fix it.

That said, I think we will likely fix it by adopting "(c) use inline assembly instead of the non-atomic volatile load" before using the implementation of RFC 3301.

resources/deliberate-ub.md
taiki-e: > Also, couldn't atomic-memcpy use https://github.com/taiki-e/atomic-maybe-uninit/ to avoid those safety requirements (potentially making it an option for use in `AtomicCell`)?

https://github.com/taiki-e/atomic-memcpy/pull/7 is a PR for this, but it was pending because I thought the code generated could be more optimized on the atomic-maybe-uninit side (https://github.com/taiki-e/atomic-maybe-uninit/issues/1).

resources/deliberate-ub.md
taiki-e: > why does crossbeam not use that?

This is because there were several issues. Many of them have already been resolved, and the main remaining issue is that no mechanism (something like `cfg(target_has_atomic)`) is provided by the atomic-maybe-uninit side to handle the lack of portability well (https://github.com/taiki-e/atomic-maybe-uninit/issues/9).

I recently introduced [such a mechanism in another crate](https://github.com/taiki-e/portable-atomic/blob/8d1002fcf62909f3d6f57327b28419597c9d2614/src/utils.rs#L247-L693) and am considering porting it to atomic-maybe-uninit.

resources/deliberate-ub.md
RalfJung: Sure, inline assembly is always an option, though usually not a nice one.

RalfJung:
I explored `bytes` a bit more and I'm fairly sure the problematic code is completely gone, as [part of a change](https://github.com/tokio-rs/bytes/commit/43ac8e5494e9404327f971937edb092b8cae6a2b) describes as "Removed ability to mutate the underlying memory from the `Bytes` type". So I removed this item from the list. The smol-rs crates are now using inline assembly instead of relying on compare_exchange having the effects of a fence, so that is fixed, too. That leaves only the `AtomicCell` issues, both for the fast path and the fallback path. :)

I'll merge this now, we can always edit this further later. This was definitely a useful update!

resources/deliberate-ub.md
```diff
@@ -0,0 +1,41 @@
+When a crate explicitly acknowledges that what it does is UB, but prefers keeping that code over UB-free alternatives (or there are no UB-free alternatives), that is always a concerning sign.
+We should evaluate whether there truly is some use-case here that is not currently achievable in well-defined Rust, and work with crate authors on providing a UB-free solution.
+
+## Known cases of deliberate UB
+
+### Cases related to concurrency
+
+* crossbeam's `AtomicCell` "fallback path" uses a SeqLock, which [is well-known to not be compatible with the C++ memory model](http://www.hpl.hp.com/techreports/2012/HPL-2012-68.pdf).
```
64, line 8: This link doesn't load for me, I think because of the `http://` rather than `https://`.

RalfJung: Good catch, I'll fix it in the repo. (My browser is in https-always mode so even the http link works here...)

Issue #436, opened 2023-08-01
Rendered version of the book no longer updates


RalfJung:
https://rust-lang.github.io/unsafe-code-guidelines/ should be updated on a push to this repo, but Travis CI has long stopped working. This needs to be ported to Github Actions.

@Mark-Simulacrum @pietroalbini is there a configuration we can just copy for that? Is there setup you need to do on the admin side?

RalfJung:
Looks like the RFCs repo also [deploys a book to GH-pages](https://github.com/rust-lang/rfcs/blob/master/.github/workflows/deploy.yml), though they need some extra Pythin logic to create the book in the first place. Still seems like a reasonable template for us.

RalfJung:
Yay this worked. :)

PR #452, opened 2023-08-19
port CI to GHA

RalfJung:
Fixes https://github.com/rust-lang/unsafe-code-guidelines/issues/436. But actually landing this will need some @rust-lang/infra help. The GHA script is mostly copied from the RFCs repo and I think some settings are required to let it actually deploy the book. Though I do seem to have the power to edit https://github.com/rust-lang/unsafe-code-guidelines/settings/pages myself so maybe I can just try that...?


RalfJung:
Well, let's see if this works...

Issue #304, opened 2021-09-01
Terminology around layout, ABI, representation, and bit validity


joshlf:
I'm writing the docs for the following type:

```rust
#[repr(packed)]
pub struct Unalign<T>(T);
```

I'd like to explain in the docs that the layout/ABI/representation/bit validity/whatever is the same as that of `T` with the exception of alignment, but I'm not sure what terminology to use. A few observations:
- Per the [Rust reference](https://doc.rust-lang.org/reference/type-layout.html), "layout" refers at least to size, alignment, and field offsets
- In some docs - such as those for [MaybeUninit](https://doc.rust-lang.org/std/mem/union.MaybeUninit.html#layout) - the phrase "ABI" is also used
- The UCG WG [has used the phrase "validity"](https://github.com/rust-lang/unsafe-code-guidelines/blob/master/active_discussion/validity.md) to refer to the set of allowed bit patterns for a type
- The phrase "representation" seems to be fairly consistently used to refer to the `repr` attribute (whether or not the attribute is present and, if present, what values are given)

My questions are:
- Does "layout" also encompass a type's bit validity?
  - If not, what phrase refers to a type's bit validity? Do we just have to say "layout and bit validity" if we want to encompass everything?
- What does the phrase "ABI" mean?

As a meta point, it seems like the terminology has been somewhat standardized over time as folks have gravitated towards a common set of terms, but there's still some disagreement (e.g., I haven't seen "ABI" used in most places, but it does show up). It would be good to decide on a standard set of terms, and to standardize their usage across the docs.

Lokathor:
I believe that ABI refers to, at least, how the data is passed between data. It might also mean other things, but it at least means that much.

joshlf:
> how the data is passed between data

Passed between functions?

Lokathor:
Yes. When data is passed between functions both functions must agree where the caller will put the data for the callee's arguments, and also where the callee will leave the data for the caller's return value.

RalfJung:
Yeah, sadly we never reached consensus for what exactly is part of "layout" and "ABI"... also see https://github.com/rust-lang/unsafe-code-guidelines/issues/122.

I honestly don't even really care which terms we end up using, as long as we have names for all the relevant things and we can all agree on using the same names. :)

joshlf:
[...]

PR #457, opened 2023-08-31
separate ABI from Layout

RalfJung:
In my experience, excluding the function call ABI better reflects the actual uses of the term "layout". So this aligns the glossary with how terms are used in practice.

Fixes https://github.com/rust-lang/unsafe-code-guidelines/issues/304

### DIFF
--- a/reference/src/glossary.md
+++ b/reference/src/glossary.md
@@ -1,5 +1,17 @@
 ## Glossary
 
+### ABI (of a type)
+[abi]: #abi-of-a-type
+
+The *function call ABI* or short *ABI* of a type defines how it is passed *by-value* across a function boundary.
+Possible ABIs include passing the value directly in one or more registers, or passing it indirectly as a pointer to the actual data.
+The space of all possible ABIs is huge and extremely target-dependent.
+Rust therefore does generally not clearly define the ABI of any type, it only defines when two types are *ABI-compatible*,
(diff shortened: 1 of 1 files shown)


reference/src/glossary.md
```diff
@@ -1,5 +1,17 @@
 ## Glossary
 
+### ABI (of a type)
+[abi]: #abi-of-a-type
+
+The *function call ABI* or short *ABI* of a type defines how it is passed *by-value* across a function boundary.
```
bjorn3, line 6: Maybe call this "calling convention" instead? The ABI is the combination of the memory layout, calling convention, linkage model and every other part of the interface between compiled object files, dynamic libraries or the kernel.

RalfJung: "ABI" is the term used in existing documents such as [this one](https://rust-lang.github.io/rfcs/3391-result_ffi_guarantees.html). I wouldn't mind a different term but that might be fighting windmills...

CAD97: Calling convention isn't quite correct either, because two types can have the same calling convention but not the same layout, and what's meant when discussing a type's function call ABI is the exact set of properties needed to ensure two function signatures are ABI compatible, which includes both calling convention and layout.

CAD97 (approved):

RalfJung:
@rust-lang/opsem is anyone not okay with this PR? If I don't hear any complaints the next few days I'm going to merge. :)

digama0 (approved):

Issue #213, opened 2019-10-17
Define "allocation"


gnzlbg:
We use the word "allocation" in the abstract machine, e.g., when talking about provenance, or when saying that each stack variable lives on its own "allocation". And we might start using it when talking about the guarantees of `Rc` (see https://github.com/rust-lang/rust/issues/64484). So I think we should define this word in the glossary for now.

While it is possible to write Rust programs that have multiple address spaces, I think it would be reasonable to just start by assuming that there is only just one address space.

Lokathor:
I think that we should br careful about such a simplification. Usually it's easier to ignore complications than to stick them into the design later.

gnzlbg:
Note that two allocations with the same address but in distinct address spaces are already distinct allocations, so one can think of the address space a pointer refers to as being part of its provenance.

RalfJung:
IMO the only satisfying definition of allocation is to actually give a (preliminary) instance of the [Rust memory interface](https://github.com/rust-lang/unsafe-code-guidelines/blob/master/wip/memory-interface.md). Everything else is going to be imprecise and not very useful IMO.

PR #464, opened 2023-09-27
define 'allocation' and 'memory address'

RalfJung:
Fixes https://github.com/rust-lang/unsafe-code-guidelines/issues/213

### DIFF
--- a/reference/src/glossary.md
+++ b/reference/src/glossary.md
@@ -16,7 +16,7 @@ For instance `#[repr(C)] struct S(i32)` is (guaranteed to be) layout-compatible 
 [abstract byte]: #abstract-byte
 
 The *byte* is the smallest unit of storage in Rust.
-Memory allocations are thought of as storing a list of bytes, and at the lowest level each load returns a list of bytes and each store takes a list of bytes and puts it into memory.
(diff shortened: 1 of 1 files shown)


reference/src/glossary.md
```diff
@@ -119,6 +136,12 @@ Note that layout does not capture everything that there is to say about how a ty
 Note: Originally, *layout* and *representation* were treated as synonyms, and Rust language features like the `#[repr]` attribute reflect this. 
 In this document, *layout* and [*representation*][representation relation] are not synonyms.
 
+### Memory Address
+[memory address]: #memory-address
+
+A *memory address* is an integer value that identifies where in the physical machine memory some data is stored.
```
mkroening, line 142: Bare-metal programmers may confuse this with virtual/physical memory addresses from paging.

Is the term physical machine already established in this context?

RalfJung: Ah right, I meant "physical" in contrast to the Abstract Machine, but there's always more layers of abstractions further down the stack. I adjusted the wording, is it better now?

mkroening: Great, thank you! :)

saethlin (approved):

RalfJung:
No objections, one approval -- I'll land this, please file an issue or a PR if you disagree with anything. :)

PR #538, opened 2024-10-09
update the docs for layout guarantees of option-like enums

RalfJung:
This updates our docs to match the FCP in https://github.com/rust-lang/rust/pull/130628. The UCG reference has been a valuable resource for figuring out what is and is not guaranteed about layout, so it seems worth keeping it up-to-date.

(Medium-term, I expect this to be superseded by @chorman0773's new layout chapter for the reference, but we're not there yet.)

### DIFF
--- a/reference/src/layout/enums.md
+++ b/reference/src/layout/enums.md
@@ -295,49 +295,44 @@ apply, as described below.
 
 #### Discriminant elision on Option-like enums
 
-(Meta-note: The content in this section is not fully described by any RFC and is
-therefore "non-normative". Parts of it were specified in
-[rust-lang/rust#60300]).
+(Meta-note: The content in this section have been turned into stable guarantees
+[via this
+FCP](https://github.com/rust-lang/rust/pull/130628#issuecomment-2402761599).).
 
-[rust-lang/rust#60300]: https://github.com/rust-lang/rust/pull/60300
-
-**Definition.** An **option-like enum** is a 2-variant `enum` where:
+**Definition.** The fully monomorphized form of a 2-variant `enum` is called an **option-like enum**
+if all of the following are satisfied:
 
 - the `enum` has no explicit `#[repr(...)]`, and
-- one variant has a single field, and
-- the other variant has no fields (the "unit variant").
+- one variant has a single field with a type that guarantees discriminant
+  elision (to be defined below), and
+- the other variant has only 1-ZST fields (the "unit variant").
 
 The simplest example is `Option<T>` itself, where the `Some` variant
 has a single field (of type `T`), and the `None` variant has no
-fields. But other enums that fit that same template fit.
+fields. But other enums that fit that same template also qualify, e.g.
+`Result<T, ()>` or `Result<(), T>`.
 
 **Definition.** The **payload** of an option-like enum is the single
 field which it contains; in the case of `Option<T>`, the payload has
 type `T`.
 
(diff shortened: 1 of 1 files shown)


RalfJung:
@rust-lang/opsem could someone take a look at this so we can get it merged? :)

reference/src/layout/enums.md
```diff
@@ -295,49 +295,43 @@ apply, as described below.
...
+[via this
+FCP](https://github.com/rust-lang/rust/pull/130628#issuecomment-2402761599).).
 
 **Definition.** An **option-like enum** is a 2-variant `enum` where:
 
 - the `enum` has no explicit `#[repr(...)]`, and
-- one variant has a single field, and
-- the other variant has no fields (the "unit variant").
+- one variant has a single field with a type that guarantees discriminant
+  elision (to be defined below), and
```
saethlin, line 306: Restricting the definition of an option-like enum based on the type of the single field is quite strange to me. Because I'd think that excludes `Option<T>` from the definition, because `T` could be `usize` or any type without guaranteed discriminant elision.

RalfJung: @saethlin does this answer your question? I don't think I quite understood the question, TBH.

saethlin: Whoops, I didn't notice you'd responded.

The text here is just confusing to me, because what I'm imagining is a reader saying "Is this an option-like enum":
```rust
enum Thing<T> {
    Yes(T),
    No,
}
```
Because by the text above, `Thing<T>` is not "an option-like enum" because its single field does not guarantee discriminant elision. This would be though:
```rust
enum Thing {
    Yes(NonZeroUsize),
    No,
}
```

...but *directly below this* the reference already says that `Option<T>` is an example of an option-like enum.

:shrug: 

I think you're making an assumption that the reader will interpret words a bit differently when a generic parameter is included. If other can confirm that they have read this and don't find the wording confusing I'm happy to be overruled.

chorman0773: For this reason, in my type-layout docs PR, I chose to separate the idea of "Is an enum for which discriminant elision may apply" (in my text, called a *discriminant elision elible*)  and "A field of this type is subject to discriminant elision" (called *elision candidate type*). When the two are combined is when the result is guaranteed.
I also note explicitly that the determiniation is post-mono. and explicitly call out both `Option<T>` and `Result<T,E>`.

RalfJung: I am implicitly assuming that we are talking about a fully monomorphic type here. It doesn't make sense to ask about the layout of a generic type. I can make that more explicit (though this will come at the cost of readability). I don't see how splitting this concept into two separate concepts is helpful.

saethlin:
The definition at the top is at least something to point to. I'm still not enthused about the wording, but I think it's better to merge this and wait for someone to actually be confused than continue to haggle over it.

saethlin (approved):

saethlin:
I don't know which of the merge buttons you want to use here, so I'll let you merge this :)

RalfJung:
I see only one merge button?^^ :)

ia0:
I guess the question was between merge, rebase, and squash (all sharing the same button using a drop down for selection). But given the git history and how this PR was merged, the answer was indeed merge 😅 This can be enforced with a branch rule if needed, such that only one option is available in the drop down, making such question automatically answered in the future.

PR #551, opened 2025-01-09
Added definitions to the glossary

adfernandes:
Specifically:
- regulatory approach
- memory safety (spatial and temporal)
- mc/dc coverage
- integer overflow
- functional purity

as per [this issue](https://github.com/rustfoundation/safety-critical-rust-consortium/issues/123#issuecomment-2551605265)


adfernandes:
Thanks for the comments, @RalfJung - in retrospect, I think I agree with you.

If I could as not to close this issue quite yet, I'm going to bring this up to the committee, because I think there's a fair bit of conflation between "Unsafe Rust" and "**Safety-Critical** Unsafe Rust".

_Needs some thought..._

PR #566, opened 2025-05-04
Archive most pages in the reference

Noratrieb:
People continue being confused by these pages. This archives them effectively deleting their content. The old pages are still linked via GitHub permalink for the historians. This archiving strategy creating dead links to the existing pages.


Noratrieb:
@RalfJung

ia0:
The Introduction of the Reference [says](https://rust-lang.github.io/unsafe-code-guidelines/introduction.html#:~:text=It%20is%20largely%20abandoned%20right%20now) "[This reference] is largely abandoned right now". Is the plan to resurrect it? Or should _that_ disclaimer be made more visible? Or maybe even remove all sections except the Glossary?

Lokathor:
if parts are known to be not current, we should delete them rather than have bad information in the book.

Noratrieb:
I've also thought about unpublishing the entire book off GitHub pages (so it's still present on GitHub but no longer published) but I didn't like the idea of making the links dead. On the other hand we could just stub it all out on GitHub pages with a note on why it was removed and where to find the version.

Noratrieb:
> if parts are known to be not current, we should delete them rather than have bad information in the book.

It depends on what we want this book to be. If it's just about accurate descriptions then the entire thing should be deleted, because those belong in the *Reference*.

RalfJung:
I don't have a strong opinion on how to best frame this. However it seems odd that in the table of contents, we now have "suggestions" mixed with things that are not labeled as "suggestions".

Cc @rust-lang/opsem @chorman0773

chorman0773:
Frankly, my $.02 is that we deprecate the UCG as a whole and transition to making guarantees in the Reference (and other documents like the minirust spec for a more programmatic definition). In my view, this has been the *de facto* state for the past couple years anyways, with the repo being used as an issue tracker for outstanding technical questions for T-opsem.

saethlin:
I agree.

Noratrieb:
Okay, I'll rework the PR to archive these pages.

Noratrieb:
I implemented this, archiving all the pages except the glossary by replacing them with a stub.

RalfJung:
LGTM. In terms of t-opsem people, @saethlin also agreed and nobody disagreed. @chorman0773 as a main contributor to these documents also agreed. So I'll go ahead and land this.

PR #570, opened 2025-05-23
add list of assumptions we make about LLVM

RalfJung:
Some of the things we want people to do with Rust can currently not be expressed in LLVM in a way that is fully backed by the LLVM LangRef. Let's collect a list of those cases.


CAD97 (approved):

RalfJung:
Cc @nikic -- you are probably aware of these, but if you need to convince someone which parts of LLVM are important to figure out for Rust, maybe this helps. ;)

